PR #32, opened 2017-08-04
Adds Image component

### LINKED ISSUE #21: Image component

hakimelek:
* Adds Image component.

Fixes #21

### DIFF
--- a/.storybook/Image.js
+++ b/.storybook/Image.js
@@ -12,14 +12,7 @@ storiesOf('Image', module)
   })(() => (
     <Image src='https://www.priceline.com/home/public/assets/images/photos/photo-aruba.jpg' />
   )))
-
-  .add('Fixed width', () => (
-    <Image
-      src='https://www.priceline.com/home/public/assets/images/photos/photo-aruba.jpg'
-      width='300px'
-    />
-  ))
-
+  
   .add('Responsive width, with Box', () => (
     <Box width={1/2}>
       <Image
--- a/.storybook/Image.js
+++ b/.storybook/Image.js
@@ -1,0 +1,29 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { withInfo } from '@storybook/addon-info'
+import { Image, Box } from '../src'
+
+const description = 'A low-level layout component that renders an image'
+
+storiesOf('Image', module)
+  .add('Image component', withInfo({
+    text: description,
+    inline: true
+  })(() => (
+    <Image src='https://www.priceline.com/home/public/assets/images/photos/photo-aruba.jpg' />
+  )))
+
+  .add('Fixed width', () => (
+    <Image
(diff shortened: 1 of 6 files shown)


src/Image.js
```diff
@@ -0,0 +1,18 @@
+import styled from 'styled-components'
+import { space, width, color } from 'styled-system'
```
jxnblk: I think you can ditch this import, since it's not being used

.storybook/Image.js
```diff
@@ -0,0 +1,29 @@
...
+    text: description,
+    inline: true
+  })(() => (
+    <Image src='https://www.priceline.com/home/public/assets/images/photos/photo-aruba.jpg' />
+  )))
+
+  .add('Fixed width', () => (
+    <Image
+      src='https://www.priceline.com/home/public/assets/images/photos/photo-aruba.jpg'
+      width='300px'
```
jxnblk: The `width` and `height` attributes are unitless pixel values – different from CSS. The `width` and `height` attributes should **always** be set on images to help with page layout and prevent reflows when images load. CSS values can override these HTML attributes, if needed, for responsive sizing, but generally that can be handled with wrapping layout components like `<Box />`

README.md
```diff
@@ -184,6 +184,22 @@ The `width` prop accepts an array value to set different widths at different bre
...
+<Image src='https://priceline.com/myimage.png' />
+
+// Adding fixed width
+<Box src='https://priceline.com/myimage.png' width='300px' />
+```
+
+Prop | Type | Description
+---|---|---
+src | string | Sets the image source of the element
+width | string | Sets the image width
```
jxnblk: We should probably leave this as the native HTML attribute and use layout components like `<Box />` to handle width.

README.md
```diff
@@ -184,6 +184,22 @@ The `width` prop accepts an array value to set different widths at different bre
...
 
+### `<Image />`
+
+Use the `<Image />` component to render a primitive image that saves the aspect ratio.
+
+```jsx
+<Image src='https://priceline.com/myimage.png' />
+
+// Adding fixed width
+<Box src='https://priceline.com/myimage.png' width='300px' />
```
jxnblk: Should be `Image` right?

src/__tests__/Image.js
```diff
@@ -0,0 +1,19 @@
...
+describe('Image', () => {
+  test('renders', () => {
+    const json = renderer.create(<Image src={imageSrc} />).toJSON()
+    expect(json).toMatchSnapshot()
+  })
+
+  test('width prop sets width sizing', () => {
+    const json = renderer.create(<Image src={imageSrc} width='300px' />).toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveProperty('props.width', '300px')
```
jxnblk: ✂️

PR #35, opened 2017-08-11
Add icons and Icon component

### LINKED ISSUE #8: Add icons and Icon component

jxnblk:
Fixes #8


hakimelek:
Nice shiny blue ;)
Do you we always need to run `npm run prepare` in order for the icons to work?

hakimelek:
Maybe a good idea to include `withInfo`, and one with an example for how to pass color and size.

jxnblk:
The [`npm prepare`](https://docs.npmjs.com/misc/scripts) script is automatically ran before publishing and after a local install, so generally the icons will be parsed and written to JSON automatically. This is similar to how we keep the `dist` file ignored in git, but it's generated when publishing

PR #41, opened 2017-08-20
Button component

hakimelek:
Adds the Button component.

#19


src/BackgroundImage.js
```diff
@@ -0,0 +1,26 @@
+import styled from 'styled-components'
```
jxnblk: As we discussed in person, let's merge the BackgroundImage PR first and see if that clears up some of these unrelated changes

src/Button/Button.js
```diff
@@ -0,0 +1,103 @@
+import React from 'react'
```
jxnblk: Let's keep the folder hierarchy flat for now, which will help with imports and merging – I don't see much value in putting a module named `Button` inside a folder of the same name

src/Button/GreenButton.js
```diff
@@ -0,0 +1,11 @@
+import styled from 'styled-components'
+import Button from './Button'
+
+const GreenButton = styled(Button)`
```
jxnblk: 👍

jxnblk: Not to open a can of worms, but it might be worth sharing this pattern with your team and getting a sense of what they think of the naming convention here

src/Button/Button.js
```diff
@@ -0,0 +1,103 @@
...
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import { fontSize, space } from 'styled-system'
+import { theme } from '../'
+
+const size = props => {
+  switch(props.size) {
+    case 'small':
```
jxnblk: This seems a lot more readable than the ternaries

src/Button/Button.js
```diff
@@ -0,0 +1,103 @@
...
+      return { 
+        height: '32px',
```
jxnblk: Fixed heights on buttons and inputs used to be sketchy with Firefox – make sure you do some monkey testing in a few browsers to make sure it looks okay

src/Button/Button.js
```diff
@@ -0,0 +1,103 @@
...
+  
+  &:hover {
+    background-color: ${props => props.theme.colors.blue1};
+  }
+
+  &:disabled {
+    opacity: 0.25;
+  }
+
+  width: ${props => props.fullWidth ? '100%' : 'initial' };
```
jxnblk: This is okay for now, but if you pull this out into its own function like `size`, you can avoid adding `width: initial` when `fullWidth` prop is falsy

src/Button/Button.js
```diff
@@ -0,0 +1,103 @@
...
+
+  ${size}
```
jxnblk: Did you mean to add the `space` function here as well?

src/Button/OutlineButton.js
```diff
@@ -0,0 +1,22 @@
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import Button from './Button'
+
+const OutlineButton = styled(Button)`
+  background-color: ${props => props.theme.colors.white};
```
jxnblk: I'd go with `background-color: transparent` here for support on non-white backgrounds

src/Button/OutlineButton.js
```diff
@@ -0,0 +1,22 @@
...
+  color: ${props => props.color || props.theme.colors.blue};
```
jxnblk: You could use a default prop here instead of the `||`

src/Button/OutlineButton.js
```diff
@@ -0,0 +1,22 @@
...
+  border-width: 2px;
+  border-color: ${props => props.color || props.theme.colors.blue};
```
jxnblk: In CSS, `border-color` is the same as the `color` for the element, so this line isn't needed. Same goes for if you were to refactor this with `boxShadow`

src/Button/OutlineButton.js
```diff
@@ -0,0 +1,22 @@
...
+
+  &:hover {
+    background-color: ${props =>  props.theme.colors.white};
```
jxnblk: I thought the hover state was a dark blue color?

src/Button/OutlineButton.js
jxnblk: I'd also either use the styled-system `color` function or a function that can pick up values from `theme.colors` by key since that's how the Box component works. E.g. the prop `color='blue'` gets the color `theme.colors.blue`

src/__tests__/Button.js
```diff
@@ -0,0 +1,79 @@
...
+    expect(json).toHaveStyleRule('font-size', '16px')
+  })
+
+  test('fullWidth prop sets width to 100%', () => {
+    const json = renderer.create(<Button fullWidth theme={theme} />).toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('width', '100%')
+  })
+
+  test('renders GreenButton', () => {
```
jxnblk: Maybe pull these into their own test files to mirror the `src` folder?

src/BackgroundImage.js
hakimelek: I will rebase on master

src/Button/Button.js
hakimelek: We can make it flat for now, but I think we will have to introduce some hierarchy later on in imo.

src/Button/Button.js
hakimelek: Sure!

src/Button/Button.js
hakimelek: So maybe just `${props => props.fullWidth ? 'width: 100%' : null}`

hakimelek: So maybe just `${props => props.fullWidth ? 'width: 100%' : null}`?

src/Button/Button.js
hakimelek: I didn't try it, does it add `p`, `m`, `w` ?

src/Button/OutlineButton.js
hakimelek: So maybe add `${color}` imported from `styled-system`?

src/Button/OutlineButton.js
hakimelek: We need that yet!

src/Button/Button.js
jxnblk: Yeah, I think that's a cleaner pattern to follow for styled-components

jxnblk: Might need the `;` in there too for it to compile though

src/Button/Button.js
jxnblk: Yes, the `space` function adds these props (not `width` though): https://github.com/jxnblk/styled-system#space

src/Button/OutlineButton.js
jxnblk: Maybe leave out the hover states and refactor after the colors are added then?

src/Button/GreenButton.js
```diff
@@ -0,0 +1,11 @@
...
+  background-color: ${props => props.theme.colors.green};
+
+  &:hover {
+    background-color: ${props => props.theme.colors.green1};
+  }
+`
+export default GreenButton
```
BeniCheni: Can this component become from <GreenButton /> to <ColorButton color='green' />? (thinking that maybe the button can be the colored button more dynamic instead of a fixed color; 
 same thought applies to the `RedButton`)

jxnblk: We discussed this in person, but my suggestion was to start with explicitly named components, and abstract into props later if we need it. Part of the thinking is that having a component available is a little more obvious than props. 

Based on what I've seen in the product, we're mostly only using green and blue buttons, and by handling these as separate components we'll need less logic to handle hover, focus, and active states when those are added in.

src/BackgroundImage.js
jxnblk: You should be able to pull the latest from master now and merge it into this branch. If you run into really hairy merge conflicts, I can help out

src/Button/Button.js
jxnblk: Maybe, but we have no idea what that hierarchy should be yet – I'd say generally hold off on organizing things into folders until it becomes a development problem

jxnblk:
Went ahead and pushed a few changes to get this merged in. We'll need to add hover states in once we have the full color scales in (#49), which can be handled in a separate PR

PR #48, opened 2017-08-25
<Text> Support For HTML h1-h6 Elements

BeniCheni:
Support HTML heading elements(i.e. h1-h6) using syntax such as:
```
<Text.h3 />
```

Relate to issue @ pricelinelabs/design-system#47
text-headings


.storybook/Text.js
```diff
@@ -45,3 +45,9 @@ storiesOf('Text', module)
       <Text color='green'>Hello Green</Text>
     </div>
   ))
+  .add('Using <h3> heading', () => (
+    <div>
+      <Text color='purple'>Hello Purple</Text>
```
hakimelek: I don't see `<Text.h3>` in the story.

hakimelek:
No tests? :'(

BeniCheni:
@hakimelek , thanks for the feedback. Added unit tests to cover h1 - h6 test cases and updated snapshot. Please review again at your convenience.

.storybook/Text.js
BeniCheni: Nice catch. Updated the story,

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pcln-design-system",
-  "version": "1.0.0-1",
+  "version": "1.1.0-1",
```
jxnblk: Leave the version number as-is until after we merge. The other PRs are likely to go to the next release as well. Also fwiw, the `-1` denotes a beta version and we wanna hold off on a v1 until things are a little more stable

src/Text.js
```diff
@@ -63,6 +63,12 @@ Text.propTypes = {
   py: numberStringOrArray
 }
 
+Text.h1 = Text.withComponent('h1')
```
jxnblk: This looks great! I was suggest that these be rolled into a Heading component per #20 – I think it can make the JSX a little more readable to see where headings are, but open to other thoughts

jxnblk: Also we may want to set a default `fontSize` prop that would pick up from `theme.fontSizes` - maybe some sensible default for general headings

package.json
BeniCheni: Great thought. Thanks for sharing. Reverted the version back to `1.0.0-1`.

src/Text.js
BeniCheni: @jxnblk , cool. I've checked out the details of #20. Thanks for cross-referencing to #48. I've assigned #48 & #20 to myself.

This PR would address the scope of #48. If it looks okay, we can take the PR and resolve #48. I can try out the Heading component in local and tackle #20.

Does that sound like a viable plan? :)

BeniCheni: Started working on another feature branch named [heading-component](https://github.com/pricelinelabs/design-system/tree/heading-component) to tackle #20 .

Since all tests pass & 1 approval is received, going to take this PR to resolve the scope #48. A new PR from [heading-component](https://github.com/pricelinelabs/design-system/tree/heading-component) will be submitted for #20.

Issue #50, opened 2017-08-25
Consider adding aliases for breakpoint sizes, media queries, and perhaps other arrays of pixel sizes


747823:
This would be an alternative to using the array index. For example, consider `theme.mediaQueries`.
Current usage inside styled component:
```
${props => props.theme.mediaQueries[0]} {
  css here
}
```

Aliased version would do the same thing, but look like this:
```
${props => props.theme.mediaQueries.sm} {
  css here
}
```

Size aliases could follow the bootstrap naming convention, so basically:
`xs`, `sm`, `md`, `lg`, `xl`

PR #51, opened 2017-08-25
Added aliases for breakpoints and media queries

747823:
#50

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -3,6 +3,16 @@ import icons from '../icons.json'
 
 const createMediaQuery = n => `@media screen and (min-width:${n}em)`
 
+const addAliases = (arr, aliases) =>
+  aliases.forEach((key, i) =>
+    Object.defineProperty(arr, key, {
+      enumerable: false,
+      get() {
+        return this[i]
+      }
+    })
+  )
+
 export const breakpoints = [
   32,
   40,
@@ -11,6 +21,16 @@ export const breakpoints = [
 ]
 
 export const mediaQueries = breakpoints.map(createMediaQuery)
+
+const aliases = [
+  'sm',
+  'md',
+  'lg',
+  'xl'
+]
+
+addAliases(breakpoints, aliases)
+addAliases(mediaQueries, aliases)
 
 export const space = [
   0,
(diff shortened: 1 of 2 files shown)


src/theme.js
```diff
@@ -3,6 +3,16 @@ import icons from '../icons.json'
 
 const createMediaQuery = n => `@media screen and (min-width:${n}em)`
 
+const addAliases = (arr, aliases) =>
+  aliases.forEach((key, i) =>
+    Object.defineProperty(arr, key, {
+      enumerable: false,
```
BeniCheni: Is there a specific reason the to-be defined `key` object property not enumerable? Perhaps `for...in loop` or `Object.keys()` type of code can be benefited with the `key` property, if it's enumerable?

747823: Basically yeah I didn't want the aliases to show up in the array of breakpoints by accident, in case someone is iterating over it with something like `for...in`. It could be a breaking change if they did that, and I think it's just more intuitive not to change the enumerable behavior... they are aliases after all, not separate breakpoints.

jxnblk: This is handy – I wouldn't have thought to define properties on the array

src/theme.js
```diff
@@ -12,6 +22,16 @@ export const breakpoints = [
 
 export const mediaQueries = breakpoints.map(createMediaQuery)
 
+const aliases = [
+  'sm',
+  'md',
+  'lg',
+  'xl'
```
jxnblk: I like these aliases, but I know some prefer to have things spelled out more explicitly (e.g. `xsmall`, `medium`), do you think there's a general preference one way or another among your team members?

jxnblk:
Also, can you add some docs in the README.md for this? Eventually, I think we'll split up the docs into multiple files, but need to have this documented somewhere

src/theme.js
BeniCheni: My vote goes to having these aliases spelled out. (i.e. the longest one here would be `extraLarge`; still only 10 letters) The benefit is to have clearly expressed terminology to avoid ambiguity, while the effort is still at most 10 letters. 😄

747823: @BeniCheni We could have both 😂

PR #54, opened 2017-08-30
Link component

### LINKED ISSUE #39: Link component

BeniCheni:
This PR implements the `Link` component, in interest to resolve #39.

###### Sample Story - Link open self
```
<Link href='https://www.priceline.com/home/' target='_self' color='blue'>
  Open the Priceline Home in the same window
</Link>
```
link-open-self


hakimelek:
Wow! Amazing PR! Looks good to me!

.storybook/Link.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { withInfo } from '@storybook/addon-info'
+import { Link } from '../src'
+
+const description = 'A type of the <Text> component' +
```
jxnblk: I'd prefer using the backpack syntax here rather than concatenating strings

src/Link.js
```diff
@@ -0,0 +1,27 @@
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import Text from './Text'
+
+const anchorText = Text.withComponent('a')
```
jxnblk: Since the `<a>` tag generally inherits all the typographic styles of its parent, I don't think this needs to extend the `Text` component – i.e. I think this PR should be as simple and as small a change as possible

src/Link.js
```diff
@@ -0,0 +1,27 @@
...
+Link.defaultProps = {
+  cursor: 'pointer',
+  textDecoration: 'none',
+  display: 'inline-block'
+}
+
+Link.propTypes = {
+  cursor: PropTypes.string,
+  textDecoration: PropTypes.string,
+  display: PropTypes.string
```
jxnblk: I wouldn't add these as props, if someone needs to extend the `Link` component in their app, they can handle that on their own. Are there parts of the design mocks where you see a need for things like this?

src/index.js
```diff
@@ -1,10 +1,11 @@
...
 export { default as Box } from './Box'
 export { default as Flex } from './Flex'
-export { default as Image } from './Image'
-export { default as BackgroundImage } from './BackgroundImage'
-export { default as Icon } from './Icon'
 export { default as Heading } from './Heading'
+export { default as Icon } from './Icon'
+export { default as Image } from './Image'
+export { default as Link } from './Link'
+export { default as Text } from './Text'
```
jxnblk: Alphabetical order? Might make merge conflicts here a little easier to manage 👍

.storybook/Link.js
jxnblk: that was a typo, "backtick"

BeniCheni: Actually I am a fan of string literal too. Updated to use backtick. Thx.

src/Link.js
BeniCheni: Good point about `<a>` tag inherits parent's styles. Rewrote to simply the `Link` to just a simple styled anchor with text decoration none.

My reasoning is that most of the designs seem to use links without any underlines or any other type of text decoration. If using of the `Link` wants a different text decorator, the user can style it via the containing element and let the style be inherited by `Link`.

src/Link.js
BeniCheni: Took a step back and re-thought about the designs using `Link`. Think it'd be safer to have the host app define these specific styles when using `Link`. No need to maintain the props here to support specific styles.

Retired these props' code to simple further. Please review again at your convenience. 👍

src/index.js
BeniCheni: Yes for Alphabetical order if it feels right to the contributors. Definitely less risk for merge conflicts & potentially less code diff as well.

src/Link.js
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components'
+
+const Link = styled.a`
+  text-decoration: none;
```
jxnblk: Should this have a hover state? Maybe check with @chrisdroukas

BeniCheni: Update the `Link` to have default style with point cursor for the time being. Kicked off a convo with @chrisdroukas about what the style should be default hover state.

Either @chrisdroukas or I can update this default hover style, once @chrisdroukas did research to finalize his decision.

jxnblk: I think an obvious one for now could be to set `text-decoration: underline` on hover. As for `cursor: pointer` that's the default for `<a>` so you don't need to set it explicitly here

BeniCheni: Updated per suggested. Thanks for helping to figure out what the default style of the `Link` could be for now.

jxnblk: Nice! The only other thing is, should this use the default blue color from the theme?

BeniCheni: Theme's blue makes sense to align with the PCLN branch color. Updated the default color of `Link` component. Thanks~ ✌️

src/Link.js
```diff
@@ -1,7 +1,9 @@
 import styled from 'styled-components'
+import { colors } from './theme'
 
 const Link = styled.a`
   text-decoration: none;
+  color: ${colors['blue']};
```
jxnblk: Instead of importing the theme module, you should use the theme context. This line would be written as `color: ${props => props.theme.colors.blue};`

You can read more about theming in the docs: https://www.styled-components.com/docs/advanced#theming

BeniCheni: Oh yeah, theme's colors are available from theme provider.

Updated to the blue from theme provider with a conditional CSS rule.
```
${props => props.theme ? `color: ${props.theme.colors.blue};` : null}
```

The thought is that if client opts out or forgets to use theme provider, then there wouldn't be any odd CSS rule key with empty string value such as: `color:` in the browser inspection. Plz review again at your next convenience.

jxnblk:
LGTM 🎉

PR #56, opened 2017-08-30
Layout examples docs

jxnblk:
Adds some example stories and a doc on how to use the Box and Flex components for different layouts


jxnblk:
@BeniCheni Yeah, per my comment on #51 we'll also need to add docs for the breakpoint aliases. I might jump in and add that later today

PR #60, opened 2017-09-12
fix(proptype): added string in the icon size proptype

codeflyer:
This PR allow to pass the size as a string in the Icon component


src/__tests__/__snapshots__/Flex.js.snap
```diff
@@ -23,9 +23,9 @@ exports[`Flex justify prop 1`] = `
   display: -webkit-flex;
   display: -ms-flexbox;
   display: flex;
-  -webkit-box-pack: space-between;
+  -webkit-box-pack: justify;
   -webkit-justify-content: space-between;
-  -ms-flex-pack: space-between;
```
jxnblk: I guess the autoprefixer got smarter

jxnblk:
LGTM

PR #62, opened 2017-09-15
Removed usage of Object.assign in favor of object spread

byrekt:
Converted the Object.assign usages with usages of the spread operator. This should prevent some issues with IE11 since this will be transpiled by babel

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-design-system",
-  "version": "1.0.0-7",
+  "version": "1.0.0-6",
   "description": "Priceline Design System",
   "main": "dist/index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-design-system",
-  "version": "1.0.0-6",
+  "version": "1.0.0-7",
   "description": "Priceline Design System",
   "main": "dist/index.js",
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pcln-design-system",
-  "version": "1.0.0-6",
```
jxnblk: For PRs we shouldn't touch the version. This PR will probably be merged before others, but in some cases, other PRs could get merged in and this would be out of sync.

Currently we bump the version after merging to master, and do a publish immediately. Once we move out of the beta versions, we'll use `npm version` to manage versions and add git tags.

src/legacyTheme.js
```diff
@@ -71,8 +71,6 @@ export const colors = {
   orange3
 }
 
-const legacyTheme = Object.assign({}, theme, {
-  colors
-})
+const legacyTheme = {...theme, colors: {...colors}}
```
jxnblk: This can be written as `{ ...theme, { colors } }` which would be one less call to Babel's `extends` function

src/theme.js
```diff
@@ -86,7 +89,7 @@ const reds = tints.map(tint(red))
 const oranges = tints.map(tint(orange))
 const purples = tints.map(tint(purple))
 
-export const colors = Object.assign({}, {
+export const colors = {...{
```
jxnblk: I don't think this needs wrapping since it's not cloning another object

byrekt: updated

jxnblk:
Awesome! LGTM :shipit:

PR #63, opened 2017-09-19
Adds support for responsive alignment

chrisdroukas:


src/Text.js
```diff
@@ -1,8 +1,8 @@
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
-import { fontSize, space, color } from 'styled-system'
+import { fontSize, space, color, responsiveStyle } from 'styled-system'
 
-export const align = props => props.align ? { textAlign: props.align } : null
+// export const align = props => props.align ? { textAlign: props.align } : null
```
jxnblk: ✂️

src/Text.js
```diff
@@ -13,13 +13,15 @@ export const caps = props => props.caps
 
 export const bold = props => props.bold ? { fontWeight: props.theme.bold } : null
 
+const align = responsiveStyle('text-align', 'align')
```
jxnblk: You *could* DRY this up by moving this into a common utility module, but I think this is fine for now

package.json
```diff
@@ -53,6 +53,7 @@
     "react-test-renderer": "^15.6.1",
     "standard": "^10.0.2",
     "styled-components": "^2.1.1",
+    "stylis": "^3.2.16",
```
jxnblk: Just a note for context here: we were having snapshot issues with an older version of this being installed due to styled-components' specified version range, and this seemed like the simplest solution to stop the snapshot's from failing on Travis

Issue #44, opened 2017-08-24
Responsive hide utility component


jxnblk:
Add a UI component that hides content per breakpoint. I think the media queries in this sort of component *could* deviate from the mobile-first min-width approach to target specific breakpoints, but still use the `theme.breakpoints` array

PR #64, opened 2017-09-19
Simple responsive visibility toggle on `Box`

chrisdroukas:


src/Box.js
```diff
@@ -1,11 +1,14 @@
 import styled from 'styled-components'
-import { space, width, color } from 'styled-system'
+import { space, width, color, responsiveStyle } from 'styled-system'
 import PropTypes from 'prop-types'
 
+const visibility = responsiveStyle('visibility', 'visibility')
```
jxnblk: I'm guessing we'll actually want to use `display: none` for this since `visibility: hidden` keeps the element in the document flow (and still takes up space). The other consideration here is that the `responsiveStyle` function only supports mobile-first min-width media queries, whereas we might actually want to only hide between two breakpoints.

src/Box.js
```diff
@@ -1,11 +1,14 @@
...
+
 const Box = styled.div`
   ${space}
   ${width}
   ${color}
+  ${visibility}
```
jxnblk: I think rather than adding this to the Box component, which is intended to be used in a **lot** of situations, that this should be an **extension** of the Box component since responsively hiding elements is not as common – e.g.:

```jsx
const Hide = styled(Box)`
  ${responsiveHideUtility}
`
```

jxnblk:
Let me know if you wanna pair on this or if you need anything from me

jxnblk:
I jumped in to fix merge conflicts here and changed the API to add a `<Hide />` component (fixes #44) that sets display none at specific breakpoints. Would love to get everyone's thoughts on this approach

src/__tests__/Hide.js
```diff
@@ -0,0 +1,37 @@
...
+    const json = renderer.create(<Hide lg theme={theme} />).toJSON()
+    expect(json).toMatchSnapshot()
+  })
+
+  test('renders with xl prop', () => {
+    const json = renderer.create(<Hide xl theme={theme} />).toJSON()
+    expect(json).toMatchSnapshot()
+  })
+
+})
```
BeniCheni: Super trivial, `Standard` picked up a warning on the empty new line on line 36:
```
/priceline-web/design-system/src/__tests__/Hide.js:37:1: Block must not be padded by blank lines.
```

Issue #73, opened 2017-09-26
WebP Support For BackgroundImage Component


BeniCheni:

BeniCheni:
Issue no longer is in scope.

PR #74, opened 2017-09-26
Implement WebP Support For BackgroundImage Component

BeniCheni:
Per #73 , there's a client app using `BackgroundImage` with WebP images for performance optimization.

In interest to support WebP, this PR is to enhance the `Background` component to check browser support for WebP and generate background image accordingly.

###### Sample Story With WebP Support With Chrome - `imageWebP` prop got rendered as a WebP Image
screen shot 2017-09-25 at 9 56 21 pm

##### Sample Story With WebP Support With FireFox - the fall-back `image` prop got rendered
screen shot 2017-09-25 at 9 56 37 pm

##### Unit Test Status
```
design-system> npm test -- -u

> pcln-design-system@1.0.0-8 test /Users/bchen/priceline-web/design-system
> jest "-u"

 PASS  src/__tests__/BackgroundImage.js
 PASS  src/__tests__/Icon.js
 PASS  src/__tests__/Text.js
 PASS  src/__tests__/Heading.js
 PASS  src/__tests__/Button.js
 PASS  src/__tests__/Box.js
 PASS  src/__tests__/Flex.js
 PASS  src/__tests__/RedButton.js
 PASS  src/__tests__/ThemeProvider.js
 PASS  src/__tests__/Link.js
 PASS  src/__tests__/Image.js
 PASS  src/__tests__/GreenButton.js
 PASS  src/__tests__/OutlineButton.js
 PASS  src/__tests__/theme.js

Test Suites: 14 passed, 14 total
Tests:       94 passed, 94 total
Snapshots:   93 passed, 93 total
Time:        4.421s
Ran all test suites.

=============================== Coverage summary ===============================
Statements   : 100% ( 160/160 )
Branches     : 100% ( 20/20 )
Functions    : 100% ( 23/23 )
Lines        : 100% ( 150/150 )
================================================================================
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,8 @@
     ],
     "collectCoverage": true,
     "coveragePathIgnorePatterns": [
-      "dist/"
+      "dist/",
+      "src/helpers/webp-support.js"
     ]
   },
   "dependencies": {
--- a/src/BackgroundImage.js
+++ b/src/BackgroundImage.js
@@ -7,7 +7,7 @@ import {
 
 const image = props => {
   /* istanbul ignore next */
-  const isWebP = props.webP && detectWebPSupport()
+  const isWebP = props.useWebP && detectWebPSupport()
   /* istanbul ignore next */
   const finalImage = isWebP
     ? props.imageWebP : props.image
@@ -35,11 +35,11 @@ BackgroundImage.propTypes = {
   /** background-image url */
   image: PropTypes.string,
   imageWebP: PropTypes.string,
-  webP: PropTypes.bool
+  useWebP: PropTypes.bool
 }
 
 BackgroundImage.defaultProps = {
-  webP: false
+  useWebP: false
 }
 
 export default BackgroundImage
--- a/src/BackgroundImage.js
+++ b/src/BackgroundImage.js
@@ -5,17 +5,12 @@ import {
   detectWebPSupport
 } from './helpers'
 
-const image = props => {
+const image = props => (
   /* istanbul ignore next */
-  const isWebP = props.useWebP && detectWebPSupport()
-  /* istanbul ignore next */
-  const finalImage = isWebP
-    ? props.imageWebP : props.image
-
-  return finalImage
-    ? { backgroundImage: `url(${finalImage})` }
-    : null
-}
+  (props.imageWebP && detectWebPSupport())
+    ? { backgroundImage: `url(${props.imageWebP})` }
+    : { backgroundImage: `url(${props.image})` }
+)
 
 const height = props => props.height
   ? { height: props.height }
@@ -34,12 +29,7 @@ BackgroundImage.displayName = 'BackgroundImage'
 BackgroundImage.propTypes = {
   /** background-image url */
   image: PropTypes.string,
-  imageWebP: PropTypes.string,
-  useWebP: PropTypes.bool
-}
-
-BackgroundImage.defaultProps = {
-  useWebP: false
+  imageWebP: PropTypes.string
 }
 
 export default BackgroundImage
--- a/src/BackgroundImage.js
+++ b/src/BackgroundImage.js
@@ -1,9 +1,21 @@
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
 
-const image = props => props.image
-  ? { backgroundImage: `url(${props.image})` }
-  : null
+import {
+  detectWebPSupport
+} from './helpers'
+
+const image = props => {
+  /* istanbul ignore next */
+  const isWebP = props.webP && detectWebPSupport()
+  /* istanbul ignore next */
+  const finalImage = isWebP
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -62,7 +62,8 @@
     ],
     "collectCoverage": true,
     "coveragePathIgnorePatterns": [
-      "dist/"
+      "dist/",
+      "src/helpers/webp-support.js"
```
BeniCheni: webp-support.js file is ignored for coverage, because WebP detect uses a canvas element to detect support and jsdom in Jest test cycle doesn’t support canvas DOM element.

src/BackgroundImage.js
```diff
@@ -1,9 +1,21 @@
...
-  ? { backgroundImage: `url(${props.image})` }
-  : null
+import {
+  detectWebPSupport
+} from './helpers'
+
+const image = props => {
+  /* istanbul ignore next */
+  const isWebP = props.useWebP && detectWebPSupport()
+  /* istanbul ignore next */
```
BeniCheni: These Istanbul ignore statements support the bypassing of the WebP Detection in Instanbul test coverage.

hakimelek:
Should we just use the image with WebP support only? I don't know if we want people to use `BackgroundImage` without webp support.

chrisdroukas:
I agree, @hakimelek — the `useWebP` flag seems redundant. It would be sufficient to just supply the `webP` image and determine browser support automatically.

In other words, if a `webP` asset is supplied and the browser supports it, show it. Otherwise, fallback.

cc @BeniCheni

BeniCheni:
@chrisdroukas & @hakimelek , thanks for your feedback. Think I can retire the `useWebP` boolean prop, and just use the `imageWebP` as first choice of image source, if WebP image source is empty string, fall back to use the `image` prop with regular image format.

chrisdroukas:
@BeniCheni would something like this work?

```
const image = props => {
  /* istanbul ignore next */
  const supportsWebP = detectWebPSupport()
  /* istanbul ignore next */
  if (props.imageWebP && supportsWebP) {
    return { backgroundImage: `url(${props.imageWebP})` }
  }
  else {
    return { backgroundImage: `url(${props.image})` }
  }
}
```

BeniCheni:
@chrisdroukas , retired the previous `useWebP` prop to simplify the code. Please review again.

src/helpers/webp-support.js
```diff
@@ -0,0 +1,8 @@
+const canvasMock = typeof document === 'object' ? document.createElement('canvas') : {}
+canvasMock.width = canvasMock.height = 1
+
+// If canvas supports webp, the returned Data URLs would  be prefixed with "data:".
+// Thus data starts at index 5, after the prefix.
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs
+export const detectWebPSupport = () => canvasMock.toDataURL
+  ? canvasMock.toDataURL('image/webp').indexOf('image/webp') === 5 : false
```
jxnblk: Thanks for putting this PR together, but I don't think something like this belongs in the design-system repo since it doesn't actually have anything to do with the CSS generated for the BackgroundImage component, but affects the `props.image` *value*.

This is a great example of where using the higher-order component (HOC) pattern would make a lot more sense and be more reusable without adding logic to a base-level style component. We could start another repo with utility HOCs if we end up making a lot of things like this, but I'd try to keep client-side logic out of this one.

An example HOC could be used like this:

```jsx
const WebPBackgroundImage = withWebP(BackgroundImage)

<WebPBackgroundImage
  fallback='path-to/fallback.jpg'
  image='path-to/image.webp'
/>
```

The HOC itself could look something like this:

```jsx
const withWebP = Component => {
  const WebPComponent = ({ image, fallback, ...rest }) => {
    const imageOrFallback = hasWebP ? image : fallback
    return <Component image={imageOrFallback} />
  }
  return WebPComponent
}
```

src/__tests__/__snapshots__/Flex.js.snap
```diff
@@ -23,9 +23,9 @@ exports[`Flex justify prop 1`] = `
   display: -webkit-flex;
   display: -ms-flexbox;
   display: flex;
-  -webkit-box-pack: justify;
+  -webkit-box-pack: space-between;
   -webkit-justify-content: space-between;
-  -ms-flex-pack: justify;
+  -ms-flex-pack: space-between;
```
jxnblk: I believe this is the same snapshot/dependency issue that @chrisdroukas was running into. Make sure you `rm -rf node_modules && npm i && npm t -- -u` here

src/helpers/webp-support.js
BeniCheni: Ok cool. Good to know the scope of WebP support can try out HOC pattern. Will close this PR. Thanks for your feedback.

PR #75, opened 2017-09-26
Add Troubleshooting section in Contributing docs

hakimelek:
Should add instructions for how to overcome the `ERROR in ./icons.json` when running storybook.


jxnblk:
Awesome! We should probably also make the icon build process a little more automatic

PR #77, opened 2017-09-27
Add dark colors to the theme

### LINKED ISSUE #49: Add darker color to the color palette

hakimelek:
This change removes tints colors from the theme and replace them with 3 versions of the color:
- brighten version of the color
- actual theme color
- darker version of the color

resolves #49


src/theme.js
```diff
@@ -76,18 +76,12 @@ const flatten = (name, colors) => colors
...
-  0.6,
-  0.8
-]
-const blues = tints.map(tint(blue))
-const grays = tints.map(tint(gray))
-const greens = tints.map(tint(green))
-const reds = tints.map(tint(red))
-const oranges = tints.map(tint(orange))
-const purples = tints.map(tint(purple))
+const blues = [brighten(blue), blue, darken(blue)]
```
jxnblk: It sounded like @oliverdumoulin wasn't using a color function for his updated color scale anymore, so I think we'll want to remove the chroma-js dependency and just hard code the values in

hakimelek: @jxnblk We are depending on the tint function (depending on chroma) in the legacy theme. Should we hard code those values as well?

jxnblk: Yeah, I think that's fine – probably worth getting rid of that dependency if we can

hakimelek:
Closing this in favor of #79.

Issue #43, opened 2017-08-23
Update legacy theme colors to match the old colors


hakimelek:

jxnblk:
Just adding a note here for posterity. As we discussed in person, I think we'll try to stick to the new naming convention to support the ability to switch to the new color palette when a project is ready to move over

PR #79, opened 2017-09-28
Modified palette and created light and dark color values.

odmln:


hakimelek:
It looks like your tests are failing. You can run `npm run test` to figure out what tests are breaking. Also, I think we will need to hard code the colors on the `lagecyTheme.js` file and get rid of `chroma.js`.

hakimelek:
Should resolve #43, #49

src/__tests__/__snapshots__/theme.js.snap
```diff
@@ -12,66 +12,67 @@ Object {
...
+    "lightGray": "#f0f2f4",
+    "lightGreen": "#cec",
+    "lightOrange": "#fec",
+    "lightPurple": "#ecf",
+    "lightRed": "#fcc",
     "orange": "#f90",
-    "orange0": "#ffebcc",
-    "orange1": "#ffd699",
-    "orange2": "#ffc266",
-    "orange3": "#ffad33",
```
jxnblk: This will probably be a "breaking" change for anyone using a `3` color. Should we map these to `2` and add a prop-type warning that it will be deprecated?

src/__tests__/__snapshots__/theme.js.snap
```diff
@@ -12,66 +12,67 @@ Object {
...
+      "#cdf",
+      "#007aff",
+      "#049",
     ],
-    "gray": "#596b7d",
-    "gray0": "#dee1e5",
-    "gray1": "#bdc4cb",
-    "gray2": "#9ba6b1",
-    "gray3": "#7a8997",
+    "darkBlue": "#049",
```
jxnblk: Are the dark colors not defined as a numbered color? Would be good to have a scale for the entire range of each hue, then add shorthands for `darkBlue` and `lightBlue` IMO

jxnblk: Disregard this. Too early and commenting on snapshots 😬

src/theme.js
```diff
@@ -82,23 +92,35 @@ const tints = [
...
-const greens = tints.map(tint(green))
-const reds = tints.map(tint(red))
-const oranges = tints.map(tint(orange))
-const purples = tints.map(tint(purple))
+const blues = [lightBlue, blue, darkBlue]
+const grays = [lightGray, gray, darkGray]
+const greens = [lightGreen, green, darkGreen]
+const reds = [lightRed, red, darkRed]
+const oranges = [lightOrange, orange, darkOrange]
+const purples = [lightPurple, purple, darkPurple]
```
jxnblk: As in the comment on the snapshot, I think these arrays should maybe have an extra duplicate color to  handle the case where someone might be using `purple3`, e.g. `[lightPurple, purple, darkPurple, purple]` then we can add a prop type warning, and remove the duplicate with a `1.0.0` release. This would ensure that the component still gets the color, just maybe not the 100% right one. Curious what others think though

hakimelek: I don't know if anyone is using color3 yet. But since this is public, we should not take the risk of just removing that. I think that doing that would ensure backward compatibility for now. I will work on it.

jxnblk: 👌

.storybook/Colors.js
```diff
@@ -12,34 +12,33 @@ const legacy = Object.keys(legacyTheme.colors)
   .map(key => ({ key, value: legacyTheme.colors[key] }))
   .filter(color => !Array.isArray(color.value))
 
-const Chip = props => <Box width={1} px={5} py={4} bg={props.color} />
+const Chip = props => props.color ?
```
BeniCheni: When checking with `Standard` linter, this line is being warned:
```
/design-system/.storybook/Colors.js:15:36: '?' should be placed at the beginning of the line.
```

BeniCheni: Additional linter warnings also picked up these, which code being linted may be outside of the scope of this PR. But thought that I'd share, if you'd like to simply update the code format along in this PR:
```
/design-system/.storybook/Colors.js:3:8: 'styled' is defined but never used.
/design-system/.storybook/Colors.js:22:19: Infix operators must be spaced.
/design-system/.storybook/Colors.js:22:24: Infix operators must be spaced.
/design-system/.storybook/Colors.js:22:29: Infix operators must be spaced.
/design-system/.storybook/Colors.js:33:14: Infix operators must be spaced.
/design-system/.storybook/Colors.js:33:19: Infix operators must be spaced.
/design-system/.storybook/Colors.js:33:24: Infix operators must be spaced.
/design-system/.storybook/Colors.js:84:5: Newline required at end of file but not found.
```

hakimelek: Will make sure to run the linter before I push!

jxnblk: Should probably go ahead and set up [prettier](https://github.com/prettier/prettier) in this repo. Any objections to that?

.storybook/Colors.js
```diff
@@ -48,34 +47,38 @@ storiesOf('Color', module)
...
+        ))}
+      </Flex>
     </div>
   ))
-  .add('Old vs. New', () => (
+  .add('New vs. Old', () => (
     <div>
       <Box p={3}>
-        <h1>Old vs. New</h1>
+        <h1>New vs. Old</h1>
```
jxnblk: 👍

.storybook/Colors.js
BeniCheni: Prettier tool would be great. 👍🏻

jxnblk:
:shipit: Merging this in

PR #82, opened 2017-09-29
feat(badge): added badge and test

Degron:
Added badge. 
screen shot 2017-09-29 at 2 53 03 pm
screen shot 2017-09-29 at 2 53 13 pm
screen shot 2017-09-29 at 2 53 23 pm

### DIFF
--- a/.storybook/Badge.js
+++ b/.storybook/Badge.js
@@ -45,6 +45,6 @@ storiesOf('Badge', module)
   ))
   .add('attention-light', () => (
     <Badge type='attention-light'>
-      ATTENTION
+      ATTENTION 
     </Badge>
   ))
--- a/.storybook/Badge.js
+++ b/.storybook/Badge.js
@@ -2,7 +2,6 @@ import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { withInfo } from '@storybook/addon-info'
 import { Badge } from '../src'
-import { YelowBadge } from '../src'
 
 storiesOf('Badge', module)
   .add('Badge component', withInfo({
--- a/.storybook/Badge.js
+++ b/.storybook/Badge.js
@@ -1,50 +1,55 @@
 import React from 'react'
 import { storiesOf } from '@storybook/react'
+import { withInfo } from '@storybook/addon-info'
 import { Badge } from '../src'
+import { YelowBadge } from '../src'
 
 storiesOf('Badge', module)
-  .add('default', () => (
-    <Badge>
-      DEFAULT
+  .add('Typography component', withInfo({
+    inline: true,
+    text: 'Use the <Badge /> component to render a primitive badge.'
+  })(() => (
+    <Badge bg='blue'>
+      INFORMATION
     </Badge>
-  ))
-  .add('information', () => (
-    <Badge type='information'>
+  )))
+  .add('blue', () => (
+    <Badge bg='blue'>
       INFORMATION
     </Badge>
   ))
-  .add('information-light', () => (
-    <Badge type='information-light'>
+  .add('blue-light', () => (
+    <Badge bg='blue-light'>
       INFORMATION
     </Badge>
   ))
-  .add('success', () => (
-    <Badge type='success'>
+  .add('green', () => (
+    <Badge bg='green'>
       SUCCESS
     </Badge>
   ))
-  .add('success-light', () => (
-    <Badge type='success-light'>
+  .add('green-light', () => (
+    <Badge bg='green-light'>
       SUCCESS
     </Badge>
   ))
-  .add('warning', () => (
-    <Badge type='warning'>
+  .add('red', () => (
+    <Badge bg='red'>
       WARNING
     </Badge>
   ))
-  .add('warning-light', () => (
-    <Badge type='warning-light'>
+  .add('red-light', () => (
+    <Badge bg='red-light'>
       WARNING
     </Badge>
   ))
-  .add('attention', () => (
-    <Badge type='attention'>
+  .add('orange', () => (
+    <Badge bg='orange'>
       ATTENTION
     </Badge>
   ))
-  .add('attention-light', () => (
-    <Badge type='attention-light'>
+  .add('orange-light', () => (
+    <Badge bg='orange-light'>
       ATTENTION 
     </Badge>
   ))
--- a/.storybook/Badge.js
+++ b/.storybook/Badge.js
@@ -1,0 +1,50 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Badge } from '../src'
+
+storiesOf('Badge', module)
+  .add('defoult', () => (
+    <Badge>
+      DEFOULT
+    </Badge>
+  ))
+  .add('informaton', () => (
+    <Badge type='information'>
+      INFORMATION
+    </Badge>
+  ))
+  .add('informaton-light', () => (
+    <Badge type='information-light'>
+      INFORMATION
+    </Badge>
+  ))
+  .add('success', () => (
+    <Badge type='success'>
+      SUCCESS
+    </Badge>
+  ))
+  .add('success-light', () => (
+    <Badge type='success-light'>
+      SUCCESS
+    </Badge>
+  ))
+  .add('warning', () => (
+    <Badge type='warning'>
+      WARNING
+    </Badge>
+  ))
+  .add('warning-light', () => (
+    <Badge type='warning-light'>
+      WARNING
+    </Badge>
+  ))
+  .add('attention', () => (
+    <Badge type='attention'>
+      ATTENTION
+    </Badge>
+  ))
+  .add('attention-light', () => (
+    <Badge type='attention-light'>
+      ATTENTION
+    </Badge>
+  ))
--- a/.storybook/Badge.js
+++ b/.storybook/Badge.js
@@ -3,17 +3,17 @@ import { storiesOf } from '@storybook/react'
 import { Badge } from '../src'
 
 storiesOf('Badge', module)
-  .add('defoult', () => (
+  .add('default', () => (
     <Badge>
-      DEFOULT
+      DEFAULT
     </Badge>
   ))
-  .add('informaton', () => (
+  .add('information', () => (
     <Badge type='information'>
       INFORMATION
     </Badge>
   ))
-  .add('informaton-light', () => (
+  .add('information-light', () => (
     <Badge type='information-light'>
       INFORMATION
     </Badge>
--- a/.storybook/Badge.js
+++ b/.storybook/Badge.js
@@ -1,10 +1,11 @@
+import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { withInfo } from '@storybook/addon-info'
 import { Badge } from '../src'
 import { YelowBadge } from '../src'
 
 storiesOf('Badge', module)
-  .add('Typography component', withInfo({
+  .add('Badge component', withInfo({
     inline: true,
     text: 'Use the <Badge /> component to render a primitive badge.'
   })(() => (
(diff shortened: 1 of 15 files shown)


.storybook/Badge.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Badge } from '../src'
+
+storiesOf('Badge', module)
+  .add('defoult', () => (
+    <Badge>
+      DEFOULT
```
msafari: (non-clocking): Typo here. `DEFAULT`

src/Badge.js
```diff
@@ -0,0 +1,79 @@
...
+      return {
+        backgroundColor: theme.colors.blue,
+        color: theme.colors.white
+      }
+    case 'information-light':
+      return {
+        backgroundColor: theme.colors.blue0,
+        color: theme.colors.black
+      }
+    default:
```
msafari: Good idea to have `information` style for default 👍

src/Badge.js
```diff
@@ -0,0 +1,79 @@
...
+  text-align: center;
+  font-size: 12px;
+  font-weight: 600;
+  padding: 0 8px;
+  ${type}
+  ${space}
+`
+
+
+Badge.propTypes = {
```
msafari: @byrekt unrelated to this PR. How do you feel about putting `propTypes` on the very top of each file? I think it's more clear self-documentation if it's all the way on top.

src/Badge.js
```diff
@@ -0,0 +1,79 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { space } from 'styled-system'
+import { theme } from '../'
+
+const type = props => {
+  switch(props.type) {
```
jxnblk: We should probably use the styled-system `color` function here, but would love to hear some thoughts. Eg. `<Badge bg='blue' />` and `<Badge bg='green' />` seems more readable in code to me and it would be consistent with other components like `Box`. If we wanted non-presentational aliases, those *could* be added in the theme, but don't think they're really necessary

src/Badge.js
```diff
@@ -0,0 +1,79 @@
...
+        color: theme.colors.white
+      }
+  }
+}
+
+
+const Badge = styled.div`
+  height: 24px;
+  line-height:24px;
+  border-radius: 100px;
```
jxnblk: Setting `border-radius: 99999px` is better for things like circles and pill shapes

src/Badge.js
```diff
@@ -0,0 +1,79 @@
...
+
+
+const Badge = styled.div`
+  height: 24px;
+  line-height:24px;
+  border-radius: 100px; 
+  display: inline-block;
+  vertical-align: middle;
+  text-align: center;
+  font-size: 12px;
```
jxnblk: Values like these should be from the theme, e.g. `{props => props.theme.fontSizes[0]}`

src/Badge.js
```diff
@@ -0,0 +1,79 @@
...
+  font-weight: 600;
+  padding: 0 8px;
```
jxnblk: This should be the set using styled-system `space` and `px` set as a defaultProp

src/Badge.js
msafari: So @Degron actually brought that up to me. I wasn't sure if we want to approach it that way or not. I suggested taking this approach because there seem to be only certain types of badges so we ensure they have the right color and background based on their type. Unless it's not a good idea to enforce it or it's not meant to do it.

src/Badge.js
byrekt: @msafari On my team there has been some debate. Some people on RC felt like (especially when the prop-types were documented with inline documentation for storybook) it got in the way, so we allow people to put the propTypes wherever they want. I prefer them at the top as well.

src/Badge.js
```diff
@@ -0,0 +1,88 @@
...
+      return {
+        backgroundColor: theme.colors.blue,
+        color: theme.colors.white
+      }
+    case 'information-light':
+      return {
+        backgroundColor: theme.colors.blue0,
+        color: theme.colors.black
+      }
+    default:
```
747823: What about mapping the type prop to an object key that points to a set of background color and color instead?

E.g.
```
const badgeColors = {
  'success': {
    backgroundColor: theme.colors.green,
    color: theme.colors.white
  }
}
```

Then in the function
```
return badgeColors[props.type]
```

Or even making each badge color a styled component that extends Badge, and then map them to keys after that. It will save a tiny bit of computing time.

src/Badge.js
```diff
@@ -0,0 +1,86 @@
...
+        color: theme.colors.black
+      }
+    case 'information':
+      return {
+        backgroundColor: theme.colors.blue,
+        color: theme.colors.white
+      }
+    case 'information-light':
+      return {
+        backgroundColor: theme.colors.blue0,
```
BeniCheni: Can `green0`, `red0`, `orange0` and `blue0 ` from legacy theme be replaced to use similar colors from new theme? If the namespaces don't exist yet in new theme, maybe we can refactor the same colors from legacy them to the new theme?

src/Badge.js
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const numberStringOrArray = PropTypes.oneOfType([
+  PropTypes.number,
+  PropTypes.string,
+  PropTypes.array
+])
+
+Badge.propTypes = {
+  type: PropTypes.string,
```
BeniCheni: Would like to understand the your thoughts behind `type` prop a little bit more. But currently, the `type` prop is tied to keywords such as `success`, or `warning`, to determine background color and color of the label inside the badge.

Found these keywords somewhat confusing. If my use case for the `Badge` is not related to any severity level, but instead, let's say, my website is about food and my UI changes the badge's background color and color of the label by type of cuisine. In such use case, type of cuisine does not fit the current keywords such as `success` and etc.

Perhaps we can just simplify the value of types to use numeric, i.e.:

- `<Badge type={1} />` => light blue background with black label

- `<Badge type={2} />` => dark blue background with white label

and so forth.

It'd be great if this PR can also update the documentation space to provide a README for `Badge` that explain the `type` prop design with use cases.

###### Documentation Space
https://github.com/pricelinelabs/design-system/tree/master/docs

###### README page with links to documentation topics
https://github.com/pricelinelabs/design-system/blob/master/README.md

As your 1st `Badge` story, you can also consider a self-documentation self with your code using `withInfo(...)`.
###### Story Code
https://github.com/pricelinelabs/design-system/blob/master/.storybook/Heading.js#L11

###### Heading Documentation Generated By `withInfo(...)` story.
https://pricelinelabs.github.io/design-system/?selectedKind=Heading&selectedStory=Heading%20component&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel

src/Badge.js
jxnblk: My thinking is that for communication between designers and developers, it's generally easier to refer to is as "the blue badge" instead of "the information badge" – e.g. presentational semantics over functional. Generally I don't think people will use the wrong color, and it should be up to the designers to ensure that color is used as intended across the UI

src/Badge.js
msafari: @Degron Looks like you refactored this `type` out and are using `bg`. Please remove the old prop from prop-types and add `bg` instead.

src/Badge.js
```diff
@@ -0,0 +1,76 @@
...
+  border-radius: 99999px; 
+  display: inline-block;
+  font-size: ${props => props.theme.fontSizes[0]}px;
+  font-weight: 600;
+  ${type}
+  ${space}
+`
+
+Badge.defaultProps = {
+  px:2,
```
msafari: 👍

msafari: really (non-blocking): please add a space after `:`

src/Badge.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { space } from 'styled-system'
+import { theme } from '../'
+
```
BeniCheni: Trivial, believe `Standard` linting picks up a warning for the extra new-line space in lines 6-7. Please consider updating code format?
screen shot 2017-10-02 at 8 42 59 pm

src/Badge.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
```
BeniCheni: Another linting warning found with unused `React` import. Please consider removing unused import?
screen shot 2017-10-02 at 8 44 20 pm

src/Badge.js
```diff
@@ -0,0 +1,76 @@
...
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { space } from 'styled-system'
+import { theme } from '../'
+
+
+function type( props ){
```
BeniCheni: Another linting warning:
screen shot 2017-10-02 at 8 46 19 pm

src/Badge.js
```diff
@@ -0,0 +1,73 @@
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { space } from 'styled-system'
+import { theme } from '../'
+
+function type (props) {
+  const badgeColors = {
+    'green': {
+      backgroundColor: theme.colors.green,
```
jxnblk: Sorry for the delay on this – we shouldn't import the `theme` object into the styled components, but instead get values from `props.theme` which is handled through [React context](https://reactjs.org/docs/context.html) and styled-components' [ThemeProvider](https://www.styled-components.com/docs/advanced#theming) – this allows us better control over the theme and allows things like the design-system's `legacy` switch to control switching between the old and new color palettes

src/Badge.js
```diff
@@ -0,0 +1,73 @@
...
+      color: theme.colors.white
+    },
+    'green-light': {
```
jxnblk: Would be good to follow the same naming convention used in #79 – e.g. [`lightGreen`](https://github.com/pricelinelabs/design-system/pull/79/files#diff-e179cfde519bf6bae73373ad10d6ae2fR45)

scripts/parse-icons.js
```diff
@@ -48,7 +48,7 @@ const getPath = nodes => nodes.children
 
 const getViewBox = nodes => nodes.properties.viewBox
 
-const removePrefix = str => str.replace(/^ic\-/, '')
+const removePrefix = str => str.replace(/^ic\-/, '') // eslint-disable-line no-useless-escape
```
jxnblk: Shouldn't the regex just be updated rather than disabling the lint?

.storybook/Badge.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { withInfo } from '@storybook/addon-info'
+import { Badge } from '../src'
+import { YelowBadge } from '../src'
```
jxnblk: Guessing this is a typo – I don't see a `YelowBadge` module

src/Badge.js
```diff
@@ -0,0 +1,72 @@
...
+      backgroundColor: props.theme.colors.blue,
+      color: props.theme.colors.white
+    },
+    'lightBlue': {
+      backgroundColor: props.theme.colors.blue0,
+      color: props.theme.colors.black
+    }
+  }
+
+  return badgeColors[props.bg]
```
jxnblk: 👍

src/Badge.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const Badge = styled.div`
+  border-radius: 99999px; 
+  display: inline-block;
+  font-size: ${props => props.theme.fontSizes[0]}px;
+  font-weight: 600;
```
jxnblk: We'll get fontWeights in the theme at some point, but this is fine for now

scripts/parse-icons.js
hakimelek: Ideally, do you think removing the \` before the `-` should fix it?

jxnblk:
Other than the comments above about the linting and the import in storybook, this LGTM 🚀

jxnblk:
Thanks for putting this together! Going to merge this in and publish a new release today

PR #86, opened 2017-10-05
Use prettier for code formatting

jxnblk:
Prettier is a code formatter that helps negate the need for listing and making sure your code follows a particular style. It runs as a precommit hook and formats code to a consistent style

See https://github.com/prettier/prettier for more about how it works

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,8 @@
   "description": "Priceline Design System",
   "main": "dist/index.js",
   "scripts": {
-    "prepare": "npm run icons && babel src -d dist --ignore __tests__",
+    "prepublish": "npm run icons && babel src -d dist --ignore __tests__",
+    "precommit": "lint-staged",
     "start": "start-storybook -p 8000 -c .storybook",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Priceline Design System",
   "main": "dist/index.js",
   "scripts": {
-    "prepare": "npm run icons && babel src -d dist --ignore __tests__",
+    "prepublish": "npm run icons && babel src -d dist --ignore __tests__",
```
BeniCheni: For my education, what would be the reason to use “prepublish” instead of “prepare”. I was under the assumption that later NPM version favors “prepare” over “prepublish”. Would like to use any special use case for “prepublish”.

jxnblk: Yeah, we talked about this in the common components meeting, but because some people here are still using npm 3, using `prepublish` is a safer bet that the package will build as expected. If we get most people up to npm 5, then we can switch back to `prepare`

PR #92, opened 2017-10-09
Rename border color to keep color scales consistent

jxnblk:
Changes `lighterGray` to `lightGray`, and the previous `lightGray` to `borderGray` – which is mostly intended for form component borders


.storybook/Colors.js
```diff
@@ -29,14 +29,17 @@ const Card = props => (
...
-      <Box
-        w={[1/3, 1/4, 1/5]}
+      <Flex
         key={key}
-        mr={3}
-        mb={1}>
-        <Card name={key} color={theme.colors[key]} />
-        <Card name={key} color={legacyTheme.colors[key]} />
-      </Box>
+        w={[1, 1/2, null, 1/3, 1/4]}
```
hakimelek: Does styled-system support 5 ranges now?

jxnblk: It supports how ever many `breakpoints` we have defined in our theme. Right now we have four, and the first number is for the non-media-query-scoped value, e.g. "xs" or below the first breakpoint

jes708:
Caution: this will impact the [Divider PR](https://github.com/pricelinelabs/design-system/pull/94), which uses `lighterGray` in `storybook`.

PR #94, opened 2017-10-09
feat(divider) create divider component

jes708:
Creates `<hr />` / divider component


src/Divider.js
```diff
@@ -0,0 +1,44 @@
...
+
+Divider.displayName = 'Divider'
+
+const numberStringOrArray = PropTypes.oneOfType([
+  PropTypes.number,
+  PropTypes.string,
+  PropTypes.array
+])
+
+Divider.propTypes = {
```
jxnblk: I recently added a `propTypes` export for styled-system that you could use to cut down the repeated code here – https://github.com/jxnblk/styled-system#proptypes

src/__tests__/Divider.js
```diff
@@ -0,0 +1,35 @@
...
+import 'jest-styled-components'
+import { Divider, theme } from '..'
+
+describe('Divider', () => {
+  test('renders', () => {
+    const json = renderer.create(<Divider />).toJSON()
+    expect(json).toMatchSnapshot()
+  })
+
+  test('width prop sets width', () => {
```
jxnblk: You shouldn't have to write tests for `styled-system` too much since it's 100% covered

.storybook/Divider.js
```diff
@@ -0,0 +1,58 @@
...
+      width={256}
+      borderColor='lighterGray' />
+  ))
+  .add('VW Width', () => (
+    <Divider
+      p={3}
+      width='50vw'
+      borderColor='lighterGray' />
+  ))
+  .add('Directional Padding', () => (
```
jxnblk: Maybe dumb question: is there a good use-case for padding on an `<hr>`? 🤔

jxnblk:
Generally this looks great! I'd say maybe add a markdown file to `docs/` just for an explanation of what the component is useful for. Eventually we'll pull the markdown in to build some better docs

.storybook/Divider.js
jes708: probably not, I think it's safe to nix the padding ability

src/Divider.js
```diff
@@ -0,0 +1,21 @@
+import styled from 'styled-components'
+import { space, width, borderColor, propTypes } from 'styled-system'
+
+const Divider = styled.hr`
+  ${space}
+  ${width}
+  border: 0;
+  border-bottom-style: solid;
+  border-bottom-width: 1px;
+  ${borderColor}
```
BeniCheni: Super trivial, but the template string(s) can be grouped together at top while the border styles with hard-coded styles at the bottom? (Just a tiny bit organization)

jxnblk: I think this is fine as-is – `border-color` needs to be declared at the end to override the `border` shorthand above. I think generally with styled components, functions at the end of the block makes more sense. This also has implications with media queries

.storybook/Divider.js
```diff
@@ -0,0 +1,43 @@
...
+import { Divider } from '../src'
+
+const description = '<hr /> with settings for padding, margin, width, and borderColor'
+
+storiesOf('Divider', module)
+  .add('Divider component', withInfo({
+    text: description,
+    inline: true
+  })(() => (
+    <Divider borderColor='lightGray' />
```
jxnblk: Just noticed you've got the `borderColor` prop on all these – let's set that as a defaultProp on Divider so there's a smart default. Also would be good to rename that value to `borderGray` since I merged #92 in

jes708: Updated @jxnblk

PR #96, opened 2017-10-10
add box-shadow styles

cenjennifer:
Add box-shadow to box component

### DIFF
--- a/src/Box.js
+++ b/src/Box.js
@@ -4,11 +4,38 @@ import PropTypes from 'prop-types'
 
 const align = responsiveStyle('text-align', 'align')
 
+
+function boxShadow(props) {
+  const shadowColor = 'rgba(0,0,0,0.08)'
+  const shadows = [
+    `0 0 4px 0 ${shadowColor}`, 
+    `0 8px 8px 0 ${shadowColor}`, 
+    `0 16px 16px 0 ${shadowColor}`, 
+    `0 32px 32px 0 ${shadowColor}`
+  ]
+  const boxShadows = {
+    'small': {
+      'box-shadow': shadows[0]
+    },
+    'medium': {
+      'box-shadow': shadows.slice(0,2).join(", ")
+    },
+    'large': {
+      'box-shadow': shadows.slice(0,3).join(", ")
+    },
+    'xlarge': {
+      'box-shadow': shadows.join(", ")
+    }
+  }
+  return boxShadows[props.boxShadowSize]
+}
+
 const Box = styled.div`
   ${space}
   ${width}
   ${color}
   ${align}
+  ${boxShadow}
 `
 
 Box.displayName = 'Box'
@@ -39,7 +66,14 @@ Box.propTypes = {
   pb: numberStringOrArray,
   pl: numberStringOrArray,
   px: numberStringOrArray,
-  py: numberStringOrArray
+  py: numberStringOrArray,
+  /** Box Shadow */
+  boxShadowSize: PropTypes.oneOf([
+    'small',
+    'medium',
+    'large',
+    'xlarge'
+  ])
 }
 
 export default Box
--- a/src/__tests__/Box.js
+++ b/src/__tests__/Box.js
@@ -38,4 +38,10 @@ describe('Box', () => {
     expect(json).toMatchSnapshot()
     expect(json).toHaveStyleRule('background-color', theme.colors.green)
   })
+
+  test('boxShadowSize prop sets box shadow', () => {
+    const json = renderer.create(<Box boxShadowSize='small' />).toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('box-shadow', '0 0 4px 0 rgba(0,0,0,0.08)')
+  })
 })
--- a/.storybook/Box.js
+++ b/.storybook/Box.js
@@ -81,3 +81,12 @@ storiesOf('Box', module)
       <Box my={3} color='white' bg='blue'>Margin Y-Axis</Box>
     </Box>
   ))
+  .add('Box Shadows', () => (
+    <Box>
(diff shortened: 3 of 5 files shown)


src/Box.js
```diff
@@ -4,11 +4,38 @@ import PropTypes from 'prop-types'
...
+    'medium': {
+      'box-shadow': shadows.slice(0,2).join(", ")
+    },
+    'large': {
+      'box-shadow': shadows.slice(0,3).join(", ")
+    },
+    'xlarge': {
+      'box-shadow': shadows.join(", ")
+    }
+  }
```
747823: Maybe I'm missing something here... why include multiple entries from the `shadows` array in the `box-shadow` value for larger sizes?

src/Box.js
```diff
@@ -4,11 +4,38 @@ import PropTypes from 'prop-types'
...
 const align = responsiveStyle('text-align', 'align')
 
+
+function boxShadow(props) {
+  const shadowColor = 'rgba(0,0,0,0.08)'
+  const shadows = [
+    `0 0 4px 0 ${shadowColor}`, 
+    `0 8px 8px 0 ${shadowColor}`, 
+    `0 16px 16px 0 ${shadowColor}`, 
+    `0 32px 32px 0 ${shadowColor}`
```
747823: I think the 0, 4, 8, 16, 32 px sizes should come from the theme probably

hakimelek: @odmln

jxnblk: I think it's okay to not pull from the `space` scale, but I *do* think these should be added to the `theme.js` file as a constant

src/Box.js
```diff
@@ -4,11 +4,38 @@ import PropTypes from 'prop-types'
...
+  }
+  return boxShadows[props.boxShadowSize]
+}
+
 const Box = styled.div`
   ${space}
   ${width}
   ${color}
   ${align}
+  ${boxShadow}
```
jxnblk: Thanks for putting this together! I think we might want to have the shadow as part of a separate `Card` component, which could extend the Box (e.g. `const Card = styled(Box)`. This would help keep the Box as mostly a layout component, then you could use the Card in places where you want the styles for shadows cc @odmln

src/Box.js
cenjennifer: @747823: I followed the box-shadow styles that @odmln provided https://github.com/pricelinelabs/design-system/issues/65

src/Box.js
cenjennifer: Sure! I'll create the Card component in a new branch with the changes mentioned in this PR

src/Box.js
```diff
@@ -4,11 +4,38 @@ import PropTypes from 'prop-types'
...
+      'box-shadow': shadows[0]
+    },
+    'medium': {
+      'box-shadow': shadows.slice(0,2).join(", ")
+    },
+    'large': {
+      'box-shadow': shadows.slice(0,3).join(", ")
+    },
+    'xlarge': {
+      'box-shadow': shadows.join(", ")
```
BeniCheni: Picked up this linting warning per quoting string with double quotes in JS. Plz consider updating?
screen shot 2017-10-10 at 6 55 14 pm

src/Box.js
```diff
@@ -4,11 +4,38 @@ import PropTypes from 'prop-types'
 
 const align = responsiveStyle('text-align', 'align')
 
+
+function boxShadow(props) {
+  const shadowColor = 'rgba(0,0,0,0.08)'
+  const shadows = [
+    `0 0 4px 0 ${shadowColor}`, 
+    `0 8px 8px 0 ${shadowColor}`, 
+    `0 16px 16px 0 ${shadowColor}`,
```
BeniCheni: Another linting warning per trailing spaces.
screen shot 2017-10-10 at 6 55 06 pm

src/Box.js
```diff
@@ -4,11 +4,38 @@ import PropTypes from 'prop-types'
 
 const align = responsiveStyle('text-align', 'align')
 
+
```
BeniCheni: While you are in this file, would you mind also update the code format to remove the extra new line per linting warning? 🙏 
screen shot 2017-10-10 at 7 01 22 pm

src/__tests__/Box.js
```diff
@@ -38,4 +38,10 @@ describe('Box', () => {
     expect(json).toMatchSnapshot()
     expect(json).toHaveStyleRule('background-color', theme.colors.green)
   })
+
+  test('boxShadowSize prop sets box shadow', () => {
+    const json = renderer.create(<Box boxShadowSize='small' />).toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('box-shadow', '0 0 4px 0 rgba(0,0,0,0.08)')
+  })
```
BeniCheni: Would you think the test cases can cover all the box shadow size alias from `small` to `xlarge`?

For box shadow aliases, would you think using short-hand terms such as `sm`, `md`,  `lg`, and `xl`? This will align to the naming convention in similar aliases in theme:
https://github.com/pricelinelabs/design-system/blob/master/src/theme.js#L26

src/Box.js
```diff
@@ -4,11 +4,38 @@ import PropTypes from 'prop-types'
...
+    'small': {
+      'box-shadow': shadows[0]
+    },
+    'medium': {
+      'box-shadow': shadows.slice(0,2).join(", ")
+    },
+    'large': {
+      'box-shadow': shadows.slice(0,3).join(", ")
+    },
+    'xlarge': {
```
BeniCheni: For box shadow aliases, would you think using short-hand terms such as `sm`, `md`,  `lg`, and `xl`? This will align to the naming convention in similar aliases in theme:
https://github.com/pricelinelabs/design-system/blob/master/src/theme.js#L26

src/Box.js
jxnblk: I'd say generally don't worry about linting since we'll likely be moving to using prettier instead – see #86

PR #97, opened 2017-10-11
feat(text-strikethrough): enable strikethrough

jes708:
Enable strikethough.  Useful for price striethroughs.

### DIFF
--- a/src/Text.js
+++ b/src/Text.js
@@ -65,5 +65,6 @@ Text.propTypes = {
 
 Text.span = Text.withComponent('span')
 Text.p = Text.withComponent('p')
+Text.s = Text.withComponent('s')
 
 export default Text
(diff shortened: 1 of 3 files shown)


src/Text.js
```diff
@@ -65,5 +65,6 @@ Text.propTypes = {
 
 Text.span = Text.withComponent('span')
 Text.p = Text.withComponent('p')
+Text.s = Text.withComponent('s')
```
BeniCheni: Can the term `s` be a little bit verbose to be, perhaps, `strike`?

jxnblk:
We do use `<s>` quite a bit here :) This LGTM

src/Text.js
jes708: I like the convention of the name matching the tag.  Takes the guesswork out of the naming.  I wouldn't want it to be confused with the obsolete `<strike>` element either

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike

PR #98, opened 2017-10-11
Feature/container

jinchi2013:


.storybook/Container.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Container, Box } from '../src'
+
+storiesOf('Container', module)
+  .add('Default align center', () => (
+    <Container maxWidth={500}>
+      <Box p={4} bg='gray0' style={{height: `100vh`}}>
+        Container Compnent
```
jes708: Typo: should say Component

src/Container.js
```diff
@@ -0,0 +1,32 @@
...
+      }
+    default :
+      return {
+        marginLeft: 'auto',
+        marginRight: 'auto'
+      }
+  }
+}
+
+const maxWidth = props => props.maxWidth ? { maxWidth: `${props.maxWidth}px` } : null
```
jes708: @jxnblk Should something like this live in `styled-system` instead?

PR #99, opened 2017-10-11
Feature/container

jinchi2013:


src/Container.js
```diff
@@ -0,0 +1,17 @@
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+
+const maxWidth = props => props.maxWidth ? { maxWidth: `${props.maxWidth}px` } : { maxWidth: props.theme.maxContainerWidth }
```
jes708: Would something like this belong in `styled-system` @jxnblk ?

jxnblk: This could be handled with the generic [`style`](https://github.com/jxnblk/styled-system#style) utility, but I don't mind this here as it's fairly readable as-is

.storybook/Container.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Container, Box, theme } from '../src'
+
+storiesOf('Container', module)
+  .add('Default align with theme max width', () => (
+    <Container theme={theme}>
+      <Box p={4} bg='gray0' style={{height: `100vh`}}>
```
jxnblk: Just a side note: we've discussed potentially deprecating the numbered color names in favor of `lightGray`, `gray`, and `darkGray` so probably best to use those instead

docs/Container.md
```diff
@@ -0,0 +1,18 @@
+
+# `<Container />`
+
+The `<Container />` component is used to align the app body at the center of screen.
+
+If you'd like to add padding or margin, please use `<Box />` inside it.
```
jxnblk: 😎

jxnblk:
I think this looks great! We probably wanna make sure to double check with @odmln and @chrisdroukas to make sure this the value they want to use for the max-width

chrisdroukas:
`1280px` is perfect!

src/__tests__/Container.js
```diff
@@ -0,0 +1,16 @@
...
+import 'jest-styled-components'
+import { Container, theme } from '..'
+
+describe('Container', () => {
+  test('renders with default theme max-width', () => {
+    const json = renderer.create(<Container theme={theme} />).toJSON()
+    expect(json).toMatchSnapshot()
+  })
+  
+  test('renders with left align', () => {
```
jxnblk: I *think* this description needs updated??

src/__tests__/Container.js
```diff
@@ -0,0 +1,16 @@
...
+    const json = renderer.create(<Container theme={theme} maxWidth={1024} />).toJSON()
```
jxnblk: The storybook should have the ThemeProvider available and you shouldn't need to pass `theme` as a prop – although it is buggy in storybook at the moment

PR #100, opened 2017-10-11
Feature/add card component

cenjennifer:
Add Card Component with extension to Box component

Changes to comments in the original PR (https://github.com/pricelinelabs/design-system/pull/96) are addressed in this PR

### DIFF
--- a/src/Card.js
+++ b/src/Card.js
@@ -29,8 +29,8 @@ function boxBorder(props) {
 
 const Card = styled(Box)`
   ${boxShadow}
+  ${boxBorder}
 `
-//${boxBorder}
 
 Card.propTypes = {
   boxShadowSize: PropTypes.oneOf([
--- a/src/Card.js
+++ b/src/Card.js
@@ -22,16 +22,16 @@ function boxShadow(props) {
 }
 
 const Card = styled(Box)`
-	${boxShadow}
+  ${boxShadow}
 `
 
 Card.propTypes = {
-	boxShadowSize: PropTypes.oneOf([
-		'sm',
-		'md',
-		'lg',
-		'xl'
-    ])
+  boxShadowSize: PropTypes.oneOf([
+    'sm',
+    'md',
+    'lg',
+    'xl'
+  ])
 }
 
 export default Card
--- a/src/Card.js
+++ b/src/Card.js
@@ -21,9 +21,16 @@ function boxShadow(props) {
   return boxShadows[props.boxShadowSize]
 }
 
+function boxBorder(props) {
+    return {
+      border: `${props.borderWidth}px solid ${props.theme.colors.borderGray}`
+    }
+}
+
 const Card = styled(Box)`
   ${boxShadow}
 `
+//${boxBorder}
 
 Card.propTypes = {
   boxShadowSize: PropTypes.oneOf([
@@ -31,7 +38,13 @@ Card.propTypes = {
     'md',
     'lg',
     'xl'
-  ])
+  ]),
+  borderColor: PropTypes.string,
+  borderWidth: PropTypes.oneOf([1, 2])
+}
+
+Card.defaultProps = {
+  borderWidth: 1
 }
 
 export default Card
--- a/src/Card.js
+++ b/src/Card.js
@@ -1,0 +1,37 @@
+import styled from 'styled-components'
+import Box from './Box'
+import PropTypes from 'prop-types'
+
+function boxShadow(props) {
+
+  const boxShadows = {
+    'sm': {
+      'box-shadow': props.theme.boxShadows[0]
+    },
+    'md': {
+      'box-shadow': props.theme.boxShadows.slice(0,2).join(', ')
+    },
+    'lg': {
+      'box-shadow': props.theme.boxShadows.slice(0,3).join(', ')
+    },
+    'xl': {
+      'box-shadow': props.theme.boxShadows.join(', ')
+    }
+  }
+  return boxShadows[props.boxShadowSize]
+}
+
+const Card = styled(Box)`
+	${boxShadow}
+`
+
+Card.propTypes = {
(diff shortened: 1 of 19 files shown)


src/__tests__/Card.js
```diff
@@ -0,0 +1,30 @@
...
+  test('renders small box shadow', () => {
+    const json = renderer.create(<Card boxShadowSize='sm' theme={theme} />).toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('box-shadow', '0 0 4px 0 rgba(0,0,0,0.08)')
+  })
+
+  test('renders medium box shadow', () => {
+    const json = renderer.create(<Card boxShadowSize='md' theme={theme} />).toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('box-shadow', '0 0 4px 0 rgba(0,0,0,0.08),0 8px 8px 0 rgba(0,0,0,0.08)')
```
cenjennifer: Jest doesn't seem to add the space from the separator in the `.join(', ')` when multiple shadows are joined together. Anyone know what might be the cause?
screen shot 2017-10-11 at 12 43 30 pm

I modified the expected styles to not include the space to satisfy the tests, though would like to fix it if anyone have any insights. Thanks!

jinchi2013: Not for sure, But I guess the `toJSON()` method remove that space.

jxnblk: My guess is this might be related to how [stylis](https://github.com/thysultan/stylis.js) parses the CSS, but not sure

src/Card.js
```diff
@@ -0,0 +1,37 @@
+import styled from 'styled-components'
+import Box from './Box'
+import PropTypes from 'prop-types'
+
+function boxShadow(props) {
+
+  const boxShadows = {
+    'sm': {
+      'box-shadow': props.theme.boxShadows[0]
```
jxnblk: 👍

src/Card.js
```diff
@@ -0,0 +1,37 @@
...
+  }
+  return boxShadows[props.boxShadowSize]
+}
+
+const Card = styled(Box)`
+  ${boxShadow}
+`
+
+Card.propTypes = {
+  boxShadowSize: PropTypes.oneOf([
```
jxnblk: Might be worth setting a defaultProp for whichever size we think is the most commonly used - cc @odmln

jxnblk:
This looks great! Thanks for making these changes

src/Card.js
```diff
@@ -0,0 +1,37 @@
...
+    },
+    'xl': {
+      'box-shadow': props.theme.boxShadows.join(', ')
+    }
+  }
+  return boxShadows[props.boxShadowSize]
+}
+
+const Card = styled(Box)`
+  ${boxShadow}
```
jes708: It might make sense to add `${borderColor}` possibly with a default color of `borderGray`.  I know it makes sense for the upcoming summary of charges from the packages team.

jes708: Or maybe `borderColor` belongs on `Box`?

cenjennifer: @jes708: Probably make sense to add it to `Box` esp. if we have a default shadow for `Card`. I can add it to `Box` if everyone agrees.

jes708: That sounds right to me.

odmln: My thought is to let `card` give the border or shadow. Let `box` handle the layout and `card` handle styling.

src/Card.js
```diff
@@ -0,0 +1,37 @@
...
+    'md': {
+      'box-shadow': props.theme.boxShadows.slice(0,2).join(', ')
+    },
+    'lg': {
+      'box-shadow': props.theme.boxShadows.slice(0,3).join(', ')
+    },
+    'xl': {
+      'box-shadow': props.theme.boxShadows.join(', ')
+    }
+  }
```
747823: RE: the box-shadow sizes being concatenated - Oh I see, so based on your  I guess it's for design, to make the larger shadows softer. That makes sense.

Another thought: I'd want to see if there's any performance impact on from this when rendering a lot of cards at once compared to not layering multiple shadows per box, especially on mobile. Might not be a concern, but it could have an impact...

src/Card.js
747823: @odmln I agree, I would say borderColor belongs on Card not on Box. But a default gray borderColor might be nice.

jes708: I like that reasoning

cenjennifer: Cool, I'll add `border` to Card. Should there be a default border size?  @odmln: What should the default color for border be?

odmln: @cenjennifer I've been using `box-shadow: 0 0 0 1px rgba(0,0,0,0.08)` instead of a pure border for things like cards, popovers, and tooltips. But if thats too light we do now have `borderGray #abb7c2` in the design system that you could use. I don't think cards will really be using different color borders, it's only inputs that will change.

src/Card.js
jxnblk: There would definitely be some performance issues *if* we transitioned/animated a Card or changed box-shadow, but I suspect static box-shadows should be fine

src/Card.js
cenjennifer: I made border a defaultProp, but not 100% sure if border and box-shadow are separate styles that can be displayed individually in a card component @odmln

src/Card.js
```diff
@@ -0,0 +1,50 @@
...
+  return boxShadows[props.boxShadowSize]
+}
+
+function boxBorder(props) {
+    return {
+      border: `${props.borderWidth}px solid ${props.theme.colors.borderGray}`
+    }
+}
```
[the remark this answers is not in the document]
cenjennifer: @hakimelek: Ok, will update in another PR

PR #103, opened 2017-10-13
feat(typography): update font-weights/ensure header weights/edit capi…

jes708:
update font-weights
ensure header weights match Zeplin typography
remove capitalization spacing
import Montserrat to storybook


docs/Heading.md
```diff
@@ -11,7 +11,7 @@ Use the `<Heading />` component to use HTML heading h1-h6 element, with all the
 <Heading.h1>Hello</Heading.h1>
 
 // Heading with <Text> properties
-<Heading.h5 m={3} bold>Hello</Heading.h5>
+<Heading.h4 m={3} bold>Hello</Heading.h4>
```
hakimelek: We should probably mention in the docs to not forget adding `<link href="https://fonts.googleapis.com/css?family=Montserrat:400,600" rel="stylesheet">` on the project where the design-system is being consumed.

jes708: updated

docs/Text.md
```diff
@@ -2,6 +2,12 @@
 
 Use the `<Text />` component to control font size, weight, alignment, and color.
 
+###### Note: To use the `Montserrat` font-family in you project, don't forget to include the following code inside your html `<head>`
```
hakimelek: 👍

jxnblk: For performance reasons, we'll probably want to investigate different font loading strategies and have that built into the boilerplate at some point

chrisdroukas: cc @BeniCheni

src/Text.js
```diff
@@ -43,6 +54,8 @@ Text.propTypes = {
     'justify'
   ]),
   caps: PropTypes.bool,
+  regular: PropTypes.bool,
+  medium: PropTypes.bool,
```
jxnblk: I like having the alias in to point at the same value 👍

src/Text.js
```diff
@@ -4,8 +4,17 @@ import { fontSize, space, color, responsiveStyle } from 'styled-system'
...
+  }
+  : null
+
+export const regular = props => props.regular ? { fontWeight: props.theme.regular} : null
+
+export const medium = props => props.medium
+  ? {
+    fontWeight: props.theme.bold,
+    'text-rendering': 'optimizeLegibility',
+    '-webkit-font-smoothing': 'antialiased'
```
jxnblk: I think the font-smoothing should only be applied to reversed text (white on dark background) and I think this would be fine to hard-code in the Button component

docs/Text.md
jxnblk: webfontloader is good. there's probably some opportunities for use to do things like cache the font in localStorage/etc – not sure what current best practices are around this sort of thing

src/Text.js
jes708: Do you think that needs to be handled differently?  i.e. should `medium` be renamed to `smooth`.

docs/Text.md
hakimelek: I have never tried localStorage caching but I have heard many times that it's a great hack for perf optimization. Also, a great article about loading and rendering fonts here: https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/webfont-optimization#optimizing_loading_and_rendering

.storybook/Text.js
```diff
@@ -30,6 +30,12 @@ storiesOf('Text', module)
       <Text align='right'>Hello Right</Text>
     </div>
   ))
+  .add('regular', () => (
+    <Text regular>Hello Medium</Text>
```
krevat: should be Hello Regular?

src/Text.js
jxnblk: I mean, we shouldn't use the font-smoothing prop on normal, dark-on-light applications. The OSes' font rendering engines work fine there. Where people tend to use this property is for white-on-dark applications. When speaking to @odmln we talked about applying this to the Button component specifically, not as a replacement for font weights generally

jes708: Alright - `medium` is removed, and hardcoded on button instead

.storybook/Text.js
jes708: Correct!

src/__tests__/__snapshots__/GreenButton.js.snap
```diff
@@ -41,7 +41,7 @@ exports[`GreenButton renders 1`] = `
   text-align: center;
   text-decoration: none;
   font-weight: 600;
-  cursor: pointer;
+  text-rendering: optimizeLegibility, -webkit-font-smoothing:antialiased cursor:pointer;
```
hakimelek: Do you have an idea what is causing this?

jes708: I do!  Fixed

src/Button.js
```diff
@@ -40,6 +40,8 @@ const Button = styled.button`
   text-align: center;
   text-decoration: none;
   font-weight: 600;
+  text-rendering: optimizeLegibility;
+  -webkit-font-smoothing: antialiased;
```
jxnblk: This was covered in #112 – can you remove these lines and merge the latest master into this branch?

src/theme.js
```diff
@@ -39,7 +39,7 @@ export const space = [
   128
 ]
 
-export const font = `'Helvetica Neue',Helvetica,Arial,sans-serif`
+export const font = `'Montserrat','Helvetica Neue',Helvetica,Arial,sans-serif`
```
jxnblk: 😎

src/Button.js
jes708: `text-rendering: optimizeLegibility;` isn't necessary?

PR #104, opened 2017-10-17
refactor card component with arrow fn

cenjennifer:
Small change refactoring `function` to `const/arrow function` commented by @hakimelek after this PR merge: https://github.com/pricelinelabs/design-system/pull/100

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ For all contributors, please be sure to read the [Contributing](docs/Contributin
 - [`<BackgroundImage />`](docs/BackgroundImage.md)
 - [`<Box />`](docs/Box.md)
 - [`<Button />`](docs/Button.md)
+- [`<Card />`](docs/Card.md)
 - [`<Heading />`](docs/Heading.md)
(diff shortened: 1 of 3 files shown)


jes708:
I'm having some trouble with the `Card` component.  I tried including it in a component, but then all of the component's tests fail as follows:

```TypeError: Cannot read property '0' of undefined

      at boxShadow (node_modules/pcln-design-system/dist/Card.js:29:43)
```


implemented as follows

```<Card mb={2} bg='white' boxShadowSize='md'>```

May be worth fixing in this PR if possible

cenjennifer:
@jes708: Try passing in `theme={theme}` in your card component test. I think it is trying to reference the `boxShadows` array in theme

jes708:
Thanks @cenjennifer, The ThemeProvider needed to be added as a wrapper in the tests.

jes708:
Not too many people have been doing this, but it might be worth it to add a link to the `<Card>` documentation in `README.md`

PR #105, opened 2017-10-17
Feat/icon chevron right

msafari:
- renamed `srollLeft` to `chevronLeft`
 - renamed `chevronLight` to `chevronDown`
 - renamed `chevronThick` to `chevronDownThick`
- added `chevronRight`

screen shot 2017-10-18 at 2 12 00 pm

old ones that got renamed still work:
screen shot 2017-10-18 at 2 11 55 pm


747823:
I think the original one is called chevronLight because it's a light stroke weight. There's also chevronThick. So if we're calling the left and right versions chevron too, it seems like we should include Light or Thick in the name to be consistent.

msafari:
@747823 I see. I think it's a bit more verbose to say `chevronRightLight` and `chevronLeftLight`
I actually think if we change them like this it would a little better:
```
{
 `chevronLight`: `chevronDown` 
 `chevronThick`: `chevronDownThick`,
 `chevronRight`: `chevronRight`,  // keep it the same
 `chevronLeft`: `chevronLeft`, // keep it the same
 // and so on if any icon is thick add the thick keyword. Otherwise keep it simple?
}
```

That's just a thought. Because I actually think that `chevronLight` and `chevronThick` has a confusing name (it doesn't imply anything about the direction)
Then again if we go with this change, anyone that was using those renamed icons have to change the name.
Thoughts @jxnblk, @byrekt?

747823:
That sounds good to me @msafari

jxnblk:
I think it'd be *great* if we didn't have two thicknesses for icons that mean the same thing :) I'd say keep the name simple for the default, which I'm guessing is the non-thick version

msafari:
@747823 @jxnblk, I renamed the chevrons like I described. I also added temporary aliases for the old names so it won't break anyone's functionality.

PR #106, opened 2017-10-17
Feature/59 button hover and focus

Degron:
Issue 98. Added hover and focus for button, put all buttons in one element in storybook


src/__tests__/Button.js
```diff
@@ -37,4 +38,10 @@ describe('Button', () => {
     expect(json).toMatchSnapshot()
     expect(json).toHaveStyleRule('width', '100%')
   })
+
+  test('disabled prop sets', () => {
```
msafari: Did `simulate('mouseEnter)` not work? I see you imported `enzyme` here but not using it. I'm assuming you wanted to mount the component and then test the styled on mouse enter?

Degron: It not working.

msafari: ohhh right! It wouldn't work because once you mount the component you can't use `toHaveStyleRule` anymore.

.storybook/Button.js
```diff
@@ -1,27 +1,38 @@
 import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { withInfo } from '@storybook/addon-info'
-import { Button } from '../src'
+import { Button, GreenButton, RedButton, OutlineButton } from '../src'
```
msafari: Nice! I like having all the button stories in one place

src/__tests__/Button.js
msafari: Ah I think they actually added support for testing pseudo element's styles https://github.com/styled-components/jest-styled-components/pull/60/files#diff-99a96acd66a8852eaa09e5a093229cc9. Let's use that to test your hover styles

msafari: Yep I tried it on your branch. Update to ` "jest-styled-components": "^4.7.1"` and then add tests for hover styles:
```
  test('disabled prop sets', () => {
    const json = renderer.create(<Button disabled theme={theme} />).toJSON()
    expect(json).toHaveStyleRule('background-color', theme.colors.blue)
    expect(json).toHaveStyleRule('background-color', theme.colors.blue, {
      modifier: ':hover',
    })
  })

  test('without disabled prop sets', () => {
    const json = renderer.create(<Button theme={theme} />).toJSON()
    expect(json).toHaveStyleRule('background-color', theme.colors.darkBlue, {
      modifier: ':hover',
    })
  })
```

Degron: I updated my test. Thank you @msafari!!!!!!

src/OutlineButton.js
```diff
@@ -3,22 +3,19 @@ import PropTypes from 'prop-types'
...
   border-style: solid;
-  border-color: currentcolor;
-  ${color}
+  color: ${props => props.theme.colors.blue};
+  border-color: ${props => props.theme.colors.blue};
+  
+  &:hover {
+    color: ${props => props.disabled ? props.theme.colors.blue : props.theme.colors.darkBlue};
+    border-color: ${props => props.disabled ? props.theme.colors.blue : props.theme.colors.darkBlue};
+    background-color: transparent;
```
msafari: Does this need to also check for `disabled`?

src/__tests__/OutlineButton.js
```diff
@@ -8,4 +8,20 @@ describe('OutlineButton', () => {
...
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('background-color', 'transparent')
+    expect(json).toHaveStyleRule('border-color', theme.colors.blue)
+    expect(json).toHaveStyleRule('color', theme.colors.blue)
+  })
+
+  test('without disabled prop sets', () => {
+    const json = renderer.create(<OutlineButton theme={theme} />).toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('border-color', theme.colors.darkBlue, {
```
msafari: @Degron I see `OutlineButton` has more styles on hover like `color`. Let's check for that here too

src/OutlineButton.js
msafari: Oh I see it's transparent even not on hover. So I would remove this here. It doesn't do anything

msafari: Talked to @Degron. This background-color needs to be there to override the hover background-color of the parent component (Button).

src/__tests__/OutlineButton.js
Degron: Fixed it.

src/__tests__/Button.js
jxnblk: Yeah, generally we shouldn't ever need to use enzyme in this repo, since it's primarily CSS and styled-components - enzyme's really mostly useful for testing interactions with state

package.json
```diff
@@ -67,6 +67,7 @@
     ]
   },
   "dependencies": {
+    "enzyme": "^3.1.0",
```
jxnblk: ✂️ - also test libs should generally be in `devDependencies` instead of `dependencies`

src/Button.js
```diff
@@ -51,6 +50,10 @@ const Button = styled.button`
     opacity: 0.25;
   }
 
+  &:hover {
+    background-color: ${props => props.disabled ? props.theme.colors.blue : props.theme.colors.darkBlue};
```
jxnblk: I'd pass `null` for the `disabled` state since it shouldn't look clickable

package.json
msafari: @jxnblk this is never used. @Degron did you remove it?

hakimelek:
Looks good to me!

PR #109, opened 2017-10-18
feat(banner): create banner and palettes

jes708:
create banner and palettes

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -141,6 +141,35 @@ export const boxShadows = [
   `0 32px 32px 0 ${shadowColor}`
 ]
 
+const palette = {
+  default: {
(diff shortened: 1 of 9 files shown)


src/theme.js
```diff
@@ -141,6 +141,35 @@ export const boxShadows = [
   `0 32px 32px 0 ${shadowColor}`
 ]
 
+const palette = {
+  default: {
```
jxnblk: We've chatted about this sort of thing with the Button components and the Badge as well – I think it's generally a lot easier to refer to the color of a component by it's actual color name – e.g. "the blue Banner" – rather than make some abstraction on top of the color names. To help understand the color meaning, I think we should handle that with better documentation

PR #111, opened 2017-10-18
Feature/banner

jes708:
screen shot 2017-10-18 at 11 52 58 am


.storybook/Banner.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Box, Banner } from '../src'
+
+storiesOf('Banner', module)
+  .add('All palettes', () => (
+    <Box>
+      <Banner align='right' mb={2} p={3} title='DEFAULT' subtitle='Secondary Text' />
```
hakimelek: I am pretty sure that we will have to change this in the future and have some design additions (adding a closing icon for example..). I would vote to use a [Compound Component](https://github.com/ReactTraining/react-workshop/blob/master/subjects/CompoundComponents/lecture.js) (now I know what does that mean :)). Maybe the Banner should extend a Card and create something like

```
<Banner palette='warning'>
  <BannerTitle></BannerTitle>
  <BannerSubtitle></BannerSubtitle>
</Banner>
```

odmln:
@jes708 @hakimelek I mocked this up and put it in [Zeplin](https://zpl.io/br1pZL1) for reference. Wanted to make it easier for you guys to see the different banner options. Basically you'd select a bg color, light or dark text, if you want an icon or not, and if you want a header or not. 

screen shot 2017-10-20 at 11 54 31 am

jes708:
@odmln I centered the icons rather than aligning them to the top.  I found this made for cleaner code and better alignment.  Let me know if this is acceptable.

screen shot 2017-10-24 at 12 41 17 pm

odmln:
@jes708 The reason I placed it top left was incase the text wraps to multiple lines. Don't know how often this will be the case but if it does wrap to 2 or 3 lines how will a centered icon look? Im open to exploration on this 👍

icons/attention.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 46.2 (44496) - http://www.bohemiancoding.com/sketch -->
+    <title>alert / attention / dark</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Icons" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="alert-/-attention-/-dark">
+            <g id="attention">
+                <polygon id="Shape" points="0 0 24 0 24 24 0 24"></polygon>
```
jxnblk: The current icon parsing script doesn't pick up the `<polygon>` element, this will need to be flattened in a design tool

package.json
```diff
@@ -47,6 +51,8 @@
     "babel-preset-stage-0": "^6.24.1",
     "camelcase": "^4.1.0",
     "codecov": "^2.3.0",
+    "enzyme": "^3.1.0",
+    "enzyme-adapter-react-15": "^1.0.2",
```
jxnblk: Is this required for enzyme now? Also, we should probably go ahead and move this repo to React 16 in another PR

src/Banner.js
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <Flex
+      {...props}
+      bg={palette[props.palette].backgroundColor}
+      color={palette[props.palette].color}
+      justify='space-between'
+      align='center'
+    >
+      {!!icon && !!props.showIcon && <Icon name={icon} mr={3} size={28} />}
```
jxnblk: Why both `icon` and `showIcon`? Seems like checking for `icon` should be okay

jxnblk: Oh, nm. I see that `icon` isn't coming from props... Mentioned this earlier, but think it would be good to keep the icon configurable via props, rather than tied to some semantic meaning. It would make it less automatic, but more flexible down the road

src/Box.js
```diff
@@ -8,6 +8,7 @@ const Box = styled.div`
   ${width}
   ${color}
   ${align}
+  ${borderRadius}
```
jxnblk: Don't think you're actually using this are you?

package.json
jes708: Yeah, this is an enzyme 3 thing.  Requires setup with an adapter now.

src/Banner.js
jes708: What would you think about `const icon = props.icon || palette[props.palette].icon`?  Would that be the best of both worlds in terms of having presets and configurations?

jes708:
@odmln Here are some more wordy banners, to visualize your concern
screen shot 2017-10-24 at 6 10 07 pm

src/Box.js
jes708: Nope

odmln:
@jes708 thanks for visualizing that for me, personally I still feel as if the icon should sit at the top left next to the bold text – @chrisdroukas what do you think?

jes708:
@odmin That should be possible.  Let me rework this a little bit

package.json
codeflyer: What happen if we move to React16 to the other app?

src/Banner.js
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <Flex
+      {...props}
+      bg={palette[props.palette].backgroundColor}
+      color={palette[props.palette].color}
+      justify="space-between"
+      align="center"
+    >
+      {!!icon &&
```
codeflyer: Just for readability, is better to add the condition in parentesis.

```(!!icon && !!props.showIcon) &&```

File: package.json (hunk 3 above)
Comment (jxnblk) [reply package.json]: I don't suspect using different versions of react here vs in our apps will really make a difference, unless we add react 16 specific things to the design-system, like returning arrays of elements, which we probably shouldn't do in the near term

File: docs/Banner.md [hunk 7]
```diff
@@ -0,0 +1,27 @@
...
+  text='This is a banner'
+  onClose={() => {}}
+  showIcon='false'
+  textAlign='right'
+/>
+```
+
+Prop | Type | Description
+---|---|---
+palette | string | Sets background-color and color. Accepts `default`, `information`, `success`, `attention`, `warning`, `gray`, `blue`, `green`, `orange`, `red`
```
jxnblk: I think for the time being it would be better to keep the color props api similar across components and use either `color` or `bg` instead of `palette` – see the implementation of [Badge](https://github.com/pricelinelabs/design-system/blob/master/src/Badge.js) for a similar situation. ("Palette" also implies what `theme.colors` is in my mind)

Once we have a set of these more complex color combinations in a handful of components, I suspect we'll find a pattern that can be abstracted up to the theme level, if that makes sense

docs/Banner.md
```diff
@@ -0,0 +1,27 @@
...
+header | string | Sets header content
+iconName | string | Sets left-hand svg icon name
+iconSize | number | Sets left-hand svg icon size
```
jxnblk: Would be great to *not* change the icon size in different situations IMO cc @odmln

src/IconButton.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import styled from 'styled-components'
+import Icon from './Icon'
+import Button from './Button'
+import PropTypes from 'prop-types'
+
+const StyledButton = styled(Button)`
```
jxnblk: I think it's generally better to give components like this more specific names, like *UnstyledButton* or *TransparentButton* – all of our buttons are "styled"

src/__tests__/Banner.js
```diff
@@ -0,0 +1,191 @@
...
+      'background-color',
+      theme.palette.lightRed.backgroundColor
+    )
+    expect(json).toHaveStyleRule('color', theme.palette.lightRed.color)
+  })
+
+  test('renders close button if onClose func is provided', () => {
+    const wrapper = shallow(<Banner onClose={() => {}} theme={theme} />)
+    const close = wrapper.find('CloseButton')
+    expect(close).toHaveLength(1)
```
jxnblk: I think this sort of thing is okay here, but when I see a lot of conditional logic checking for children, I think it's good to consider, *should this be another component* or *can this simply be handled with composition*

An alternative API for a component like this could look like:

```jsx
<Banner color='blue'>
  <Icon name='check' />
  <Text>Hello</Text>
  <CloseButton onClick={this.handleClose} />
</Banner>
```

Using `React.Children` you can generally handle any internal styling that you might need for something like this

docs/IconButton.md
```diff
@@ -0,0 +1,20 @@
+# `<IconButton />`
```
jxnblk: Should this move to #121?

icons/attention.svg
jxnblk: Chatted with Oliver about this stuff. I think this is an "artboard" left over from the design tool. The parse-icons script will just ignore this so I think it's fine if the icon renders correctly

src/Banner.js
```diff
@@ -0,0 +1,91 @@
...
+
+const Banner = props => {
+  const bannerColor = bannerColors[props.bg] || {}
+  const icon = props.iconName || bannerColor.icon
+
+  return (
+    <Box {...props} bg={bannerColor.backgroundColor} color={bannerColor.color}>
+      <Flex justify="space-between" align="flex-start">
+        {!!icon &&
+          !!props.showIcon && <Icon name={icon} mr={3} size={28} mt="-2px" />}
```
jxnblk: Is this intentionally size `28`? Icons tend to be 16, 24, or 32px

src/CloseButton.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import IconButton from './IconButton'
+import PropTypes from 'prop-types'
+
+const CloseButton = (props) => (
+  <IconButton {...props} name='closeLight' title={props.title} onClick={props.onClick} size={props.size} />
+)
```
jxnblk: Nice! Again here, setting `title`, `onClick`, and `size` are redundant since you set `{...props}`

src/Banner.js
jes708: Updating

docs/IconButton.md
jes708: #121 includes anything related to IconButton in this PR.  If any of that is removed from this PR, aside from documentation, the banner would not function properly.  Do you recommend that any files be removed from this PR?  This documentation is included in #121 already.

jes708:
@odmin forgot to update you.  The banner now has icons and close buttons stuck to the top.

screen shot 2017-10-30 at 11 19 08 am

Issue #72, opened 2017-09-25
FormField components


jxnblk:
Composite components that include labels, helper message text, and validation messaging – with props for different states (invalid, disabled, success, warning, etc)

PR #115, opened 2017-10-23
Form field

byrekt:
This PR adds 3 new components, a Label component, an Input component, and an InputField component.

Both Input and Label are simple styled components.

InputField uses both Input and Label to render something similar to that found in https://app.zeplin.io/project/596d11bc29e9db6303116850/screen/59ce7dfd4fb3d1a16d85589d

The error state however was updated last week and hasn't been implemented yet. That will come as a separate PR.

image

The Input elements appear to have a thinner border, but after talking with Oliver we decided that the border should only be bolded on focus. Otherwise it might not be apparent to a user that the element is focused.


.storybook/Input.js
```diff
@@ -0,0 +1,29 @@
...
+  })(() => (
+    <Label m={3}>
+      Label Component
+    </Label>
+  )))
+  .add('Colors', () => (
+    <Box width={400}>
+      <Input mb={3} id="input-colors-1" placeholder="No color"/>
+      <Input mb={3} id="input-colors-2" color={theme.colors.red} placeholder="Red"/>
+      <Input mb={3} id="input-colors-3" color={theme.colors.green} placeholder="Green"/>
```
jxnblk: Haven't looked too closely at the implementation yet, but the styled-system `color` function picks up colors from `props.theme` so that you can set the key as a prop value, e.g. `color='green'` would give you `theme.colors.green`. I think we'd want this `color` prop to work the same way

src/Input.js
```diff
@@ -0,0 +1,44 @@
+import styled from 'styled-components'
+import { space } from 'styled-system'
+
+const Input = styled.input`
+  appearance: none;
+  display: block;
+  width: 100%;
+  font-family: inherit;
+  font-size: ${({ theme }) => theme.fontSizes[2]}px;
```
jxnblk: Not necessary to use here, but styled-system has a [`theme`](https://github.com/jxnblk/styled-system#theme) getter utility to pick up values like this with `${styledSystem.theme('fontSizes.2')}px`

src/Input.js
```diff
@@ -0,0 +1,44 @@
...
+  border-radius: ${({ theme }) => theme.radius};
+  border-width: 1px;
+  border-style: solid;
+  border-color: ${({theme}) => theme.colors.borderGray};
+
+  ${({color}) => `
+    outline: 0;
+    border-color: ${color};
+    box-shadow: 0 0 0 0px ${color};
+  `}
```
jxnblk: Might be cleaner to pull this out into its own function?

src/Input.js
```diff
@@ -0,0 +1,44 @@
...
+  :focus {
+    outline: 0;
+    border-color: ${({color, theme}) => color || theme.colors.blue};
+    box-shadow: 0px 0px 0px 1px ${({color, theme}) => color || theme.colors.blue};
+  }
+
+  padding-top: 14px;
+  padding-bottom: 14px;
+  padding-left: 12px;
+  padding-right: 12px;
```
jxnblk: This deviates from the spacing scale, which means that the labels won't always line up with things outside of the InputField, but @byrekt and I spoke about it, and I think this is fine for component "internals" like this

src/InputField.js
```diff
@@ -0,0 +1,59 @@
...
+  /** Additional content to go in the info drawer below the element. This will be replaced by a tooltip later probably.  */
+  info: PropTypes.string
+}
+const InputField = ({
+  id,
+  label,
+  color,
+  icon,
+  info,
+  ...otherProps
```
jxnblk: I feel like the typical convention for naming this is either `...props` or `...rest` - maybe consider sticking to a common convention?

src/InputField.js
```diff
@@ -0,0 +1,59 @@
...
+}) => {
+
+  return (
+    <div>
+      <Label htmlFor={id} pl={13}>{label}</Label>
+      <Flex justify='space-between' align='center' width={1}>
```
jxnblk: Probably don't need the `justify='space-between'` here, right?

src/InputField.js
```diff
@@ -0,0 +1,59 @@
...
+        <Input
+          id={id}
+          pt={20}
+          pb={8}
+          mt={-20}
```
jxnblk: 😎

src/Label.js
```diff
@@ -0,0 +1,14 @@
+import styled from 'styled-components'
+import { space, fontSize, color } from 'styled-system'
+
+const Label = styled.label`
+  font-size: 10px;
```
jxnblk: If you're using the `fontSize` function, you should probably set this as a `defaultProp`

src/__tests__/Icon.js
```diff
@@ -24,7 +24,13 @@ describe('Icon', () => {
 
 
   test('returns false for non-existing icons', () => {
+    // Mock out console.error since we're expecting a propType warning
+    console.error = jest.genMockFunction();
```
jxnblk: 👍

src/InputField.js
byrekt: Maybe not, I'll verify

src/index.js
```diff
@@ -7,6 +7,9 @@ export { default as Divider } from './Divider'
 export { default as GreenButton } from './GreenButton'
 export { default as RedButton } from './RedButton'
 export { default as OutlineButton } from './OutlineButton'
+export { default as Label } from './Label'
```
jxnblk: We were starting to keep these in alphabetical order to help with merge conflicts, but think we fell off track a little

src/Input.js
```diff
@@ -0,0 +1,55 @@
+import styled from 'styled-components'
+import { space, theme, propTypes } from 'styled-system'
+import PropTypes from 'prop-types'
+
+const borders = ({ color, theme }) => {
+  const borderColor = (color) ? theme.colors[color] : theme.colors.borderGray
```
hakimelek: I don't think we should import theme here but rather use the theme as prop coming from the `ThemeProvider`

src/__tests__/InputField.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import 'jest-styled-components'
+import { ThemeProvider, InputField } from '..'
+
+describe('InputField', () => {
+  test('it renders', () => {
+    const json = renderer.create(
+      <ThemeProvider>
```
hakimelek: We can actually use `import { theme } from '..'` here and pass `theme` as a prop instead of using `ThemeProvider`. I start seeing people using different patterns for how to test their components. I used to it the way you did it and then I have changed my mind after I had a discussion with @jxnblk. `ThemeProvider` adds an extra dependency for testing the components, while if by using `theme={theme}`, `theme` is static and removes some overhead. I am also looking for a solution on wrapping programmatically all the components tested in Jest with a ThemeProvider.

src/__tests__/Label.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import 'jest-styled-components'
+import { Label, theme } from '..'
```
hakimelek: I don't think you are using `theme` anywhere here.

.storybook/InputField.js
```diff
@@ -0,0 +1,135 @@
...
+    return (
+      React.cloneElement(React.Children.only(this.props.children), {
+        onChange: this.onChange,
+        color: this.getBorderColor(),
+        info: (!this.state.isValid && this.state.isDirty) ? 'That\'s not 5 letters!' : null,
+        ...this.state
+      })
+    )
+  }
+}
```
hakimelek: Needs a new line here! Prettier,  where are you? :)

.storybook/InputField.js
```diff
@@ -0,0 +1,135 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { withInfo } from '@storybook/addon-info'
+import styled from 'styled-components'
+import { Box, theme, Flex, Text, InputField, Input } from '../src'
```
hakimelek: You are not using `theme` anywhere here.

hakimelek: You are not using `theme` anywhere here.

src/Input.js
byrekt: borders is not a styled component, it's just a function and doesn't have access to the props unless I explicitly pass it in on line 44. @jxnblk, what should I do?

.storybook/InputField.js
byrekt: I think that was copy pasta

src/__tests__/InputField.js
byrekt: Since InputField has children that are dependent on the theme being in context, I need to use ThemeProvider in this case :(

byrekt:
I've updated this PR with changes based on feedback from @hakimelek and @jxnblk

src/Input.js
jxnblk: Yeah, you shouldn't need to import `theme` ever. The way you pass the function to styled-components on line 44 means that styled-components will pass the full `props` object in here, adding in `props.theme` – you'll still need to use this within a ThemeProvider for it to work, but for right now, that's the pattern we're following

src/__tests__/InputField.js
jxnblk: To add some "context" to this discussion 😉  https://github.com/facebook/react/issues/7905

src/Input.js
byrekt: Oh right, forgot, that "theme" I'm importing and using isn't the one from the design-system, it's the helper function from styled system @hakimelek

src/Input.js
```diff
@@ -0,0 +1,55 @@
...
+  border-width: 1px;
+  border-style: solid;
+  border-color: ${theme('colors.borderGray')};
+
+  padding-top: 14px;
+  padding-bottom: 14px;
+  padding-left: 12px;
+  padding-right: 12px;
+
+  ::placeholder {
```
747823: Cool, didn't know this selector existed

jxnblk:
LGTM! This PR fixes #69 and #72

PR #121, opened 2017-10-26
Feature/icon button

jes708:
Breaking https://github.com/pricelinelabs/design-system/pull/111 into pieces to make it easier to review.  Here is the IconButton component to be used inside the banner.


src/IconButton.js
```diff
@@ -0,0 +1,31 @@
...
+import styled from 'styled-components'
+import Icon from './Icon'
+import Button from './Button'
+import PropTypes from 'prop-types'
+
+const TransparentButton = styled(Button)`
+  padding: 0;
+  height: auto;
+  background-color: transparent;
+  color: currentcolor;
```
jxnblk: You should be able to use `color: inherit` here

src/IconButton.js
```diff
@@ -0,0 +1,31 @@
...
+
+  hover: {
+    background-color: transparent;
+  }
+`
+
+const IconButton = props => (
+  <TransparentButton {...props} title={props.title} onClick={props.onClick}>
```
jxnblk: Might be better to destructure the Icon-specific props out and pass the rest of the props to the button `({ name, size, color, ...props })` – also setting `title` and `onClick` here is redundant since you pass `{...props}`

jes708: Cool stuff!

PR #122, opened 2017-10-26
feat(test-setup): remove need to import jest-styled-components

jes708:
remove need to import jest-styled-components in spec files


jxnblk:
So clean :shipit:

PR #125, opened 2017-10-30
Feature/hug

jes708:
Makes the following tweaks:

- Adds border radius to only top of hug
- Provides default option to obscure 1px borders
- Only applies bg and color to Hug's flex header


jxnblk:
Does the border-radius change conflict with #126 ?

jes708:
@jxnblk I'm pretty sure it doesn't.  I did a little experiment to check that.

jxnblk:
LGTM :shipit:

PR #126, opened 2017-10-30
feat(card-radius): default to 2px card border-radius

jes708:
default to 2px card border-radius


.storybook/Card.js
```diff
@@ -2,21 +2,144 @@ import React from 'react'
...
+        p={3}
+        width={200}
+        color="black"
+        bg="white"
+        borderRadius="30px"
+      >
+        XLarge Shadow
+      </Card>
+    </Box>
+  ))
```
jxnblk: Prettier went a little wild here

src/Card.js
```diff
@@ -1,6 +1,8 @@
 import styled from 'styled-components'
 import Box from './Box'
 import PropTypes from 'prop-types'
+import { borderRadius, propTypes } from 'styled-system'
+import { radius } from './theme'
```
jxnblk: I think it'd be better to avoid importing the theme into the components directly. This same value can be picked up from `props.theme.radius` right? If you want to use the styled-system `borderRadius` it picks up from a `radii` object, which I have an example of here: https://github.com/pricelinelabs/design-system/pull/123/files#diff-b985c2b287f97c8e9ceb0647ee1684f4R117

PR #129, opened 2017-10-31
Feature/hug responsive

jes708:
Caution: contains some overlap with #125.  Please review #125 first.

Allows for responsive design and adaptation.  Adds responsive example to storybook.


src/Hug.js
```diff
@@ -1,32 +1,78 @@
...
+          name={icon}
+          size={24}
+          display={iconDisplay}
+        />
+      )}
+      <Text.span fontSize={fontSize}>{props.text}</Text.span>
     </Flex>
-    {props.children}
-  </StyledCard>
+    <BorderConcealer m={showChildBorder ? 0 : `-${childBorderWidth}px`}>
```
jxnblk: I think this might be a case where doing something automatic with React.Children might actually be nice since it seems like the most likely use-case. E.g. if `child.type === Card` then remove the border – what do you think?

src/Hug.js
```diff
@@ -1,32 +1,78 @@
...
+  border-top-left-radius: ${props => props.theme.radius};
+  border-top-right-radius: ${props => props.theme.radius};
+  border-bottom-right-radius: 0;
+  border-bottom-left-radius: 0;
+  overflow: hidden;
+`
+
+const BorderConcealer = styled(Box)`
+  z-index: -1;
+  position: relative;
```
jxnblk: Not crazy about adding a new positioning context here, but with the negative margin approach I guess it's unavoidable. The border could probably be overridden with CSS as well, which would mean less logic in JS. I think `& > * { border: 0; }` should work

jxnblk: These comments are probably better for #125 - would be helpful to use more descriptive PR titles like, *Adjust child border styles in Hug component*

jxnblk:
I'll merge #125 first in case there's any clean up that you need to do on this branch afterward

PR #132, opened 2017-11-02
FormField: Changed import statements so that they don't always try an…

byrekt:
…d pull from the src folder

I was incorrectly importing modules from the /src folder.

### DIFF
--- a/src/InputField.js
+++ b/src/InputField.js
@@ -1,14 +1,20 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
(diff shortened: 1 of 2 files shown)


src/InputField.js
```diff
@@ -1,14 +1,20 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import { Input, Label, Text, Flex, Box, Icon, theme } from '../src'
```
jxnblk: FWIW you could also import from `./index` here

PR #134, opened 2017-11-03
width for background image

daljeetk:
Adding width to background image so an array can be passed to make the background image responsive


src/BackgroundImage.js
```diff
@@ -1,26 +1,29 @@
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
+import { width, responsiveStyle } from 'styled-system'
```
Degron: Why do you import `responsiveStyle` if you don't use it?

src/BackgroundImage.js
```diff
@@ -1,26 +1,29 @@
...
-const height = props => props.height
-  ? { height: props.height }
-  : null
+const height = props => (props.height ? { height: props.height } : null)
 
 const BackgroundImage = styled.div`
-  ${image}
-  ${height}
-  background-color: ${props => props.theme.colors.gray};
+  ${image} ${height} ${width} background-color: ${props =>
```
Degron: I am not sure that I like how it format now.

Degron: @jxnblk What do you think?

jxnblk: I think I'm happy to let the machines decide the best formatting :) We're trying prettier out to stop bikeshedding a little – my guess is that prettier will get better over time and handle some of this embedded CSS syntax a little better in the future

src/BackgroundImage.js
```diff
@@ -1,26 +1,29 @@
...
+  PropTypes.number,
+  PropTypes.string,
+  PropTypes.array
+])
 
 BackgroundImage.propTypes = {
   /** background-image url */
-  image: PropTypes.string
+  image: PropTypes.string,
+  width: numberStringOrArray
```
jxnblk: You can import the propTypes needed for any of the styled-system functions instead of defining them manually, see the Divider component for an example: https://github.com/pricelinelabs/design-system/blob/master/src/Divider.js#L19-L23

jxnblk:
Other than the unused import, this LGTM! Great work! :shipit:

src/BackgroundImage.js
Degron: @daljeetk In this case try put background-color on first place

src/BackgroundImage.js
daljeetk: removed

src/BackgroundImage.js
daljeetk: done

PR #135, opened 2017-11-03
New icons

jxnblk:
This PR is large, but mostly adding SVG icons to the `icons/` folder.

- Moves @odmln's new icons into this repo. He will continue to maintain new icons here
- Adds a ~~`next`~~ `legacy` prop to the Icon component to prefer using legacy icons
- Organizes icons into two folders: `legacy/` and `new/`
- (Internal) stores the legacy icons in a `legacy` object to prevent duplicate keys from clashing

To view the new icons in storybook, run `npm run prepublish && npm start`


icons/new/box checked.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jes708: `box checked.svg` seems like an unusual file name.  Are we cool with spaces in our filenames?  Would the implementation of this look like `<Icon name='box checked' />`

jxnblk: All the filenames are camelCased so it’d be `boxChecked`

package.json
```diff
@@ -78,6 +78,7 @@
     "setupTestFrameworkScriptFile": "./test-setup.js"
   },
   "dependencies": {
+    "css-spring": "^4.1.0",
```
jes708: Are you animating something?

src/Icon.js
```diff
@@ -4,16 +4,22 @@ import { space, color } from 'styled-system'
...
-const oldAndNewIcons = Object.assign({}, icons, {
+const aliases = {
   srollLeft: icons.chevronLeft,
   chevronLight: icons.chevronDown,
   chevronThick: icons.chevronDownThick
-})
+}
+
+const getPath = ({ name, next }) => {
+  if (next) {
```
jes708: I have slight qualms about the nomenclature of the `next` prop, but the documentation does make it clear as to what purpose it serves.  Without documentation, something like `legacy={false}` or `legacySet={false}` would be easy to understand.  If you didn't want to provide a defaultProp for `legacy: true`, then something like `latestVersion` could also be concise.  Food for thought.

package.json
jxnblk: Oh no! Accidentally left that in when trying to demo some things for Zac

src/Icon.js
jxnblk: Yeah, I always struggle with how to name these things. "New" would make sense, but it's obviously a reserved word... Open to suggestions for how to handle this. My other thought was to have a separate component for only the new icons

jxnblk: I put an alternative API for switching between the legacy and new icons here: https://github.com/pricelinelabs/design-system/tree/new-icons-legacy-flag/

Instead of a prop on the Icon component, it uses the ThemeProvider's `legacy` prop, e.g.

```jsx
<ThemeProvider legacy>
  <Icon name='chevronDown' />
</ThemeProvider>
```

Would like to get others thoughts on how we should handle this

jes708: I'll vote for `<Icon legacy />` with a deprecation warning for using legacy

hakimelek: I like the second PR because I think it follows the same paradigm used for all the other properties in the theme. However, it could also be nice to be able to use a `new` icon if it doesn't exist on the `legacy` set and also force an icon to be `new` if the ThemeProvider is set to `legacy`. So, my take is maybe an alternative that mixes the two PRs.

jxnblk: I should clarify that this *only* affects icons where the keys are the same from the legacy set to the new one - e.g. chevrons, calendar, and star

jinchi2013: I like the the legacy as prop on Icon

jxnblk: Yeah, seems like the Icon prop is a more flexible solution. I'm going to change the name from `next` to `legacy` and get a non-breaking version out in a few minutes

src/Icon.js
```diff
@@ -4,16 +4,22 @@ import { space, color } from 'styled-system'
 import PropTypes from 'prop-types'
 import icons from '../icons.json'
 
-// Support re-named icon names too
 // Should be removed eventually after v1.0.0
-const oldAndNewIcons = Object.assign({}, icons, {
+const aliases = {
   srollLeft: icons.chevronLeft,
```
hakimelek: was the `sroll` like this like this before?

jes708: good catch

jxnblk: Looks like it

Issue #68, opened 2017-09-25
Radio component


jxnblk:

odmln:
Zeplin link: https://zpl.io/bJE0RpD

PR #139, opened 2017-11-10
Feature/issue #68 radio button

tornglintaffychen:
screen shot 2017-12-02 at 9 45 51 pm

(Selected, Not Selected but hovered, Disabled)


src/__tests__/RadioButton.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import { mount } from 'enzyme'
```
jes708: I think you should be able to swap out all of your `mount`s with `shallow`s, since you're only testing the root wrapper.

src/RadioButton.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+`
+
+const RadioIconWrap = styled.span`
+  position: absolute;
+  height: 24px;
+  top: -2px;
+`
+
+const RadioButtonText = styled.span`
```
jes708: This isn't necessary.  Can be written inline as `<Text.span pl={4}>`

src/RadioButton.js
```diff
@@ -0,0 +1,68 @@
...
+      <RadioButtonText>{radioButtonText}</RadioButtonText>
+    </RadioButtonWrap>
+  )
+}
+
+const RadioButtonWrap = styled.label`
+  position: relative;
+  cursor: ${props => (props.isSelected || props.disabled ? null : 'pointer')};
+
+  font-weight: ${props => (props.isSelected ? '600' : '400')};
```
jes708: might be better here to reference `bold` and `regular` from `theme`

src/RadioButton.js
```diff
@@ -0,0 +1,68 @@
...
+
+  return (
+    <RadioButtonWrap
+      className={className}
+      onClick={handleClick}
+      isSelected={isSelected}
+      disabled={disabled}
+      theme={theme}
+    >
+      <RadioIconWrap>
```
jes708: This could possibly be a use case for `<IconButton />`.  Take a look at the documentation and see if you think it could make this a little simpler.  It would at least include a pointer cursor, and might require less wrapping than you have currently

tornglintaffychen: If `IconButton` takes `children` so we can pass in the `Text` in that would be good. 
But right now,  I have to have `label` wrapper to wrap around `Icon` and `Text` and all the hover, cursor logics are done in the wrapper.

src/RadioButton.js
747823: To expand on Jon's comment, you could try doing away with the positioning stuff by adding a Flex.

So i.e.
```
<RadioButtonWrap>
  <Flex>
    <Icon name={radioIconName} size={24} color={radioIconColor} />
     <Text pl={4}>{radioButtonText}</Text>
  </Flex>
</RadioButtonWrap>
```

Then adjust `mt` on the Icon to make it line up vertically. That should make the icon sit on the left side of the text

src/RadioButton.js
```diff
@@ -0,0 +1,68 @@
...
+        <Icon name={radioIconName} size={24} color={radioIconColor} />
+      </RadioIconWrap>
+      <RadioButtonText>{radioButtonText}</RadioButtonText>
```
jes708: Nonblocking: you could consider designating `radioButtonText` as `children` to allow for

```
<RadioButton>Hello</RadioButton
<RadioButton>Bonjour</RadioButton
<RadioButton>Hola</RadioButton
```

.storybook/RadioButton.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react'
+import { storiesOf, action } from '@storybook/react'
+import { RadioButton } from '../src'
+
+storiesOf('RadioButton', module).add('3 states', () => (
+  <div>
+    <di>
```
jes708: typo `di`

tornglintaffychen: oops, that's why storybook build failed. thanks!

src/RadioButton.js
```diff
@@ -0,0 +1,66 @@
...
+}
+
+const RadioButtonWrap = styled.label`
+  position: relative;
+  cursor: ${props => (props.isSelected || props.disabled ? null : 'pointer')};
+
+  font-weight: ${props =>
+    props.isSelected ? props.theme.bold : props.theme.regular};
+
+  color: ${props => (props.disabled ? props.theme.colors.borderGray : null)};
```
jes708: I think you might see some IE bugs arise from these `null`'s.  IE CSS can get screwed up, because it reads this as `color:;`.  To avoid errors, rewrite as `${props => (props.disabled ? `color: ${props.theme.colors.borderGray};` : null)}`

broox9: you could also use the `initial` or `inherit` value instead of null

src/RadioButton.js
```diff
@@ -0,0 +1,71 @@
...
+      className={className}
+      onClick={handleClick}
+      isSelected={isSelected}
+      disabled={disabled}
+      theme={theme}
+    >
+      <RadioIconWrap>
+        <Icon name={radioIconName} size={24} color={radioIconColor} />
+      </RadioIconWrap>
+      <Text.span pl={4}>{radioButtonText}</Text.span>
```
jxnblk: I think we'll want to have a low-level primitive for `<input type='radio'>` for composition. Ideally we'd have a Radio component that can be nested inside the Label – e.g.

```jsx
<Label>
  <Radio />
  Text for the radio label as children
</Label>
```
```

File: src/RadioButton.js [hunk 9]
```diff
@@ -0,0 +1,71 @@
...
+    onClick()
+  }
+
+  return (
+    <RadioButtonWrap
+      className={className}
+      onClick={handleClick}
+      isSelected={isSelected}
+      disabled={disabled}
+      theme={theme}
```
jxnblk: Whatever the `<input>` element is should probably take the full `props` object, or rest props if some stylistic props need to be used. In this case `isSelected` looks like a proxy for the `checked` attribute, and `handleClick` is renaming `onClick` – React has built-in props for handling form elements and we should stick to those for HTML semantics, React's internal prop-type checking, familiarity with existing APIs, and for interoperability with other libraries (e.g. validation)

src/RadioButton.js
```diff
@@ -0,0 +1,71 @@
...
+      <RadioIconWrap>
+        <Icon name={radioIconName} size={24} color={radioIconColor} />
+      </RadioIconWrap>
+      <Text.span pl={4}>{radioButtonText}</Text.span>
+    </RadioButtonWrap>
+  )
+}
+
+const RadioButtonWrap = styled.label`
+  position: relative;
```
jxnblk: We should generally avoid setting positioning on the visible elements in design-system components since it would create a new positioning context that would interfere with the consuming app's layout. Instead, it's common to accessibly hide the `<input>` element with something like:

```css
position: absolute;
opacity: 0;
z-index: -1;
```

src/RadioButton.js
```diff
@@ -0,0 +1,71 @@
...
+  cursor: ${props => (props.isSelected || props.disabled ? null : 'pointer')};
+
+  font-weight: ${props =>
+    props.isSelected ? props.theme.bold : props.theme.regular};
```
jxnblk: I don't know if @odmln intended the font-weight change, but since font-weight affects the width of the element, generally it's better not to change based on state

src/RadioButton.js
```diff
@@ -0,0 +1,71 @@
...
+  onClick,
+  isSelected,
+  disabled,
+  radioButtonText
+}) => {
+  const radioIconName = isSelected ? 'radioFilled' : 'radio'
+  const radioIconColor = isSelected ? 'blue' : 'borderGray'
+
+  function handleClick() {
+    if (disabled || isSelected) return
```
jxnblk: Using the native `disabled` and `checked` attributes and React's built-in `onChange` handler prevents the event from being called by default, so this can be cut

src/__tests__/RadioButton.js
```diff
@@ -0,0 +1,66 @@
...
+    const json = renderer
+      .create(
+        <RadioButton theme={theme} disabled radioButtonText="Option Disabled" />
+      )
+      .toJSON()
+    expect(json).toMatchSnapshot()
+  })
+})
+
+describe('RadioButton onClick', () => {
```
jxnblk: Using the built-in React form attributes means you don't have to test custom handlers, so this could be cut

.storybook/Radio.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react'
+import { storiesOf, action } from '@storybook/react'
+import { Radio, Label } from '../src'
+
+storiesOf('Radio', module).add('3 states', () => (
+  <div onChange={action('changed')}>
+    <Label fontSize="14px">
+      <Radio checked radioButtonText="Option Selected" />
```
jxnblk: I think maybe the `radioButtonText` prop isn't used anymore, is it?

src/Radio.js
```diff
@@ -0,0 +1,57 @@
...
+
+const Radio = props => {
+  const { className, checked, disabled, name, value } = props
+
+  const radioIconName = checked ? 'radioFilled' : 'radio'
+
+  const testID = checked ? 'SELECTED' : disabled ? 'DISABLED' : 'NOT_SELECTED'
+
+  return (
+    <RadioWrap className={className} checked={checked} disabled={disabled}>
```
jxnblk: I think we want to avoid exposing `className` outside of this component since we're using styled-components, right?

jxnblk:
This looks great, thanks for working on this! The only last request I would suggest is to add a simple markdown file to the `docs/` folder with a usage example so that we keep up with documentation

jxnblk:
This looks great! Thanks for working on this! :shipit:

hakimelek:
Fixes #68

Issue #131, opened 2017-11-02
Dropdown/Popover component


hakimelek:

BeniCheni:
Discussed and closed this #131. Further progress will be tracked in #352.

PR #141, opened 2017-11-14
WIP: Implement Dropdown Components; Misc. refactoring

BeniCheni:
Per #131, this PR implements the dropdown components highlights in this doc:
https://github.com/pricelinelabs/design-system/blob/select/docs/Dropdown.md#dropdown-components

##### Dropdown - Collapsed
screen shot 2017-11-14 at 11 15 19 am

##### Dropdown -  Expanded With Simple Options
screen shot 2017-11-14 at 11 16 14 am


.storybook/Select.js
```diff
@@ -0,0 +1,32 @@
...
+    <Box width={220}>
+      <Select># Rooms</Select>
+    </Box>
+  ))
+  .add('Select & SelectField Composition - Expanded', () => (
+    <Box width={220}>
+      <Select isExpanded># Rooms</Select>
+      <SelectFieldCard borderWidth={0} boxShadowSize="lg" mt={2}>
+        <SelectField isSelected>1 Room</SelectField>
+        <SelectField>2 Rooms</SelectField>
```
jxnblk: So, I'm guessing this isn't the `<select>` element – otherwise these would be `<option>` elements right?

src/Label.js
```diff
@@ -1,12 +1,12 @@
 import styled from 'styled-components'
-import { space, fontSize, color, propTypes } from 'styled-system'
+import { color, fontSize, propTypes, space, textAlign } from 'styled-system'
 
 const Label = styled.label`
   font-size: 10px;
```
jxnblk: Should move this to a `defaultProp` if you're adding the `fontSize` function

src/Select.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { theme } from 'styled-system'
+
+import { Box, Flex, Icon, Label } from '../src'
+
+const parseChevronIcon = isExpanded => {
+  const ChevronIcon = isExpanded
+    ? Icon.extend`
+        transform: rotate(180deg);
```
jxnblk: We should use the `chevronUp` and `chevronDown` icons here instead of rotating – the icons might be optically aligned, rather than mathematically centered

src/Select.js
```diff
@@ -0,0 +1,64 @@
...
+
+const SelectBox = Box.extend`
+  cursor: pointer;
+`
+
+const SelectLabel = Label.extend`
+  cursor: pointer;
+`
+
+const Select = props => (
```
jxnblk: I think we should reserve the component name `Select` for the actual `<select>` element since React handles form elements differently and this has a completely different API from the HTML element – e.g. `onChange` `onFocus` `onBlur`

jxnblk: I'd suggest that we call this component the `Dropdown` – since that's what it's commonly called – see https://getbootstrap.com/docs/4.0/components/dropdowns/

src/Select.js
```diff
@@ -0,0 +1,64 @@
...
+  <SelectFlex justify="center" align="center">
```
jxnblk: We'll probably need to add a lot of ARIA attributes to compensate for accessibility in this component – here's a good starting place for research: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA

src/Select.js
```diff
@@ -0,0 +1,64 @@
...
+    <SelectBox pl={5} pr={3}>
+      {parseChevronIcon(props.isExpanded)}
+    </SelectBox>
+  </SelectFlex>
+)
+
+Select.displayName = 'Select'
+
+Select.propTypes = {
+  isExpanded: PropTypes.bool,
```
jxnblk: Since dropdown components are *very* similar to the `<summary>` and `<details>` HTML elements, I tend to favor naming props like this after the HTML `open` attribute – see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details

src/Select.js
```diff
@@ -0,0 +1,64 @@
...
+  children: PropTypes.oneOfType([
```
jxnblk: Why are you redefining React's children propType here?

src/SelectField.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { Box, Text } from '../src'
+
+const parseIsSelected = props => {
+  return `
+    ${props.isSelected
+      ? `background-color: ${props.theme.colors['blue']};`
+      : null}
```
jxnblk: Just FYI, styled-components also accepts style objects when you're writing functions like this, which can make the code a little cleaner and avoid the need to nest template literals – the [polished](https://polished.js.org) library does this

src/Select.js
```diff
@@ -0,0 +1,64 @@
...
+  border-style: solid;
+  border-color: ${theme('colors.borderGray')};
+`
+
+const SelectBox = Box.extend`
+  cursor: pointer;
+`
+
+const SelectLabel = Label.extend`
+  cursor: pointer;
```
jxnblk: I'd avoid changing the Label styles here, if we want to change the default cursor here, we might want to change that across the board – either way that's probably something we wanna run by the design team

src/Select.js
```diff
@@ -0,0 +1,64 @@
...
+`
+
+const Select = props => (
+  <SelectFlex justify="center" align="center">
+    <SelectBox pl={3}>
+      <Icon name="key" size={24} color="blue" />
```
jxnblk: I'm guessing this icon should be configurable and optional?

src/Select.js
BeniCheni: Sounds good. I used transform rotation because I was observing different between `chevronUp` and `chevronDown`. I wanted to avoid the graphic difference for the icons.

Updated to use `chevronUp` for the `isExpanded` style. Will check with the designer to assess the icon difference to see if there's a real concern.

src/Label.js
BeniCheni: Good catch. The original implementation of `Label` does indeed have default prop for `fontSize`, and this `font-size` CSS were already part of the styled component's CSS.

Updated to remove the `font-size` CSS, and only use `fontSize` default prop.

.storybook/Select.js
jxnblk: No, the select is a completely separate component and should not be confused with this, which is a Dropdown (this is also the term used in the Sketch document) – we have separate use cases for both. For Select, we'll want a styled [`<select>`](https://html.spec.whatwg.org/#the-select-element) – see my comments below

BeniCheni: @jxnblk ,
Caught up with the rest of the review comments, and spoke to @hakimelek to avoid confusion also. Gotcha. Will rename the components to `Dropdown` and `DropdownField` for this PR. No más confusion between dropdown and select.

Will mention you in comments for review again, once code is updated. 🙏

src/Select.js
BeniCheni: @jxnblk, great feedback on the potential support of the ARIA attributes. I've created #152 issue as a follow-up item to this PR.

src/Select.js
BeniCheni: Updated the component name to `Dropdown`.

src/Select.js
BeniCheni: Thanks for the reference. Updated the prop name to `open` to match HTML native attribute.

src/Select.js
BeniCheni: @jxnblk 
I am thinking that the children of the `Dropdown` element can be either a node (i.e. object, number, or string; pretty any type), or an array of nodes. So I went ahead and defined the prop type with that thought.

Is there any side effect with my definition of the children prop type? What are the pros and cons if the children prop type is not defined?

src/SelectField.js
BeniCheni: Great feedback. Took a look at the "polished" library to learn about their pattern. Updated here to a pattern closer to the suggested idea. Please review again.

src/Select.js
BeniCheni: I took a step back and review the pattern and UX. Agree that for now, we don't wanna set a cursor style on `Label` (essentially an HTML native `<label>` tag). Let me talk to @chrisdroukas & @odmln about if the `Label` should have a default cursor style for all, or should it not.

src/Select.js
BeniCheni: Totally nice suggestion. Updated to offer a `withIcon` boolean prop to allow client to opt in to use the icon, or opt out. Thanks~ ✌️

BeniCheni:
@jxnblk,

Thanks for your first round of review. Updated and/or replied your comments. Please review again and see if the updates look okay?

src/Dropdown.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { theme } from 'styled-system'
+
+import { Box, Flex, Icon, Label } from '../src'
+
+const DropdownFlex = Flex.extend`
+  cursor: pointer;
+  height: 48px;
```
jxnblk: Should this be `min-height`? What happens when text wraps?

src/Dropdown.js
```diff
@@ -0,0 +1,50 @@
...
+  border-width: 1px;
+  border-style: solid;
+  border-color: ${theme('colors.borderGray')};
+`
+
+const Dropdown = props => (
+  <DropdownFlex justify="center" align="center">
+    {props.withIcon && (
+      <Box>
+        <Icon name="key" size={24} color="blue" />
```
jxnblk: I suspect this would be easier to handle with composition, e.g.:
```
<Dropdown>
  <Icon name='key' />
  Dropdown Label
</Dropdown>
```

Theres a similar discussion here: https://github.com/pricelinelabs/design-system/issues/155

.storybook/Dropdown.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Box, Card, Dropdown, DropdownField } from '../src'
+
+const DropdownMenu = Card.extend`
+  height: 200px;
+  overflow: scroll;
+`
```
jxnblk: It looks like DropdownMenu is meant to be shown and hidden by the end user – seems like it would be good to make this as part of the same "compound component" – I could imagine the API looking something like:

```jsx
<Dropdown open={dropdownOpen}>
  <DropdownButton>
    Dropdown Label
  </DropdownButton>
  <DropdownMenu>
    <DropdownItem>Item One</DropdownItem>
    <DropdownItem>Item One</DropdownItem>
    <DropdownItem>Item One</DropdownItem>
    <DropdownItem>Item One</DropdownItem>
  </DropdownMenu>
</Dropdown>
```

Mostly just stealing the naming convention from Bootstrap: https://getbootstrap.com/docs/4.0/components/dropdowns/

.storybook/Dropdown.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Box, Card, Dropdown, DropdownField } from '../src'
+
+const DropdownMenu = Card.extend`
+  height: 200px;
+  overflow: scroll;
```
jxnblk: When setting `overflow: scroll` it's good practice to also set `-webkit-overflow-scrolling: touch` for iOS

src/Dropdown.js
BeniCheni: Nice feedback. Updated to use `min-height` to account for long label name that wraps.

src/Dropdown.js
BeniCheni: @jxnblk, reviewed the concern of #155 and updated the story (`.storybook/Dropdown.js`) to implement a more compound composition. Thanks for the nice feedback. Plz review again.

.storybook/Dropdown.js
BeniCheni: Great names for `DropdownButton` & `DropdownItem`. Updated to adapt those names, mirroring what Bootstrap doc suggests.

.storybook/Dropdown.js
BeniCheni: Anther great CSS rule to learn. Thanks. Updated.

BeniCheni:
@jxnblk , thanks for the 2nd round of review and suggestion/encouragement for more improvements for Bootstrap pattern & ARIA support. Please review again at your convenience.

The updated PR has taken inspiration of the [Bootstrap Single button dropdowns](https://getbootstrap.com/docs/4.0/components/dropdowns/#single-button-dropdowns):

- `Dropdown`,

- `DropdownButton`- with more defined tag attributes such as `type="button"`

- `DropdownMenu`

- `DropdownItem`

- ARIA support with `aria-expanded`, `aria-haspopup`, `aria-labelledby`

src/DropdownButton.js
```diff
@@ -15,7 +15,10 @@ const DropdownButton = OutlineButton.extend.attrs({
   height: 100%;
   min-height: 44px;
   cursor: ${props => props.cursor};
-  border-color: ${theme('colors.borderGray')};
+  border-top-color: ${theme('colors.borderGray')};
```
hakimelek: why not `${props => props.theme.colors.borderGray}` ?

BeniCheni: @hakimelek , good catch. Updated. There's no special use case here to use the `theme(...)` utility. More correct to use the color(s) provided by client's theme provider.

jxnblk: The `styledSystem.theme` utility is an existential getter, which won't throw if `theme.colors` is undefined; it can also accept a fallback value – that's the only difference between the syntax change here

hakimelek: Yes, sorry I was not aware of what `theme` from `styled-system` is doing! It might be better to adopt it as an alternative pattern to use the `theme` props

BeniCheni: The fallback value of the `systemStyle.theme` utility sounds interesting. Gonna try out the theme utility with a fallback value in the next design system feature, if there's a use case of theme. Thanks for the educational information, @jxnblk . +1

jinchi2013:
@BeniCheni @jxnblk 
hi, how's it going with this PR? We have plan to use it in this sprint. Any update?

BeniCheni:
@jinchi2013, I've been allocated to other projects that don't required the `Dropdown` immediately. So I couldn't spend much time following up my own WIP progress. Please don't depend your sprint plan on this WIP version of the `Dropdown`. Hope you can find a plan B not to depend on this WIP `Dropdown`.

jxnblk:
@jinchi2013 can you elaborate on the requirement? This PR for the Popover might be close to ship, and I imagine it might eventually be used for the Dropdown component #223

jinchi2013:
@jxnblk  here is it
screen shot 2018-02-20 at 6 10 52 pm

jxnblk:
@jinchi2013 For the *Add a Hotel* part, I think that should be the Select component for accessibility reasons. You won't have the same look as the mockup above, but I think that should be fine.

The Dropdown/Popover is more for modal things that break out of the document flow and contain buttons, menus, etc. – like the TravelerSelection or the Calendar components

jinchi2013:
@jxnblk OK, got you, will discuss this will Jake. Thanks

BeniCheni:
Since the scope changed in the project that was going to adapt the dropdown component, closing this PR as no further update to complete this WIP PR>

PR #145, opened 2017-11-14
feat(banner-color): update dark colors for banner - allow text node

jes708:
update light banners to corresponding dark text colors
allow node as PropType for banner text


.storybook/Banner.js
```diff
@@ -735,7 +721,19 @@ storiesOf('Banner', module)
...
       />
     </Box>
   ))
+  .add('With text node', () => (
+    <Banner
+      textAlign="center"
+      mb={2}
+      p={2}
+      header="custom text"
+      text={
```
jxnblk: Is there a good reason to create a custom prop rather than use `props.children` here? I think custom props like this only make sense when you want to poke "holes" into a component – e.g. `leftColumn`/`rightColumn`

jes708: I'll allow either `text` or `children` since removing `text` at this point would be a breaking change.

hakimelek:
So have been having a chat with @krevat about this, and I think this is the behavior that we would like to have:
pasted image at 2017_12_04 01_38 pm

I am not a fan of having the `text` prop behave as a render prop and can take any children and apply the default style. I feel like the `Banner` should be a `Box` that take children and check if one of the child is of type `Text` it would default it to the default banner font size. The `<InputField>` component is a good example for that. 

@odmln @chrisdroukas do you have any thoughts about this?

jxnblk:
I think I agree with the comment above, an API like the following seems simpler for anything beyond just a string of text in the Banner:

```jsx
// Default styles
<Banner>
  Hello
</Banner>
```

```jsx
// Custom composition
<Banner>
  <Icon name='check' />
  <Text fontSize={4}>Hello</Text>
</Banner>
```

odmln:
Yeah i agree, treating it the same way we handled InputField seems like a smart move.

hakimelek:
maybe we should simply add `children` to the `Banner` and then can add another PR about making it smarter about `Text` and `Icon` to address #163.

jxnblk:
Yeah, it seems like #163 and this PR are overlapping a bit. Should we discuss in person and do a code walkthrough of each sometime this week?

hakimelek:
@jxnblk have been chatting with @krevat and @jes708 this morning and we decided to move on with #166, the changes will include adding support for `children`, update dark colors for banner and update the docs. The `Banner` should still take `text` prop for backward compatibility. 

closing this in favor of #166 for now.

Issue #67, opened 2017-09-25
Checkbox component


jxnblk:

jxnblk:
@odmln @msafari be sure to sync up/put some screenshots, example css, etc on these form elements

odmln:
Zeplin link: https://zpl.io/bJE0RpD

msafari:
Thanks @odmln

PR #146, opened 2017-11-15
Feature/checkbox

broox9:
Adding a checkbox component (no label) based on design: https://app.zeplin.io/project/59cab1381d42232d25e4fdb6/screen/59fb90f364de93233e6d02e3

I didn't use any `styled-system` stuff because I'm still familiarizing myself with  it.  If it would benefit the code / dev ease of use, let me know
screen shot 2017-12-21 at 10 20 24 am


src/Checkbox.js
```diff
@@ -0,0 +1,111 @@
...
+        {this.state.isChecked ? (
+          <Icon
+            name="check"
+            size={size - 4}
+            color={currentTheme.colors.white}
+          />
+        ) : null}
+
+        <input
+          type="checkbox"
```
jxnblk: We spoke about this at lunch, but I think all the base form elements should accept all the non-style-related props as "passthrough" props – e.g. letting others use this Checkbox component in exactly the same way the `<input type='checkbox'>` element works. This should be less API surface area for new people to learn, and means you wouldn't have to define new props for things like `checked`, `disabled`, `onChange`, or even styled-components' `innerRef` prop

src/Checkbox.js
```diff
@@ -0,0 +1,111 @@
...
+          name={name}
+          id={id}
+          value={value}
+          disabled={isDisabled}
+          defaultChecked={this.state.isChecked}
+          style={{ display: 'none' }}
```
jxnblk: For accessibility reasons, it's better to *not* set `display: none` on the underlying form element, instead, using `position: absolute` and `opacity: 0` is one way to accessibly hide form elements. An example can be seen here: http://wtfforms.com

src/Checkbox.js
```diff
@@ -0,0 +1,111 @@
...
+import Flex from './Flex'
+import Box from './Box'
+import theme from './theme'
+
+class Checkbox extends React.Component {
+  constructor({ value, isChecked }) {
+    super()
+    this.state = {
+      value,
+      isChecked
```
jxnblk: By using React's form props, you can make this component completely stateless, which means it could be used as either a controlled or uncontrolled form element

src/Checkbox.js
```diff
@@ -0,0 +1,111 @@
...
+        isChecked={this.state.isChecked}
+        isDisabled={isDisabled}
+        size={size}
+        theme={currentTheme}
+        data-test="design-system-checkbox"
+        onClick={() => this.handleClick()}
+      >
+        {this.state.isChecked ? (
+          <Icon
+            name="check"
```
jxnblk: I *think* @odmln intended this component and the Radio component to use two different icons for the checked and unchecked states – `box` and `boxChecked` I believe... https://pricelinelabs.github.io/design-system/?selectedKind=Icon&selectedStory=Icons&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel

jxnblk:
Looks great so far! I don't think there's probably much benefit in using styled-system for this particular component, so I think that's completely fine

src/Checkbox.js
broox9: true, shame on me for not having accessibility in mind

.storybook/Checkbox.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import styled from 'styled-components'
+import { storiesOf, action } from '@storybook/react'
+import { Checkbox, theme } from '../src'
+
+storiesOf('Checkbox', module).add('Checkbox states', () => (
+  <div>
+    <section>
+      <h4>Checkbox</h4>
```
BeniCheni: Trivial, but it would be a good, consistent idea if your story also uses components from design system, instead of `div`, `section`, `h4. `Box` and `Heading` are available. `section` tag can be achieve with `. withComponent` from styled-components API.

src/Checkbox.js
```diff
@@ -0,0 +1,111 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import Icon from './Icon'
+import Flex from './Flex'
+import Box from './Box'
+import theme from './theme'
```
BeniCheni: If you like single `import` statement for your imported components, maybe you can try out a single import with components in alphabetical order in title case (i.e. `Box`), and then other artifacts such as `theme` with lowercase? (the order is totally subjective opinion, but just thought to share to at least have the components alphabetically sorted in the import:
```
import { Box, Flex, Icon, theme } from '../src'
```

src/Checkbox.js
```diff
@@ -0,0 +1,111 @@
...
+class Checkbox extends React.Component {
+  constructor({ value, isChecked }) {
+    super()
+    this.state = {
+      value,
+      isChecked
+    }
+  }
+
+  handleClick() {
```
BeniCheni: Think the UI interaction code such as `handleClick` should be handled off to the containing app to implement either controlled or uncontrolled checkbox via state. The primitive components in design system seem to be pure UI components as stateless.

BeniCheni: Since the design system components are stateless, peophaps you can take a look at the ‘onClick’ implementation in ‘CloseBox’, for the similar click on the checkbox? It’s really up to the story or the client app to define the code for clicking on the checkbox, pagans?
https://github.com/pricelinelabs/design-system/blob/master/src/CloseButton.js#L13

broox9: yeah, "checkAction"  is that function.  maybe I should rename it.  I have to have another function call it internally though. The checkbox you click on is really a facade, and I want to return the actual uncontrolled state, name, and value of the real hidden checkbox.

src/Checkbox.js
broox9: *derp*

broox9:
ugh, now tests aren't passing

src/Checkbox.js
```diff
@@ -0,0 +1,118 @@
...
+          defaultChecked={checked || defaultChecked}
+          {...restProps}
+          onChange={e => this.handleClick(e)}
+          style={{ opacity: 0, zIndex: -1 }}
+        />
+        <StyledCheckbox
+          data-test="checkbox-facade"
+          disabled={disabled}
+          size={size}
+          theme={currentTheme}
```
hakimelek: Same here, I don't see the need to import `theme` and pass it to your styled component. `ThemeProvider` should take of that for you. If you would like to import `theme` outside of a styled component you can wrap your component with a `withTheme` HOC.

src/Checkbox.js
```diff
@@ -0,0 +1,115 @@
...
+        htmlFor={name}
+        data-test="design-system-checkbox"
+        disabled={disabled}
+      >
+        <input
+          type="checkbox"
+          name={name}
+          id={this.props.id || name}
+          disabled={disabled}
+          defaultChecked={checked || defaultChecked}
```
jxnblk: We should probably keep the props passed in to the `<input>` the same, it seems like this would affect how `checked` and `defaultChecked` work, right?

src/Checkbox.js
```diff
@@ -0,0 +1,115 @@
...
+          {...restProps}
+          onChange={e => this.handleClick(e)}
```
jxnblk: I think we discussed this, but I think it's simpler to pass all props in to the `<input>` so that `onChange`, `onClick`, `onBlur`, etc. all work as expected – did you run into other issues with this since we last spoke?

src/Checkbox.js
```diff
@@ -0,0 +1,115 @@
...
+          style={{ opacity: 0, zIndex: -1, position: 'absolute' }}
```
jxnblk: 👍

src/Checkbox.js
```diff
@@ -0,0 +1,115 @@
...
+        />
+        <StyledCheckbox
+          data-test="checkbox-facade"
+          disabled={disabled}
+          size={size}
+        >
+          <Icon
+            name="check"
```
jxnblk: Per talking to @odmln, I think this should be either the `boxChecked` or `boxEmpty` icon based on `props.checked`, which should eliminate the need to use the StyledCheckbox component

src/Checkbox.js
```diff
@@ -0,0 +1,115 @@
...
+            size={size - 4}
+            color={currentTheme.colors.white}
+          />
+        </StyledCheckbox>
+        {children}
+      </StyledLabel>
```
jxnblk: Similar to the Radio component in #139 it'd be great to keep the label element separate so that the Checkbox component is more composable, e.g.:

```jsx
<Label>
  <Checkbox {...checkboxProps} />
  Check it
</Label>
```

src/Checkbox.js
broox9: the main issue I have is that the display of the checkbox is separate from the actual element itself.  I have to change the display based on the native state of the checkbox, which I cannot reference without giving the component some kind of state itself

hakimelek:
Should resolve #67.

PR #149, opened 2017-11-15
Reduce confusion when using hr with flex-direction: column

jes708:
Add default margins to reduce confusion when using hr with flex-direction: column


src/Divider.js
```diff
@@ -11,7 +11,9 @@ const Divider = styled.hr`
 Divider.displayName = 'Divider'
 
 Divider.defaultProps = {
-  borderColor: 'borderGray'
+  borderColor: 'borderGray',
+  ml: 0,
+  mr: 0
```
jxnblk: Cool approach – I think setting `width: 100%` also works, but this is totally fine

jes708: I tried w={1} which negated the left and right margins.  This solution seemed to fix the issue without causing any conflicts with other space props.

PR #153, opened 2017-11-16
Added new icons + modified old icons

odmln:
Added 40+ new icons and made changes to a few old icons.


icons/new/attention.svg
```diff
@@ -1,9 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
jxnblk: If anyone was using this icon, this would mean a breaking change – did you rename this one?

odmln: I think I removed it by accident :/

icons/new/cruise.svg
```diff
@@ -1,12 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
hakimelek: @odmln should we put this back?

icons/new/zoom out.svg
```diff
@@ -1,10 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
hakimelek: Same for this one

hakimelek: Renamed to zoomOut

icons/new/cruise.svg
hakimelek: Renamed to cruises.svg

PR #154, opened 2017-11-17
feat(include-theme): end the testing nightmare

jes708:
Include theme in *almost* every component, to prevent nightmarish missing nested theme.

This issue had been especially troublesome when testing with `mount` and `getState`.  This should prevent the issue, and still allow ThemeProvider precedence.


src/InputField.js
```diff
@@ -75,4 +64,21 @@ const InputField = ({ id, label, color, icon, info, ...props }) => {
...
+  label: PropTypes.string.isRequired,
+  /** The optional icon that shows up on the right side of the element. */
+  icon: PropTypes.string,
+  /** Optional border and icon color. */
+  color: PropTypes.string,
+  /** Additional content to go in the info drawer below the element. This will be replaced by a tooltip later probably.  */
+  info: PropTypes.string
+}
+
+InputField.defaultProps = {
```
jxnblk: I don't *think* we need theme provided to composite components, do we? No big deal

jes708: That should be true as long as there are no dependencies on `props.theme`. I've removed a few unnecessary default themes!

src/__tests__/__snapshots__/Banner.js.snap
```diff
@@ -118,6 +118,138 @@ exports[`Banner renders with blue bg 1`] = `
       height={24}
       mr={3}
       mt="-2px"
+      theme={
```
jxnblk: Didn't think about it before, but this definitely makes the snapshots a lot bigger 😅

PR #158, opened 2017-11-20
Use fractional max-width values in Hide component - fixes #151

### LINKED ISSUE #151: Refactor Hide to not use max-width
`<Hide />` should use `min-width` only breakpoints to prevent gaps.  There is a gap at 768px for instance.

jxnblk:
Currently, there is a gap between breakpoints in the Hide component, which means at specific viewport widths, both `xs` and `sm` will be set to `display: none`. Generally, I'd advise against using the Hide component to show different elements *across* breakpoints, but this fix allows you do to so.

### DIFF
--- a/src/Hide.js
+++ b/src/Hide.js
@@ -3,14 +3,21 @@ import PropTypes from 'prop-types'
 import Box from './Box'
 import theme from './theme'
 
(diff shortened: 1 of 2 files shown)


jinchi2013:
Looks good to me!

PR #162, opened 2017-11-28
feat(InputField): add InputField component

byrekt:
This PR is to add support for a compound version of the InputField component. What this allows is the consumer will have more flexibility in the appearance of the component. 

Please view https://github.com/pricelinelabs/design-system/blob/feature/compound-input-components/docs/InputField.md for more information.

image

feat(InputField): replace old InputField with backwards compatibile compound component

feat(input-field): add unit tests for InputField component

docs(input-field): add documentation for InputField component

### DIFF
--- a/.storybook/InputField.js
+++ b/.storybook/InputField.js
@@ -2,118 +2,161 @@ import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { withInfo } from '@storybook/addon-info'
 import styled from 'styled-components'
-import { Box, Flex, Text, InputField, Input } from '../src'
+import { Box, Flex, Text, InputField, Label, Input, Icon } from '../src'
 
 storiesOf('InputField', module)
   .add(
-    'InputField component',
-    withInfo({
-      inline: true,
-      text:
-        'This component renders an input element with a built-in label and optional icon. It makes use of the Input and Label components.'
-    })(() => (
-      <Box width={500} pr={3}>
-        <Text bold pt={4} pb={2}>
-          Hey, ya turkey! Enter a word that has 5 characters in it.
+  'InputField component',
+  withInfo({
+    inline: true,
+    text:
+      'This component renders an input element with a built-in label and optional icon. It makes use of the Input and Label components.'
+  })(() => (
+    <Box width={500} pr={3}>
+      <Text bold pt={4} pb={2}>
+        Hey, ya turkey! Enter a word that has 5 characters in it.
         </Text>
-        <InputController>
-          <InputField
-            icon="amenityPool"
-            label="A 5 Letter Word"
-            id="form-field-3"
-            placeholder="e.g. Berry"
-          />
-        </InputController>
-      </Box>
-    ))
+      <InputController>
+        <InputField onChange={() => {}}>
+          <Label>5 Letter Word</Label>
+          <Icon name="amenityPool" />
+          <Input id="form-field" placeholder="Enter a 5 letter word" />
+          <Icon name="fitness" />
+        </InputField>
+      </InputController>
+    </Box>
+  ))
   )
   .add(
-    'Sample States',
-    withInfo('Renders a styled Input element')(() => (
-      <div>
-        <Flex width={900}>
-          <Box width={1 / 2} pr={3}>
-            <Text bold pt={4} pb={2}>
-              Credit Card
+  'Sample States',
+  withInfo('Renders a styled Input element')(() => (
+    <div>
+      <Flex width={900}>
+        <Box width={1 / 2} pl={3}>
+          <Text bold pt={4} pb={2}>
+            Placeholder with Chevron
             </Text>
-            <InputField
-              icon="locked"
-              label="Card number"
-              id="form-field-3"
-              placeholder="0000 0000 0000 0000"
-            />
-          </Box>
-          <Box width={1 / 2} pl={3}>
-            <Text bold pt={4} pb={2}>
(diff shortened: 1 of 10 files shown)


src/InputField.js
```diff
@@ -9,72 +9,144 @@ import Label from './Label'
...
-  id: PropTypes.string.isRequired,
-  /** Text to be used as the label for the input element. */
-  label: PropTypes.string.isRequired,
-  /** The optional icon that shows up on the right side of the element. */
-  icon: PropTypes.string,
-  /** Optional border and icon color. */
-  color: PropTypes.string,
-  /** Additional content to go in the info drawer below the element. This will be replaced by a tooltip later probably.  */
-  info: PropTypes.string
+  children: function(props, propName, componentName) {
```
byrekt: This prop type checking may be overkill, but since this component does still have a pretty rigid API I thought it worth doing.

src/Input.js
```diff
@@ -22,7 +22,7 @@ const Input = styled.input`
   display: block;
   width: 100%;
   font-family: inherit;
-  font-size: ${theme('fontSizes.2')}px;
+  font-size: ${theme('fontSizes.1')}px;
```
hakimelek: this should probably be replace with `${props => props.theme.fontSizes[1]}`

src/InputField.js
```diff
@@ -9,72 +9,144 @@ import Label from './Label'
 import Input from './Input'
 import theme from './theme'
```
hakimelek: I think that we should not import `theme` in the component unless you would like to use it as a default prop as what @jes708 did on some other components. However, there are sometimes a lot of gotchas in this method if the component has a lot of other nested components that depend on theme during testing. The other alternative is to wrap the test with a `ThemeProvider`. And I see that you are not using styled-component here, so you are good to go for now.

src/InputField.js
```diff
@@ -9,72 +9,144 @@ import Label from './Label'
...
+    if (label) {
+      LabelChild = <Label>{label}</Label>
+    }
+    if (!InputChild) {
+      InputChild = <Input />
+    }
+
+    return (
+      <Box>
+        {LabelChild &&
```
hakimelek: We should probably show the label if `InputChild.props.value` contains a value. So that we hide the label if the input is not filled. It would also be awesome to add that nice `transition: padding .2s ease-in-out;` to slide the text down when the user start typing.

src/InputField.js
```diff
@@ -9,72 +9,144 @@ import Label from './Label'
...
+            <Box mr={-4} ml={8 - iconAdjustment} mt={LabelChild ? -12 : 1}>
+              {BeforeIcon}
+            </Box>
+          )}
+          {React.cloneElement(InputChild, {
+            pt: LabelChild ? 20 : 14,
+            pb: LabelChild ? 8 : 14,
+            mt: LabelChild && -20,
+            pr: AfterIcon && 40,
+            pl: BeforeIcon ? 40 : 2,
```
hakimelek: might be a good idea to just add the translation here as a style prop. I don't see the need for an entire new styled component  ¯\_(ツ)_/¯.

src/Label.js
```diff
@@ -1,24 +1,30 @@
 import styled from 'styled-components'
-import { space, fontSize, color, propTypes } from 'styled-system'
+import { space, fontSize, fontWeight, color, propTypes } from 'styled-system'
 import theme from './theme'
```
hakimelek: Don't see the need for importing theme here as well.

src/Label.js
```diff
@@ -1,24 +1,30 @@
 import styled from 'styled-components'
-import { space, fontSize, color, propTypes } from 'styled-system'
```
hakimelek: Why importing `propTypes` from  `styled-system`? why not `prop-types`?

src/InputField.js
```diff
@@ -9,72 +9,144 @@ import Label from './Label'
...
-  )
+    let InputChild
+    let inputPosition = -1
+    let LabelChild
+    let BeforeIcon
+    let AfterIcon
+    let inputId
+    let iconAdjustment
+
+    React.Children.forEach(children, (child, index) => {
```
hakimelek: Awesome! 👍

src/Input.js
byrekt: Actually @jxnblk advised me to use that `theme` helper for this when I originally worked on the `Input` component.

src/Label.js
byrekt: This `propTypes` is one provided by styled-system that contain the prop-types for several of the styled-system components so you can append them to the component's prop types. See line 14-18

src/Label.js
byrekt: Importing the `theme` here is what @jes708 added so that he could add the theme as a default prop.

byrekt:
Now with even more CSS transitions:
input-field-label

.storybook/InputField.js
```diff
@@ -32,84 +32,89 @@ storiesOf('InputField', module)
...
-              icon="checkLight"
-              color="darkGreen"
-              label="Email Address"
-              defaultValue="oliver.dumoulin@priceline.com"
-              id="form-field-3"
-              placeholder="example@test.com"
-            />
+            <InputField color="green">
+              <Label>Email address</Label>
+              <Input id="form-field-3" defaultValue="olvier@example.com" />
```
hakimelek: Can we add a placeholder for each one of these? It's just have a better experience with a placeholder!

PR #164, opened 2017-11-28
Add numbered colors deprecation warning

hakimelek:
If you would like to use something like `blue0` now. You will see something like this in your console.
image

Also, added some flags in the storybook
image

Thanks @747823 for the help :)

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -62,7 +62,21 @@ const darkPurple = '#407'
 const flatten = (name, colors) =>
   colors.reduce((a, b, i) => {
     const color = {
-      [name + i]: b
+      [name + i]: {
+        enumerable: true,
+        get() {
+          console.warn(
+            `Priceline Design System Warning: Using numbered colors like ${[
+              name + i
+            ]} will be deprecated in the next theme. Use light${name
+              .charAt(0)
+              .toUpperCase() + name.slice(1)}, ${name} or dark${name
+              .charAt(0)
+              .toUpperCase() + name.slice(1)} instead.`
+          )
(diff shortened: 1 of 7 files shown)


hakimelek:
I find out that the console.warn gets thrown on pretty much every test we are running. I will be digging more to how to solve that problem and have that thrown only when a numbered color gets called.

src/theme.js
```diff
@@ -62,7 +62,21 @@ const darkPurple = '#407'
 const flatten = (name, colors) =>
   colors.reduce((a, b, i) => {
     const color = {
-      [name + i]: b
+      [name + i]: {
+        enumerable: true,
+        get() {
```
jxnblk: Personally, I think this would be better handled as a custom prop type on the components that use `color` and `bg` props. The custom propTypes could go into their own module to reuse across multiple components, see this for an example: https://reactjs.org/docs/typechecking-with-proptypes.html#proptypes

hakimelek: What when consumed from a `theme`?

jxnblk: Good point. I think we could almost consider consuming directly from the theme a *private API* and not part of the supported API – hopefully no one's using the old color names from the theme 😬

hakimelek: A lot of people do :/ I have seen that a lot, and that's why a warning would hopefully make people do the change, at least for now, until we deprecate all the numbered colors.

jxnblk: Good to know. I'd say if you see it in the wild, try to leave comments or make changes to use the new color names where possible

Issue #163, opened 2017-11-28
Update Banner as a compound component.


hakimelek:
Use [InputField](https://github.com/pricelinelabs/design-system/blob/master/src/InputField.js) as a reference. Also take a look at #155

PR #166, opened 2017-12-01
#163 add children rendering to banner

krevat:
Wanted to be able to render html / other design system components inside a banner instead of just passing in text. Didn't remove any functionality, only added.

before: `<Banner text='Lorem ipsum' />`
after: `<Banner> **go crazy** </Banner>`

image

### DIFF
--- a/src/Banner.js
+++ b/src/Banner.js
@@ -15,7 +15,7 @@ const bannerColors = {
   },
   lightGreen: {
     backgroundColor: 'lightGreen',
-    color: 'text',
+    color: 'darkGreen',
     icon: 'success'
   },
   red: {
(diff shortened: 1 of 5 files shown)


src/Banner.js
```diff
@@ -62,6 +62,7 @@ const Banner = props => {
         <Box width={1} align={props.textAlign}>
           <Heading.h5>{props.header}</Heading.h5>
           <Text.span fontSize={1}>{props.text}</Text.span>
+          {props.children}
```
jes708: Would the implementation from my PR suit your needs?

https://github.com/pricelinelabs/design-system/pull/145/files#diff-fa3de661379cadd7a682d81cecc3699aR64

krevat: Oh cool didn't see that. I think it's more intuitive to use props.children but I guess that's not up to me.

krevat:
@jes708 does everything look ok? added your commit

jxnblk:
Awesome, looks great! Thanks for putting this together!

PR #167, opened 2017-12-01
Add Tooltip Component

krevat:
New tooltip component. 🎉 
Use for error notifications / general info / general fun tooltip time.

image

e.g.
image


src/Tooltip.js
```diff
@@ -0,0 +1,130 @@
...
+const tooltipAlign = props => {
+  return props.right
+    ? { right: 0 }
+    : props.center
+      ? { left: '50%', width: 'auto', transform: 'translateX(-50%)' }
+      : null
+}
+
+const TooltipContent = styled(Box)`
+  display: ${({ isVisible }) => (isVisible ? 'inline' : 'none')};
```
jxnblk: Just catching up here – what's the rationale for setting `display: none` instead of not rendering or returning false? If it's for CSS transitions, I think it makes sense to handle in CSS, otherwise it's probably better to not render the HTML at all

jxnblk: e.g. something like `{showTooltip && <Tooltip />}` seems like it would keep the API surface area of the design system smaller

src/Tooltip.js
```diff
@@ -0,0 +1,130 @@
...
+  left: PropTypes.bool,
+  right: PropTypes.bool,
+  isVisible: PropTypes.bool
+}
+
+const defaultProps = {
+  position: 'bottom',
+  color: 'text',
+  bg: 'white',
+  align: 'right'
```
jxnblk: We've started adding `theme` as a default prop to help with testing – would be good to pass it to the TooltipContent component IMO

src/Tooltip.js
```diff
@@ -0,0 +1,130 @@
...
+}
+
+class Tooltip extends React.Component {
```
jxnblk: This could be a stateless functional component

src/Tooltip.js
```diff
@@ -0,0 +1,130 @@
...
+  constructor() {
+    super()
+  }
+  render() {
+    const { children, ...props } = this.props
+    return (
+      <div style={{ position: 'relative' }}>
+        <TooltipContent p={2} mb={3} mt={2} {...props}>
```
jxnblk: You can pass `props.children` along with the `{...props}` here for fewer LOC

src/Tooltip.js
krevat: @jxnblk how come it was recently removed from here >>
screen shot 2017-12-07 at 10 38 59 am

hakimelek: @krevat It was remove from icons since we really didn't need theme there. 
@jxnblk I feel like it is a workaround for now but I think we should figure out another way to solve testing components with using the theme. I would still prefer wrapping the component in the test with a `ThemeProvider`.

src/Tooltip.js
```diff
@@ -0,0 +1,124 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Box from './Box'
+import Card from './Card'
+
+import theme from './theme'
```
747823: Looks like you aren't using the `theme` or `Card` imports.

src/Tooltip.js
```diff
@@ -0,0 +1,124 @@
...
+
+import styled from 'styled-components'
+
+const arrowShadow = props => {
+  return props.top
+    ? {
+        'box-shadow':
```
747823: Not sure what lint config this is using but I pasted it into a project that has eslint-config-standard and it's telling me there are 2 extra spaces here.

jxnblk: This project doesn't use linting, it uses [prettier](https://github.com/prettier/prettier) to help avoid commenting on code style :)

src/Tooltip.js
jxnblk: Happy to discuss it again if there's any concerns, but we already discussed passing `theme` as a default prop, and the other components do this as well.
For this PR, I think it would be good to follow conventions with the other components in this library – we also don't define `propTypes` and `defaultProps` at the beginning of the file elsewhere, but that's something we haven't discussed yet.

hakimelek: Yes, I am kinda questioning the convention here haha. I think that adding `theme` as a default prop is a good workaround for simple components that doesn't contain styled-components as a children that consumes the `theme` in their styles. If the component has a lot of nested components the `theme` will not exist vs in `ThemeProvider` it will since it's passed through the context.

krevat:
@747823 @jxnblk i addressed your changes. can you take another gander

PR #169, opened 2017-12-04
Changed lightGray, lightOrange, orange, and darkOrange colors.

odmln:
After reviewing with the design team we thought it would be best to change/update a few of the colors in the design system. The first being lightGray — we wanted something lighter with more of a blue tint. The entire orange palette has also been updated slightly to allow for more legibility.


package.json
```diff
@@ -40,8 +40,8 @@
   ],
   "license": "MIT",
   "devDependencies": {
-    "@storybook/addon-info": "^3.2.0",
-    "@storybook/react": "^3.2.3",
+    "@storybook/addon-info": "^3.2.17",
```
hakimelek: Do we need the package updates? If not, I would prefer having these on a different PR.

PR #171, opened 2017-12-05
Badge color, transform, and spacing fixes

odmln:
Light BG badges now pull in their dark counterparts for text. Orange badge uses dark text for legibility purposes. Also added `text-transform: uppercase` so you no longer have to type in all caps and some `letter-spacing` to make it read better.


src/Badge.js
```diff
@@ -1,61 +1,61 @@
+import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import { space } from 'styled-system'
+import { type, space, color } from 'styled-system'
```
jxnblk: I don't think `type` is part of the styled-system API 🤔

src/Badge.js
```diff
@@ -1,61 +1,61 @@
...
+  letter-spacing: 0.64px;
+  ${type} ${space} ${color};
 `
 
-Badge.defaultProps = {
-  px: 2,
-  py: 1,
-  theme: theme
+const Badge = props => {
+  const badgeColor = badgeColors[props.bg] || {}
```
jxnblk: I like that this API sticks to the current convention, but it is probably worth discussing whether or not we want a color-combination-specific prop to handle both foreground and background colors

src/Badge.js
```diff
@@ -1,61 +1,61 @@
...
+  return <Base {...props} {...badgeColor} />
```
jxnblk: This should probably still be handled with the `type` function passed to styled-components from before. What you have here is more React-like, but less consistent with the rest of the components in the design system

src/Badge.js
odmln: Yeah I don't know why i put that there 🤦

Issue #175, opened 2017-12-11
In <Hug> Component, if there is no icon needed, the <Text.span> wouldn't be 100% width


jinchi2013:
```
   <Flex bg={bg} color={color} p={p} align="center">
      {!!icon && (
        <HideableIcon
          mr={2}
          mt="-2px"
          mb="2px"
          name={icon}
          size={24}
          display={iconDisplay}
        />
      )}
      <Text.span fontSize={fontSize}>{props.text}</Text.span>
    </Flex>
```
If there is no icon, `<Text.span>` wouldn't take 100%.  
In some case, the copy needs to be right aligned without the icon.
I think in this case just `<Text/>` would be better

@jxnblk @jes708

jxnblk:
Yes, generally I think using the non-span version of the Text component should be preferred. Text.span is good for when you need to baseline align inline elements

PR #176, opened 2017-12-11
Hug: use Text component to have 100% width in the hug

jinchi2013:
for issue [#175](https://github.com/pricelinelabs/design-system/issues/175)


jxnblk:
Awesome, thanks!

PR #180, opened 2017-12-12
feat(link-pointer): cursor should be pointer on link

jes708:


src/Link.js
```diff
@@ -4,6 +4,7 @@ import { color } from 'styled-system'
 import theme from './theme'
 
 const Link = styled.a`
+  cursor: pointer;
```
jxnblk: Not opposed to this change, but when a link has an `href` (which it should) the cursor will be a pointer – *not* setting this style can sometimes be a good way to notice accessibility issues

PR #182, opened 2017-12-13
Update icon and button

haasalex16:
IconButton: You could not get new icons as it was not passing 'legacy' to the Icon component

Button: If you passed in any padding it would be defined before the default padding - so would not be applied.  

See below where "px:8px" is passed in but the default padding: '0 18px' is lower on stylesheet so will always take precedent.

screen shot 2017-12-13 at 12 09 05 pm


jxnblk:
LGTM

PR #186, opened 2017-12-15
Fix: add aliases for renamed icons in #153

jxnblk:


src/Icon.js
```diff
@@ -8,7 +8,21 @@ import icons from '../icons.json'
...
+  cruise: icons.cruises,
+  description: icons.document,
+  hotel: icons.hotels,
+  allInclusive: icons.inclusive,
+  radioFilled: icons.radioChecked,
+  radio: icons.radioEmpty,
+  add: icons.radioPlus,
+  minus: icons.radioMinus,
+  businessSeat: icons.seatBusiness,
+  economySeat: icons.seatEconomy
```
hakimelek: `place.svg` is to be added: https://github.com/pricelinelabs/design-system/pull/153/files#diff-894dd3a94686119694d230b58e6b41b9

hakimelek: plane.svg is to be added: https://github.com/pricelinelabs/design-system/pull/153/files#diff-894dd3a94686119694d230b58e6b41b9 

Might need to fallback to `flights`.

PR #188, opened 2017-12-17
Add legacy Icon name deprecation warning

### LINKED ISSUE #187: Add proptype warnings for deprecated icons name
Add warnings for the following aliases:
- scrollLeft
- chevronLight
- chevronThick
- box
- car
- cruise
- description
- hotel
- allInclusive
- radioFilled
- radio
- add
- minus
- businessSeat
- economySeat
- plane

More aliases may be defined here: https://github.com/pricelinelabs/design-system/blob/master/src/Icon.js#L8

hakimelek:
Should fix #187.


src/Icon.js
```diff
@@ -70,7 +70,18 @@ const allKeys = Object.keys({
 })
 
 Icon.propTypes = {
-  name: PropTypes.oneOf(allKeys).isRequired,
+  name: props => {
+    const name = props.name
```
jxnblk: I think the second argument is the `name`, which means you could skip this line, right?

src/Icon.js
```diff
@@ -70,7 +70,18 @@ const allKeys = Object.keys({
...
+    if (aliases[name] && !icons[name] && !icons.legacy[name]) {
+      console.warn(`Using '${name}' as an Icon name has been deprecated.`)
```
jxnblk: It'd be nice to say *use <iconKey> instead* so that people know what the new name is

src/__tests__/Icon.js
```diff
@@ -49,4 +49,14 @@ describe('Icon', () => {
     expect(icon).toBe(null)
     jest.mock
   })
+
+  test('Setting a deprecated icon name should throw a warning', () => {
+    console.warn = jest.genMockFunction()
+    const icon = renderer.create(<Icon name="description" />).toJSON()
+
+    // We expect one propType warning.
+    expect(console.warn.mock.calls.length).toBe(1)
```
jxnblk: 👍

jxnblk:
BTW, using the word `fixes` before the issue number (rather than addresses) will close the issue once this is merged in. See https://help.github.com/articles/closing-issues-using-keywords/

hakimelek:
Made an update to the warning message. We will have to keep the condition for not including legacy and new icon names part of the warning check because some of the names that are in the aliases actually are used in the new icons set, example `minus` and `plus`.

src/__tests__/Icon.js
```diff
@@ -49,4 +49,14 @@ describe('Icon', () => {
...
+    expect(icon).not.toBe(null)
+    jest.mock
```
jxnblk: What does this do?

jxnblk: ✂️

Issue #190, opened 2017-12-18
Remove id attribute from Radio component


jxnblk:
Since HTML documents can only have one `id` per document, the Radio component should not include a hard-coded `id` attribute. See https://github.com/pricelinelabs/design-system/blob/master/src/Radio.js#L23

PR #194, opened 2017-12-18
Remove id attribute from Radio component

sivdahlberg:
Removed id from radio button. Id has to be passed in as prop.

### DIFF
--- a/src/__tests__/Radio.js
+++ b/src/__tests__/Radio.js
@@ -7,7 +7,7 @@ describe('Radio', () => {
     const json = renderer
       .create(
         <ThemeProvider>
-          <Radio checked />
+          <Radio checked id="radio-input" />
         </ThemeProvider>
       )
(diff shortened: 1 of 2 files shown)


src/__tests__/Radio.js
```diff
@@ -7,7 +7,7 @@ describe('Radio', () => {
     const json = renderer
       .create(
         <ThemeProvider>
-          <Radio checked />
+          <Radio checked id="radio-input" />
```
BeniCheni: Trivial, if the PR removes the `id` attribute, can the test code also remove the `id` attribute?

jxnblk: I'm guessing this was added to compensate for the snapshot changes, but I think it's okay to update the snapshots with this PR by running `npm t -- -u`

jxnblk:
Fixes #190

PR #195, opened 2017-12-19
Add Slider component

### LINKED ISSUE #70: Slider component
e.g. range input

jxnblk:
Fixes #70 

screen shot 2017-12-19 at 11 13 38 am

Adds [rc-slider](https://github.com/react-component/slider) as a dependency, which is the same package used in the current pcln-filters components. If we'd prefer to keep dependencies like this out of the design-system, this component can become its own package. I also vetted this library, which works better with CSS-in-JS libraries like styled-components, but seemed less mature: https://github.com/sghall/react-compound-slider

There is a discrepancy in the designs between what's on Zeplin and a screenshot I received from @odmln – we'll need direction on the following visual aspects before merging in:

- size of track, handles, and handle border
- disabled state
- hover state
- focus state
- active state

Zeplin:
screen shot 2017-12-19 at 11 19 06 am

Screenshot:
screen shot 2017-12-12 at 5 22 18 pm


.storybook/Slider.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { withState } from 'recompose'
+import { Slider } from '../src'
+
+const DemoSlider = withState('value', 'onChange', [32, 64])(Slider)
```
jxnblk: I used recompose to allow the sliders to be interactive in Storybook, but can remove this if it's not something we'd want to use in other stories

src/Slider.js
```diff
@@ -0,0 +1,73 @@
...
+  touch-action: none;
+
+  ${space} ${color} & .rc-slider-rail, & .rc-slider-track {
+    height: 8px;
+  }
+  & .rc-slider-handle {
+    width: 32px;
+    height: 32px;
+    margin-left: -16px;
+    margin-top: -12px;
```
jxnblk: I grouped size-related properties together to make this easier to adjust

src/Slider.js
```diff
@@ -0,0 +1,73 @@
...
+    cursor: grab;
+    border-radius: 9999px;
+    border: solid 4px currentcolor;
+    background-color: #fff;
+    touch-action: pan-x;
+
+    &:hover {
+    }
+    &:active {
+    }
```
jxnblk: I left these in as placeholders until we get direction on what the hover, focus, and active states should look like

src/Slider.js
```diff
@@ -0,0 +1,78 @@
...
+
+Slider.defaultProps = {
+  allowCross: false,
+  color: 'blue',
+  theme
+}
+
+Slider.propTypes = {
+  ...propTypes.space,
+  ...propTypes.color
```
jxnblk: Prop types for styled-system functions – all other prop type definitions should be handled by the rc-slider component

src/Slider.js
```diff
@@ -0,0 +1,78 @@
...
+    .rc-slider-handle {
+      box-shadow: none;
+      cursor: default;
+    }
+  }
+`
+
+Slider.defaultProps = {
+  allowCross: false,
+  color: 'blue',
```
broox9: where is this prop used, within the RC slider?

PR #197, opened 2017-12-29
fixed Radio alignment

### LINKED ISSUE #191: Radio component does not need to extend Input since it's hidden
Currently the Radio component extends Input, but does not use any of its styles. The RadioInput component can use the HTML element directly, e.g. `styled.input` - see https://github.com/pricelinelabs/design-system/blob/master/src/Radio.js#L39

tornglintaffychen:
This also resolves #191

### DIFF
--- a/src/Radio.js
+++ b/src/Radio.js
@@ -4,7 +4,6 @@ import PropTypes from 'prop-types'
 import { color } from 'styled-system'
 import Icon from './Icon'
 import Text from './Text'
-import Input from './Input'
 
 const propTypes = {
   checked: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


src/Radio.js
```diff
@@ -37,8 +38,7 @@ const RadioWrap = styled.span`
 `
 
 const RadioInput = styled(Input)`
-  border-style: none;
-  display: inline;
+  display: none;
```
jxnblk: Setting `display: none` on form elements creates accessibility issues – there are techniques for visibly hiding elements in an accessible way, which we should do here. See @broox9's PR for an example: https://github.com/pricelinelabs/design-system/pull/146/files#diff-4fff159aacbfb02718d812c3eddff067R45

msafari: yeah you're right! I think Input already has `appearance: none` so we can just get rid of this

jxnblk:
LGTM :shipit:

PR #199, opened 2018-01-10
fix(Icon): removed space props from svg element

codeflyer:
This PR add a filter on the props passed to the svg element to prevent the Unknown Props React warning: https://reactjs.org/warnings/unknown-prop.html

### DIFF
--- a/src/Icon.js
+++ b/src/Icon.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { space, color } from 'styled-system'
+import { space, color, propTypes, cleanElement } from 'styled-system'
 import PropTypes from 'prop-types'
 import icons from '../icons.json'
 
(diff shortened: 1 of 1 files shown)


src/Icon.js
```diff
@@ -33,20 +33,26 @@ const getPath = ({ name, legacy }) => {
...
+  ...propTypes.space
+}
+
 const Base = ({ name, size, legacy, ...props }) => {
   const icon = getPath({ name, legacy })
   if (!icon) return false
 
   return (
-    <svg
+    <CleanSvg
```
jes708: Is it possible that any apps had margins assigned to their `Icon`'s?  I wouldn't want to break anyone's layout.

codeflyer: The wrapper just remove the mr, mx, etc.. props. that are not used from the svg element. The warning in the console is thrown because there are some app (e.g. search) that add this values.
...but while I'm writing this I've seen that probably is enough to remove the prop from the parent component. I'll check.

src/Icon.js
```diff
@@ -33,20 +33,26 @@ const getPath = ({ name, legacy }) => {
   return icons.legacy[name] || icons[name] || aliases[name]
 }
 
+// Remove `space` props from the `svg` element prevents react warnings
+const CleanSvg = cleanElement('svg')
+CleanSvg.propTypes = {
+  ...propTypes.space
```
jxnblk: Would probably make sense to add propTypes for the `color` function as well

jxnblk:
Thanks! Generally this LGTM, but I think snapshots need to be updated with `npm t -- -u`

msafari:
@hakimelek opened again: https://github.com/pricelinelabs/design-system/pull/232

PR #201, opened 2018-01-11
Stamp component creation

odmln:
I built a new component called `<Stamp>` which is a styled div. It was made for merchandising purposes on listings. Can be used with or without icon. Example below:

`<Stamp color='blue'>
     <Icon name='pin' size={16} /> top location
</Stamp>`

screen shot 2018-01-11 at 5 13 46 pm


hakimelek:
Looks good, can you consider adding a section in the docs about the `Tag` component.

docs/Stamp.md
```diff
@@ -0,0 +1,25 @@
+# Stamp
+
+Use the `<Stamp> </Stamp>` component to render a primitive text Stamp.
```
jxnblk: Might be good to clarify where this component could be used, or what the rationale for it is

src/Stamp.js
```diff
@@ -0,0 +1,75 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { space } from 'styled-system'
+import theme from './theme'
+
+const type = props => {
```
jxnblk: NB: I'd suggest naming this something more explicit like `getColorCombination` – these sorts of color combinations and the ones in Badge, Button, etc. would be good to abstract out to the theme at some point

src/Stamp.js
```diff
@@ -0,0 +1,75 @@
...
+  }
+  return StampColors[props.color]
+}
+
+const Stamp = styled.div`
+  display: inline-flex;
+  align-items: center;
+  vertical-align: top;
+  min-height: 24px;
+  font-size: ${props => props.theme.fontSizes[0]}px;
```
jxnblk: NB: I'd say it might be better to handle font-size with the `fontSize` utility and a default prop for code consistency with the other components

odmln:
@jxnblk I added better docs, removed type and set fontSize as a default prop.

src/Stamp.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { space, fontSize, color } from 'styled-system'
+import theme from './theme'
+
+const getColorCombination = props => {
```
msafari: Is this unused?

src/Stamp.js
```diff
@@ -0,0 +1,76 @@
...
+  min-height: 24px;
+  font-weight: 600;
+  text-transform: uppercase;
+  letter-spacing: 0.05em;
+  border-radius: 2px;
+  border-width: 1px;
+  border-style: solid;
+  border-color: ${props =>
+    props.disabled ? null : props.theme.colors.borderGray};
+  ${space} ${fontSize} ${color};
```
msafari: if you want your colorCombination to be actually applied. you can replace `${color}` with `${getColorCombination}`

src/Stamp.js
odmln: @msafari good call, I just removed `getColorCombination`.

PR #210, opened 2018-01-24
bugfix(InputField): add default margins to InputField

byrekt:
I noticed when trying to use an InputField that global `input` styles for margin in a host app were causing the margins to be incorrect. 
margin-issue

So modified the code so that if "showLabel" evaluates to false I set some default margins instead of only applying a margin when it's true.

### DIFF
--- a/src/InputField.js
+++ b/src/InputField.js
@@ -114,7 +114,7 @@ class InputField extends React.Component {
         {showLabel &&
           React.cloneElement(LabelChild, {
             pl: BeforeIcon ? 40 : 2,
-            mt: '6px',
+            mt: '5px',
             style: labelStyles,
             htmlFor: inputId
           })}
--- a/src/Input.js
+++ b/src/Input.js
@@ -34,6 +34,8 @@ const Input = styled.input`
   padding-left: 12px;
   padding-right: 12px;
 
+  margin: 0;
+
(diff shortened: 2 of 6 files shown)


jxnblk:
:shipit: LGTM

src/InputField.js
```diff
@@ -127,9 +127,9 @@ class InputField extends React.Component {
           {React.cloneElement(InputChild, {
             'aria-label':
               !showLabel && inputPlaceholder ? inputPlaceholder : null,
-            mt: showLabel && -20,
+            mt: showLabel ? -20 : 0,
```
jxnblk: Just noticed that the Input component itself doesn't reset its margin to `0` – is it worth making an update to that component instead?

byrekt: oh yeah that might be a better solution. I'll post an update tomorrow morning

src/InputField.js
```diff
@@ -114,7 +114,7 @@ class InputField extends React.Component {
         {showLabel &&
           React.cloneElement(LabelChild, {
             pl: BeforeIcon ? 40 : 2,
-            mt: '6px',
+            mt: '5px',
```
byrekt: Someone's update to border width means this value had to be modified.

PR #211, opened 2018-01-26
Docs website

jxnblk:
To help make it easier to get started and use the design system, this adds a documentation site setup to the `docs/` folder.

screen shot 2018-01-24 at 3 07 28 pm

Features:

- :atom_symbol: Static site generated with React
- :nail_care: Uses styled-components SSR features for inlined CSS
- `M↓` Reads data from our existing markdown docs files
- :information_desk_person: Uses the design system components to render markdown as HTML using [remark-react][remark-react]
- :computer:  Includes live code examples inline using [react-live][react-live]
- :gem: Includes download link to sketch assets
- :rainbow: Click to copy color values
- :v: Click to download icon SVGs
- :sunglasses: Thanks for @odmln for help with the design

## Storybook

We'll continue to use Storybook for development, and the published version will
likely be moved to https://pricelinelabs.github.io/design-system/storybook in a separate PR.
This PR does not handle automating the build of the documentation site.

## Build setup

The `docs/` folder now has it's own `package.json` file to manage dependencies and scripts for building the docs site.
It depends on `pcln-design-system`, and in order to make changes in both the main package and the docs at the same time, you can use [npm link](https://docs.npmjs.com/cli/link) for local development.

The docs site uses the [x0][x0] library to build a static site with a client-side bundle.
This library makes use of `async` `await`, so you'll need Node.js 7.6 or above in order to build the site.
Under the hood, x0 is a basic webpack config and dev server that takes a single React component as input.
This should be an easy library to eject from in the future, if we decide not to use it, since it's completely React-based.

The only custom API that x0 makes use of is the [`getInitialProps`](https://github.com/c8r/x0#fetching-data) static method.
This allows for async and server-side data (in our case, filesystem reading of the markdown files) to be passed in as props to the component.
If we decide to use another setup for static site generation, this can be handled with a separate Node build script.

Caveats: the main client side bundle for the docs site is currently quite big,
partially due to the use of [remark-react][remark-react].
I hope we can add some bundle size and performance improvements later.

**Running in development mode**

```sh
cd docs
npm install
npm start
```

**Building the static site locally**

```sh
npm run dev-build
```

This will generate the site in the `docs/site/` folder, which can be served with a local file server like http-server or `python -m SimpleHTTPServer`

### Other options that were considered for static site generation

- **[Gatsby][gatsby]** Gatsby is great for building performant, static sites with React. The one limitation I ran into with this was not easily being able to add live code examples that are generated from markdown code blocks. I would guess that this could be possible by building a plugin for Gatsby, but seemed like a lot of extra work at the time.
- **[Next.js][nextjs]** Although Next.js is intended for server-side React applications, it can also export static sites.
Next.js works well with a handful of defined routes, but generating routes based on the markdown files in our docs folder felt like more work than it would be worth.
- **[After.js][afterjs]** This library is also intended for server-side applications and is newer, so I haven't had the chance to vet it fully yet, but it seems promising.
- **[Docusaurus][docusaurus]** This came out recently, and I haven't had the chance to vet it yet, but could be an alternative in the future.
- **[react-styleguidist][react-styleguidist]** This is more inline with being a Storybook replacement, but uses markdown for documentation instead of a custom API like Storybook uses.

**Note**: any changes outside of the `docs/` folder are due to prettier formatting

[x0]: https://github.com/c8r/x0
[remark-react]: https://github.com/mapbox/remark-react
[react-live]: https://github.com/FormidableLabs/react-live
[gatsby]: https://github.com/gatsbyjs/gatsby
[nextjs]: https://github.com/zeit/next.js
[afterjs]: https://github.com/jaredpalmer/after.js
[docusaurus]: https://docusaurus.io
[react-styleguidist]: https://github.com/styleguidist/react-styleguidist


hakimelek:
npm install and npm start I am getting the following error: 
```
 x0  @compositor/x0

⠋ /design-system/docs/node_modules/@compositor/x0/lib/static/client.js:66
  )
  ^
SyntaxError: Unexpected token )
    at createScript (vm.js:53:10)
    at Object.runInThisContext (vm.js:95:10)
    at Module._compile (module.js:543:28)
    at Module._extensions..js (module.js:580:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/mhakim13/Dropbox/Coding_Projects/design-system/docs/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
    at require (internal/module.js:20:19)
```

When I removed
```js
  config.resolve.modules.unshift(
     dirname,
     path.join(process.cwd(), 'node_modules'),
     path.join(dirname, 'node_modules'),
   )
```
inside @composer/x0/lib/dev/index.js, it seems to work without any problem.

jxnblk:
@hakimelek thanks! That looks like it's a trailing comma in a function call issue. Should remove that for better node compatibility

docs/Contributing.md
```diff
@@ -89,15 +89,15 @@ Generally the workflow looks like this:
 10. Push the git tags created with the npm CLI to GitHub with `git push --tags`
 
 
-## Beta Version
+### Beta Version
 
 As of the date this was written, the Design System is currently in a beta.
 The package versioning reflects this with the version `1.0.0-x` convention.
 Once the library is in a more stable and mature state, we will release a stable `1.0.0`
 and strictly follow sematic versioning from that point on.
```
msafari: semantic *

PR #212, opened 2018-01-26
(WIP) Animated loader

odmln:
Working on building an animated loading component. Right now theres only one custom animation, `bounce`, but once we dig more into motion I imagine we can add more options. Color is the only prop at the moment. Open to thoughts & feedback, specific use cases, etc.

Examples of Loader by itself and contained inside of a button below
bounce


jxnblk:
@odmln @sqott does this PR need some more work/discussion? Would like to close this out since there doesn't seem to be much activity on this

odmln:
@jxnblk i think this can be closed out, can be revisited when we further explore motion in the DS.

PR #214, opened 2018-01-29
bug(InputField): remove defaultValue as possible prop to Input since …

byrekt:
…it's a controlled component

When I wrote the component initially I intended for it to be a controlled component but based whether the label showed up on whether or not a `defaultValue` prop was passed to the `<Input />` component nested inside. As a result, if you wanted the `Label` to show up when some initial value was in the `<Input />` component you had to pass a default value, which was causing an `uncontrolled/controlled` component warning. 

For this PR I'm no longer passing a `defaultValue` prop to the `<Input />` and instead basing whether or not the label shows up on whether or not the `value` prop of the `<Input />` has a value when the `<InputField />` constructor is called.

### DIFF
--- a/src/InputField.js
+++ b/src/InputField.js
@@ -40,16 +40,16 @@ class InputField extends React.Component {
   constructor(props) {
     super(props)
 
-    let hasDefaultValue
+    let hasInitialValue
 
     React.Children.forEach(props.children, child => {
       if (child && child.type === Input) {
-        hasDefaultValue = child.props.defaultValue
+        hasInitialValue = !!child.props.value
       }
     })
 
     this.state = {
-      showLabel: hasDefaultValue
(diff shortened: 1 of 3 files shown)


src/InputField.js
```diff
@@ -40,16 +40,16 @@ class InputField extends React.Component {
   constructor(props) {
     super(props)
 
-    let hasDefaultValue
+    let hasInitialValue
 
     React.Children.forEach(props.children, child => {
       if (child && child.type === Input) {
-        hasDefaultValue = child.props.defaultValue
+        hasInitialValue = child.props.value
```
jxnblk: Should this be `!!child.props.value` to set the state as a boolean?

byrekt: yeah definitely. I've updated it.

Issue #208, opened 2018-01-18
Tooltip z-index prop


msafari:
I think Tooltip should allow z-index as a prop. Tooltips usually are all the way on top of the rest of UI. It'd be nice if host app can set the z-index without having to do this:
```
const StyledTooltip = styled(Tooltip)`
  z-index: 10;
`
```

msafari:
Resolved in https://github.com/pricelinelabs/design-system/pull/215

PR #215, opened 2018-01-30
tooltip - add zIndex prop with default of auto

ronhenderson:


msafari:
I think you need to run `npm run icons` and then `npm test` to update snaps. That may be why travis is failing

src/Tooltip.js
```diff
@@ -96,20 +96,22 @@ const propTypes = {
   top: PropTypes.bool,
   center: PropTypes.bool,
   left: PropTypes.bool,
-  right: PropTypes.bool
+  right: PropTypes.bool,
+  zIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.string])
```
msafari: 👍

jxnblk:
Fixes #208

src/Tooltip.js
```diff
@@ -96,20 +96,22 @@ const propTypes = {
...
   align: 'right',
-  theme: theme
+  theme: theme,
+  zIndex: 'auto'
 }
 
 const Tooltip = ({ children, ...props }) => {
   return (
-    <div style={{ position: 'relative' }}>
+    <div style={{ position: 'relative', zIndex: props.zIndex }}>
```
jxnblk: I know this element was already using inline styles, but I think it would be good to use a styled-component here. If you want to make that change, that'd be cool, otherwise we can follow up in a separate PR

jxnblk: If this were a styled-component, we could easily avoid passing a value for zIndex when it's undefined, e.g. `${props => typeof props.zIndex === 'number' ? { zIndex: props.zIndex } : null}`

hakimelek: @ronhenderson Any updates on this? We can go ahead and merge it and follow up with another PR to move the styles to styled-components.

PR #216, opened 2018-01-30
Add InputGroup component

### LINKED ISSUE #209: Date range input
For the calendar component

jxnblk:
This can be used to visually group multiple inputs together, such as with date range inputs.

screen shot 2018-01-30 at 2 52 33 pm

Fixes #209


src/InputGroup.js
```diff
@@ -0,0 +1,37 @@
...
+
+  & ${Input} {
+    border: 0;
+    box-shadow: none;
+  }
+`
+
+InputGroup.propTypes = {
+  ...propTypes.space,
+  ...propTypes.borderColor
```
BeniCheni: +1 to share an implementation of using `propTypes` from `styled-system` for sharable CSS-driven propType definition such as `borderColor`.

src/__tests__/InputGroup.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import { InputGroup } from '..'
+
+describe('InputGroup', () => {
+  test('renders', () => {
+    const json = renderer.create(<InputGroup />).toJSON()
+    expect(json).toMatchSnapshot()
+  })
```
BeniCheni: Would you think it'd be a valuable test case to test out the usage of the `borderColor` prop support in an additional unit test code?

src/InputGroup.js
```diff
@@ -0,0 +1,37 @@
...
+  border-style: solid;
+  ${borderColor}
+  ${space}
+
+  & > ${Box} {
+    width: 100%;
+    flex: 1 1 auto;
+  }
+
+  & ${Input} {
```
BeniCheni: +1 on this "advanced" example of style selector syntax in `styled-components` with a design system component.

src/InputGroup.js
```diff
@@ -0,0 +1,37 @@
...
+  align-items: center;
+  border-radius: ${system.theme('radius')};
+  border-width: 1px;
+  border-style: solid;
+  ${borderColor}
+  ${space}
+
+  & > ${Box} {
+    width: 100%;
+    flex: 1 1 auto;
```
BeniCheni: @jxnblk ,
This `flex` seems to be a short-hand property for `flex-grow`, `flex-shrink`, and `flex-basis`. (https://developer.mozilla.org/en-US/docs/Web/CSS/flex)

Would you think it'd be too verbose if the code spells all off the implicit rules to avoid using short-hand property?

Also `flex` is only supported >= ie11. Would you have concern about ie10 support?
flex-support

src/__tests__/InputGroup.js
jxnblk: The `borderColor` function is 100% covered in the styled-system library, so we shouldn't need to test it here IMO

src/InputGroup.js
jxnblk: Normally, I'd say avoid nested selectors like this, but I think the InputGroup is an exception to that rule

src/InputGroup.js
jxnblk: I'll change to not use the shorthand. I *think* (but might be wrong) that the `width` property will help this work as expected in IE10

chrisdroukas: I thought we had dropped support for <IE11.

BeniCheni: @chrisdroukas , think IE 10 may be still in "business". We can confirm with the "committee" that authors the browser support.

src/InputGroup.js
hakimelek: I like this!

PR #218, opened 2018-02-01
fix(InputField): Make positioning of Label account for border around …

byrekt:
…the InputField

Not sure how this wasn't noticed before, but since input field has a border, the content of it is shifted over 1 px to the right. The Label has no border and the padding did not account for the border, so it looks too far left. This adds a transparent left border to the Label.

The alternative is to add a pixel of padding-left to the Label.

Before:
image

After:
image

### DIFF
--- a/src/InputField.js
+++ b/src/InputField.js
@@ -19,7 +19,8 @@ const fadeIn = keyframes`
   }
 `
 const labelStyles = {
-  animation: fadeIn + ' 0.3s'
+  animation: fadeIn + ' 0.3s',
+  border: '1px solid transparent'
 }
 
 const getInputStyles = showLabel => {


jxnblk:
This seems fine, but I think it'd be more explicit to add an extra 1px padding instead of an invisible border. It also looks like the border is added on all sides which would change the vertical alignment, wouldn't it?

odmln:
@byrekt @jxnblk This might be a dumb question (also might be a separate issue) but what if we do a box-shadow instead of border on the input? It wouldn't affect the input width/height and would give it a smoother border-radius. 

Default: `box-shadow: 0 0 0 1px ${theme('colors.borderGray')}`
Focus: `box-shadow: 0 0 0 2px ${theme('colors.blue')}`

jxnblk:
A box-shadow also works, which also supports `inset` positioning if that's something we'd want

msafari:
@byrekt @jxnblk I agree! using box-shadow seems like a better solution here

PR #220, opened 2018-02-05
Add optional Label styles

jxnblk:
- `white-space: nowrap` can be useful when used for radios and checkboxes
- The hidden styles aren't great accessibility wise, but better to have a label element than not at all

### DIFF
--- a/.storybook/Label.js
+++ b/.storybook/Label.js
@@ -50,17 +50,17 @@ storiesOf('Label', module)
   ))
   .add('nowrap', () => (
     <Flex>
-      <Label.nowrap>
+      <Label nowrap>
         <Radio checked />
         Round-trip
-      </Label.nowrap>
-      <Label.nowrap>
+      </Label>
+      <Label nowrap>
         <Radio checked />
         One-way
-      </Label.nowrap>
-      <Label.nowrap>
+      </Label>
+      <Label nowrap>
         <Radio checked />
         Multi-destination
-      </Label.nowrap>
+      </Label>
     </Flex>
   ))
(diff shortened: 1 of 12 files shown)


.storybook/Label.js
```diff
@@ -48,3 +48,19 @@ storiesOf('Label', module)
...
+      <Label.nowrap>
+        <Radio checked />
+        One-way
+      </Label.nowrap>
+      <Label.nowrap>
+        <Radio checked />
+        Multi-destination
+      </Label.nowrap>
+    </Flex>
+  ))
```
hakimelek: What about a story for `<Label.hidden>`?

broox9:
what is the consequence of a hidden label for accessibility?

jxnblk:
Hiding labels is generally not a good idea. This is a pretty good writeup: https://a11yproject.com/posts/placeholder-input-elements/ – and somewhat related https://a11yproject.com/posts/how-to-hide-content/

.storybook/Label.js
```diff
@@ -48,3 +48,19 @@ storiesOf('Label', module)
       <Input id="sample-input" />
     </div>
   ))
+  .add('nowrap', () => (
+    <Flex>
+      <Label.nowrap>
```
hakimelek: Actually, a second thought about this, why not make the `Label` take a boolean `nowrap` prop. I think it's much cleaner.

jxnblk: This sort of comes down to a static vs dynamic thing. My initial thinking was that `nowrap` and `hidden` wouldn't be a style that would change dynamically, but we don't really follow this pattern for styles, so I think you're probably right in that this would be more consistent if it were a prop

PR #221, opened 2018-02-05
Add Select form component

### LINKED ISSUE #71: Select component

jxnblk:
Fixes #71 

screen shot 2018-02-05 at 4 34 33 pm
screen shot 2018-02-05 at 4 34 42 pm


.storybook/Select.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Select, Label, Box } from '../src'
+
+storiesOf('Select', module)
+  .add('default', () => (
+    <Box>
+      <Label htmlFor="cabinClass">Cabin Class</Label>
+      <Select id="cabinClass" name="cabinClass" defaultValue="Premium Economy">
```
BeniCheni: For my education, where is the support code for the `defaultValue` attribute? I read up on the MDN `<select>` [doc](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) thinking that it's a native html attribute, but it doesn't seem to be referenced. Though my local testing via story confirms that `defaultValue` attribute works as expected. +1

src/Select.js
```diff
@@ -0,0 +1,42 @@
...
+import styled from 'styled-components'
+import system, { space, fontSize } from 'styled-system'
+
+import theme from './theme'
+import Flex from './Flex'
+import Icon from './Icon'
+
+const SelectBase = styled.select`
+  appearance: none;
+  display: block;
```
BeniCheni: What does this block display try to ensure for the `select` style? When I toggled off the block display rule off in the story's inspector, the UI still looks okay. Trying to learn if this block display rule is trying to address certain use case.

.storybook/Select.js
jxnblk: The `defaultValue` is part of React's form API – it allows the select to change in the storybook without the need for state management. We probably won't be using `defaultValue` in our apps or other components, since we use controlled components with the `value` and `onChange` props.

https://reactjs.org/docs/uncontrolled-components.html

src/Select.js
jxnblk: I think the `<select>` element defaults to `display: inline-block` – this is just to ensure that it's `display: block` though the `width: 100%` will probably make it behave similarly in the layout

PR #223, opened 2018-02-14
WIP: Add responsive Popover layout component

jxnblk:
This could be used for date pickers, traveler selection, autocomplete inputs, etc. cc @chrisdroukas @odmln @BeniCheni @jes708 @hakimelek


.storybook/Popover.js
```diff
@@ -0,0 +1,175 @@
...
+      open={props.open}
+      onDismiss={e => {
+        props.update(toggle('open'))
+      }}
+    >
+      <Flex p={props.open ? [2, 2, 0] : 0} align="center">
+        <Box width={1}>
+          <InputField
+            onChange={e => {
+              // why?
```
hakimelek: why?

src/Popover.js
```diff
@@ -0,0 +1,84 @@
...
+  ${system.theme('mediaQueries.1')} {
+    position: absolute;
+    left: 0;
+    box-shadow: ${system.theme('boxShadows.2')};
+  }
+`
+
+Popover.Body.defaultProps = {
+  p: 3,
+  mt: 2,
```
hakimelek: margin-top should probably be 0 on small screen sizes.
image

src/__tests__/__snapshots__/CloseButton.js.snap
```diff
@@ -12,6 +12,7 @@ exports[`CloseButton renders without props 1`] = `
   display: inline-block;
   vertical-align: middle;
   text-align: center;
+  -webkit-text-decoration: none;
```
jxnblk: Stylis added a bunch of vendor prefixes, so be sure to `npm install` to get the latest for snapshot tests

https://github.com/thysultan/stylis.js/issues/95

src/Popover.js
```diff
@@ -0,0 +1,84 @@
...
+  theme
+}
+
+class Popover extends React.Component {
+  static propTypes = {
+    open: PropTypes.bool.isRequired,
+    onDismiss: PropTypes.func.isRequired
+  }
+
+  constructor() {
```
msafari: Do we really need the constructor? Seems like we can just have handleDismiss as a class function?

jxnblk: I prefer custom class methods in the constructor to avoid the `.bind` syntax, but I don't mind either way

.storybook/Popover.js
```diff
@@ -0,0 +1,175 @@
...
+  Box,
+  Heading,
+  Text,
+  Hide
+} from '../src'
+
+const withState = C =>
+  class extends React.Component {
+    constructor(props) {
+      super()
```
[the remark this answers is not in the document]
jxnblk: It's usually not necessary, but it could be good for consistency in code style – see https://stackoverflow.com/questions/30571875/whats-the-difference-between-super-and-superprops-in-react-when-using-e

jxnblk:
Think this needs a little more work to ensure it will work for the Calendar, TravelerSelection, and Typeahead components

jxnblk:
Closing this one out in favor of a newer approach based on some of the recent work for the calendar and typeahead

PR #225, opened 2018-02-14
feat(InputField): New optional prop to always show Label regardless s…

byrekt:
…tate

New use case for our SEM landing page where they want the label to show up regardless of whether or not the user has begun typing. Previously we relied on the placeholder to serve as the label. 

The following image illustrates it:
image

### DIFF
--- a/.storybook/InputField.js
+++ b/.storybook/InputField.js
@@ -110,6 +110,26 @@ storiesOf('InputField', module)
             </InputField>
           </Box>
         </Flex>
+        <Flex width={900}>
+          <Box width={1 / 2} pl={3}>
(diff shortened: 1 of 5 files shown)


jxnblk:
LGTM

Just a thought about the API around this: it might be nice for this to be a prop on the Label component itself, e.g.

```jsx
<InputField>
  <Label autohide>This one shows/hides</Label>
  <Input />
</InputField>

<InputField>
  <Label>This one is always visible</Label>
  <Input />
</InputField>

```

byrekt:
I really like @jxnblk 's suggestion. I'm going to open a task for that.

PR #234, opened 2018-02-24
Adds Chromatic regression tests

zol:
Hi,

I’m one of the founders at [Chromatic](http://chromaticqa.com), a visual regression testing tool. We’ve been using Priceline's design system as a test case for Chromatic itself (we really like your design aesthetic too).  

As a token of our appreciation, I’d like to offer the design system team Chromatic for free. We’ll also give you hands on support to make sure you have a smooth ride and can get the most out of the tool without taking up much of your time. I've made a small PR that adds visual regression tests. 

We showed Chromatic to @hakimelek late last year and his response was that it looked promising but we dropped the ball on following up. We've made a bunch of improvements since then too.

Chromatic will quickly and easily let you review any visual changes introduced by new code -- it’s especially handy for open source projects where it can be time consuming to evaluate PRs being opened out of the blue by unknown authors.

I’ve already run Chromatic on your design system to establish a baseline for future tests (you can view the results [here](https://www.chromaticqa.com/builds?appId=5a85e56213417800200978ab)). You can run additional tests anytime by calling `npm chromatic`. You could try running `npm chromatic` in this branch after changing a component to see an example of the review flow.

Chromatic really shines when you run the above command from within your CI system as it will then tag PRs with the result of the regression tests, like in the image below.

screen shot 2018-02-23 at 4 42 52 pm

Let me know if you have any questions, once this PR is merged I can help get chromatic running in your CI builds too if you'd like.

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import 'react-chromatic/storybook-addon'
 import { configure, addDecorator } from '@storybook/react'
 import { injectGlobal } from 'styled-components'
 import { ThemeProvider, Box } from '../src'
@@ -15,9 +16,7 @@ injectGlobal([], {
 
 addDecorator(story => (
   <ThemeProvider>
-    <Box p={3}>
-      {story()}
-    </Box>
+    <Box p={3}>{story()}</Box>
   </ThemeProvider>
 ))
 
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "coverage": "jest --coverage",
     "codecov": "jest && codecov",
     "standard": "standard",
-    "test": "jest"
+    "test": "jest",
+    "chromatic": "chromatic test --storybook-addon  --script-name start --app-code=\"uf9gdg938po\""
   },
   "author": "Brent Jackson",
   "contributors": [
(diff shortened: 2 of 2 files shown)


hakimelek:
Hey @zol, 

Thanks for the PR!

Chromatic is still an ongoing discussion in the company. I have let most of the designers who are involved in building UI components know about it and the feedback I have been getting was generally positive with some concerns about the extra step that will be introduced in the process of building, updating the components. I am still thinking that we should keep try using it with the design system and see how it goes.

I have opened another PR with the Travis integration, I have tested it and it seems to work. Let me know if I am missing anything.

Meanwhile @odmln, @chrisdroukas and @jxnblk please take a look at the builds inside Chromatic whenever a new PR is created to help with the review process. We can keep sharing feedback as we use it.

zol:
@hakimelek Sweet! The PR looks great. I understand your team's concern re: adding an extra step. We've found the trade-off in time is worth it to add confidence, I'm curious to see whether you end up finding that or not. I'll mention there is also a command line flag `--exit-zero-on-changes` that you can include in Travis as a way to not fail the build in case changes are detected -- this would make the review step optional.

hakimelek:
Closing this in favor of #236.

PR #235, opened 2018-02-25
feat(tooltip) Add optional prop of width to allow for sizing the 'content' element

LKLopez:
When tooltip is used with large messages there are no developer assignable media breakpoints allowable. These changes allow the developer to implement breakpoints.


hakimelek:
Since the `Tooltip` extends `Box`,  you are able to change the width.  I think we could maybe just mention that in the docs. No extra steps needs.

jxnblk:
Closing this out since it seems like this functionality is already available

PR #236, opened 2018-02-26
Adding chromatic in Travis

hakimelek:
Similar to #234 with Travis integration. 

Adding CHROMATIC_APPCODE as an env variable to travis.


zol:
You may need to double check the `CHROMATIC_APPCODE` variable is exposed for public PRs. Not sure about Travis in particular but some CI systems won't pass env vars for builds started by non-trusted users for security reasons. If this is the case, you can safely commit the `appCode` into your code, it's ok for it to be publicly viewable!

hakimelek:
Thanks for the quick review! I will add the `--exit-zero-on-changes`. I am assuming that the non-trusted users issue is not a problem for Travis, since we have other env variables already setup and never had any problems with the build when other users than owners for example push branches.

package.json
```diff
@@ -78,6 +80,7 @@
     "setupTestFrameworkScriptFile": "./test-setup.js"
   },
   "dependencies": {
+    "babel-runtime": "^6.26.0",
```
jxnblk: You didn't mean to add this dependency, did you?

zol: It's not required for Chromatic.

hakimelek: Nope, good catch, that was part of the old branch.

zol:
👍

PR #246, opened 2018-02-28
Remove legacy colors with warnings

### LINKED ISSUE #239: v2: Remove numbered color names and related warnings

hakimelek:
See ya numbered colors 👋!

Should fix #239.


msafari:
@hakimelek our team still hasn't moved over to new colors. Removing the legacy colors before everyone moves over will break our A/B testing culture :D Let's not merge this in until we make sure at least internally all our teams have moved over to new colors

hakimelek:
@msafari This is part of 2.0.0 which will introduce breaking changes when you will upgrade to it. You can still use v1 and when you are ready to move fully to the new theme, upgrade to 2.0.0. Npm will not bump to v2 unless you explicitly do it.

PR #247, opened 2018-02-28
Remove legacy theme

### LINKED ISSUE #238: v2: Remove legacy theme and related props

hakimelek:
See ya legacy theme 👋

should fix #238.


src/ThemeProvider.js
```diff
@@ -20,11 +19,10 @@ export const Base = styled.div`
   }
 `
 
-const ThemeProvider = ({ legacy, customBreakpoints, ...props }) => {
-  const baseTheme = legacy ? legacyTheme : nextTheme
-  const breakpoints = customBreakpoints || baseTheme.breakpoints
+const ThemeProvider = ({ customBreakpoints, ...props }) => {
+  const breakpoints = customBreakpoints || nextTheme.breakpoints
```
jxnblk: Should we just rename this to `theme`?

hakimelek: We are using theme as a const in the next line. I didn't know what to change its name to (maybe themeWithBreakpoints haha) so I kept the first imported theme as nextTheme.

msafari:
@hakimelek our team still hasn't moved over to new colors. Removing the legacy colors before everyone moves over will break our A/B testing culture :D Let's not merge this in until we make sure at least internally all our teams have moved over to new colors

PR #251, opened 2018-03-01
 Add build step for docs deployment

jxnblk:
Updates the travis config to deploy the docs site on merges into master


docs/README.md
```diff
@@ -30,16 +30,21 @@ To build the docs site for production, run:
...
 ```sh
 npm run dev-build
 ```
 
-Then run a local server in the `docs/` folder.
-The site should be available at `http://localhost:8000/site`, where `8000` is the port for your local server.
+Then run a local server in the root folder.
+
+```sh
+python -m SimpleHTTPServer
```
hakimelek: python?

jxnblk: Probably the easiest way to run a local server on a Mac since it's built in. This is just an example, and any server would work

PR #252, opened 2018-03-01
v2: Update styled-system dependency

### LINKED ISSUE #243: v2: Update to styled-system v2

jxnblk:
Fixes #243

See the [release notes](https://github.com/jxnblk/styled-system/releases/tag/v2.0.0) for styled-system


package.json
```diff
@@ -78,9 +78,10 @@
     "setupTestFrameworkScriptFile": "./test-setup.js"
   },
   "dependencies": {
+    "clean-tag": "^1.0.4",
     "prop-types": "^15.5.10",
-    "styled-components": "^2.1.1",
-    "styled-system": "^1.0.8-0"
+    "styled-components": ">=2.1.1 || >=3.0.0",
```
BeniCheni: Interesting. Did not know about the `||`, and this PR inspired the curiosity to poke around on the the [**Ranges**](https://docs.npmjs.com/misc/semver#ranges) topic in the npm doc about the `||` used in dependency's version. 🙏 for this cool PR with `styled-components` update. 👍 
> A range is composed of one or more comparator sets, joined by ||. A version matches a range if and only if every comparator in at least one of the ||-separated comparator sets is satisfied by the version.

src/theme.js
```diff
@@ -10,7 +10,7 @@ const addAliases = (arr, aliases) =>
     })
   )
 
-export const breakpoints = [32, 40, 48, 64]
+export const breakpoints = [32, 40, 48, 64].map(n => n + 'em')
```
BeniCheni: Curious about a use case of an array with value of `0`. Would CSS unit `0em` be okay? (Don't think it would break any UI, but just checking on the thought of CSS rule with 0 em unit.
```
[0, 0, 0, 0].map(n => `${n}em`)
```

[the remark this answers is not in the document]
jxnblk: I could go either way on this, planning on merging this but can address this in the main v2 PR

Issue #242, opened 2018-02-27
v2: loosen dependency versions and move styled-components to peerDependencies


jxnblk:
Per styled-components' recommendation: https://www.styled-components.com/docs/faqs#i-am-a-library-author-should-i-bundle-styledcomponents-with-my-library

PR #255, opened 2018-03-02
loosen styled-components version

psierks:
To be able to use the styled-components 3.x and benefit from the performance changes in that project.


hakimelek:
@psierks  Thanks for the PR. Loosing the styled-components version is on the roadmap for 2.0 as well (#242).

PR #257, opened 2018-03-07
feat(add-banner-color): allow custom colors

jes708:
Lift the limitation to only allow preset banner colors.  Allow custom background and text colors.

screen shot 2018-03-07 at 1 40 10 pm


jes708:
Note: I updated this PR to make the inner margins (between text and icons) 8px, as a request from design.

.storybook/Banner.js
```diff
@@ -85,6 +85,16 @@ storiesOf('Banner', module)
         bg="lightRed"
         onClose={action('closed')}
       />
+      <Banner
+        textAlign="left"
```
haasalex16: This is causing "Warning: Failed prop type: The Box `align` prop will deprecated. Please use Text instead"  Think it could be cleaned up in the PR as well?

jes708: Good idea.  I've removed it by changing `Box` to `Text`.

PR #258, opened 2018-03-08
Docs: Use schemeless URL for webfont

jxnblk:


docs/src/App.js
```diff
@@ -75,7 +75,7 @@ class App extends React.Component {
           <link
             key="webfont"
             rel="stylesheet"
-            href="http://fonts.googleapis.com/css?family=Montserrat:400,600|Roboto+Mono"
+            href="//fonts.googleapis.com/css?family=Montserrat:400,600|Roboto+Mono"
```
odmln: @jxnblk Do we want to change the font weights to 500 & 700 while we're at it?

jxnblk: I think we'll want to wait until v2 is ready since the font weights in the theme are still 400 and 600, right?

jxnblk:
Merging this since it's a quick fix for the docs

docs/src/App.js
odmln: True, good call. We can hold off.

PR #261, opened 2018-03-15
Added history icon to library.

odmln:


odmln:
Tests are failing because of a `-webkit-text-decoration: none;` snapshot bug that has nothing to do with the icon.

Issue #219, opened 2018-02-01
Ensure theme is included as a default prop for all styled components


jxnblk:
We include the `theme` as a default prop to styled-components to make setting up tests in other libraries easier. Some components are missing theme as a default prop, including:
- [ ] Radio
- [ ] Checkbox
- [ ] Flag

PR #262, opened 2018-03-20
Feature/add theme default prop

LakshmiSattu:
Relates to Issue #219
Adding theme as a default prop to all styled components that have missing theme default prop.


src/Banner.js
```diff
@@ -98,7 +99,8 @@ Banner.propTypes = {
 Banner.defaultProps = {
   bg: 'green',
   textAlign: 'left',
-  showIcon: true
+  showIcon: true,
+  theme: theme
```
jxnblk: I don't *think* the composite components need to have the `theme` set as a default prop since the styled components' should have a theme prop already

src/Banner.js
```diff
@@ -99,8 +98,7 @@ Banner.propTypes = {
 Banner.defaultProps = {
   bg: 'green',
   textAlign: 'left',
-  showIcon: true,
-  theme: theme
+  showIcon: true
```
LakshmiSattu: @jxnblk  Removed theme as default prop for banner, as it is a composite component. Thanks!

msafari:
@LakshmiSattu thank you for adding this. Your changes look good but since this branch is so behind, I'll decline the PR and open a rebased one.

PR #263, opened 2018-03-22
feat(add-animationDuration-to-theme): add animation duration

mattcpchen:
Add animation duration to theme
Note: this is slightly adjusted from the original requirement. Instead of creating for variables for durationFast, durationNormal, durationSlow, durationSlowest. all these settings are wrapped within one object as  duration :{fast, normal, slow, slowest}


jxnblk:
Thanks for this! It looks like you'll need to update snapshots since this is a change to the theme. You can update snapshots by running `npm t -- -u`

PR #269, opened 2018-03-29
fix(InputField): pass px explicitely

msafari:


src/__tests__/__snapshots__/InputField.js.snap
```diff
@@ -316,7 +316,7 @@ exports[`InputField it renders a with a both icons 1`] = `
 .c1 {
   margin-left: 8px;
   margin-right: -32px;
-  margin-top: 8px;
+  margin-top: 2px;
```
msafari: You can see in the snapshot that it's now the correct value. It was supposed to be 2px for margin top but would get 8px because space[index=2] = 8

src/InputField.js
```diff
@@ -88,7 +88,7 @@ class InputField extends React.Component {
         if (child.type === Icon) {
           if (inputPosition < 0) {
             BeforeIcon = child
-            iconAdjustment = child.props.size - Icon.defaultProps.size
+            iconAdjustment = child.props.size - 24
```
msafari: this calculation is off if it's anything besides 24 (current default prop). So it shouldn't refer to a dynamic variable if it only works with one value

PR #271, opened 2018-04-03
Inherit color from parent for input

Degron:


hakimelek:
Why would the color would be set to inherit its parent? I think changing the color of the input box should be just done through an `extend`. This is a solution for flexibility even though I don't see the reason why would you change the color when consuming the Input.

jxnblk:
LGTM :shipit:

PR #275, opened 2018-04-09
Responsive flag

cenjennifer:
Making `Flag` component responsive

### DIFF
--- a/src/Flag.js
+++ b/src/Flag.js
@@ -27,12 +27,13 @@ const FlagShadow = styled(Box)`
 `
 
 const FlagRight = styled(Box)`
-  display: inline-block;
   flex: none;
   background-color: ${props => theme(`colors.${props.color}`)(props)};
-  border-radius: ${theme('radius')};
+  border-radius: 0 ${theme('radius')} ${theme('radius')} 0;
   /* for 32 x 8 triangle */
   transform: skew(-14deg);
+  position: relative;
+  z-index: 1;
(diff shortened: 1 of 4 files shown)


jxnblk:
Thanks for this! It could be helpful if you uploaded some before and after screenshots to see what the change here looks like

cenjennifer:
@jxnblk - Here's the screenshot. So desktop should still look as is, but when its mobile, this is what it looked like:
Before:
screen shot 2018-04-10 at 12 38 19 pm

After:
screen shot 2018-04-10 at 12 39 40 pm

src/Flag.js
```diff
@@ -13,29 +15,25 @@ const darkBorderColor = props => {
...
   flex: none;
-  min-height: 32px;
-  margin-left: -10px;
   background-color: ${props => theme(`colors.${props.color}`)(props)};
-  border-radius: ${theme('radius')};
+  border-radius: 0 ${theme('radius')} ${theme('radius')} 0;
   /* for 32 x 8 triangle */
   transform: skew(-14deg);
+  position: relative;
+  z-index: 1;
```
msafari: n-b: Are these z-indexes gonna be problematic if there are any z-index conflicts in the app that consumes this?

msafari: We can also always update if it is an issue

cenjennifer: Great! Thanks :)

PR #281, opened 2018-04-12
Update styled-components from v2 to v3

hakimelek:


package.json
```diff
@@ -83,7 +83,7 @@
   "dependencies": {
     "prop-types": "^15.5.10",
     "rc-slider": "^8.5.0",
-    "styled-components": "^2.1.1",
+    "styled-components": "^3.2.4",
```
jxnblk: I *think* we should go ahead and move this to `peerDependencies` and `devDependencies` as well as allowing a looser version range (`>=2.0 || >=3.0`) to ensure this doesn't break for anyone still using v2 of styled-components

jxnblk:
See #250 – we could potentially merge that one into master (v1)

package.json
hakimelek: The loose version is already there in v1.  I would love to get this to go since there are no breaking changes that we are using in v3.

jxnblk: Didn't realize that it was already in peerDependencies, but this should move to devDependencies for sure

hakimelek:
@jxnblk Yes, we can move forward with #250 since it's updating related dependencies, however, I am wondering if styled-components with React 16 playing well on a React 15 project.

jxnblk:
I doubt the styled-components library is using any React 16-specific APIs. You could install React 16 in this repo locally to do a gut check that it still works

jxnblk:
And for #250 – I'll probably bring in the changes from master into the `next` branch after this is merged

hakimelek:
Great! Styled-components v3 seems to work fine with React 15 on my local machine.

PR #286, opened 2018-04-20
Use Input value to determine label visibility in InputField

jxnblk:
Removes state from the InputField component, negating the need to pass an `onChange` prop. I've added some code to support backwards compatibility with the current API, but would like some input from @byrekt in case I'm missing something about the implementation here

### DIFF
--- a/src/InputField.js
+++ b/src/InputField.js
@@ -36,29 +36,27 @@ const getInputStyles = showLabel => {
       }
 }
 
+const noop = () => {}
+
 class InputField extends React.Component {
-  constructor(props) {
-    super(props)
-
-    let hasInitialValue
-
-    React.Children.forEach(props.children, child => {
-      if (child && child.type === Input) {
-        hasInitialValue = !!child.props.value
-      }
-    })
-
-    this.state = {
-      showLabel: hasInitialValue
-    }
+  static defaultProps = {
+    // for backwards-compatibility
+    onChange: noop
   }
 
-  onInputChange = event => {
-    this.setState({
-      showLabel: event.target.value
-    })
+  // for backwards-compatibility
+  handleChange = onChange => e => {
+    this.props.onChange(e)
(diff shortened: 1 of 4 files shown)


src/InputField.js
```diff
@@ -36,29 +36,27 @@ const getInputStyles = showLabel => {
...
-      }
-    })
+const noop = () => {}
 
-    this.state = {
-      showLabel: hasInitialValue
-    }
+class InputField extends React.Component {
+  static defaultProps = {
+    // for backwards-compatibility
```
jxnblk: These shims for backwards-compatibility could be removed in v2

src/InputField.js
```diff
@@ -36,29 +36,27 @@ const getInputStyles = showLabel => {
...
   }
 
-  onInputChange = event => {
-    this.setState({
-      showLabel: event.target.value
-    })
+  // for backwards-compatibility
+  handleChange = onChange => e => {
+    this.props.onChange(e)
+    if (typeof onChange !== 'function') return
```
BeniCheni: +1 on these `onChange` in a chained functional calls, with proper inner `onChange` type check.

src/InputField.js
```diff
@@ -36,29 +36,27 @@ const getInputStyles = showLabel => {
...
+    onChange(e)
+  }
 
-    this.props.onChange(event)
+  hasValue = () => {
+    const { children } = this.props
+    return React.Children.toArray(children).reduce(
```
BeniCheni: Trivial, but if only `children` prop is destructed from `this.props` & used, (instead of multiple destructed props), is there any benefit about the destructing pattern here over the dot-notation object access, such as using `this.props.children` directly in the array `.reduce` call? (Is it for expressive, typing-less reason for `children`?)
```
return React.Children.toArray(this.props.children).reduce( ... )
```

Issue #228, opened 2018-02-21
Add easing curves to theme


sqott-zz:
easeInOut: `0.50, 0.00, 0.25, 1.00` [Example](http://cubic-bezier.com/#.5,0,.25,1)
Used for standard animations

easeOut: `0.00, 0.00, 0.25, 1.00` [Example](http://cubic-bezier.com/#0,0,.25,1)
Used for elements entering the screen

easeIn: `0.50, 0.00, 1.00, 1.00` [Example](http://cubic-bezier.com/#.5,0,1,1)
Used for elements exiting the screen

James300:
Looks like this got merged 👍 Cleaning up this ticket now

James300:
![](https://github.trello.services/images/mini-trello-icon.png) [Add easing curves to theme](https://trello.com/c/BpA4DGAj/23-add-easing-curves-to-theme)

PR #290, opened 2018-04-23
Add easing curves

sqott:
In reference to #228 - adds standard easing curves for animation. 

easeInOut - base easing for core animations
easeOut - used for elements entering frame
easeIn - used for elements exiting frame

### DIFF
--- a/src/theme.js
+++ b/src/theme.js
@@ -152,9 +152,9 @@ export const duration = {
 }
 
 // animation easing curves
-const easeInOut = `cubic-bezier(0.5, 0, 0.25, 1)`
-const easeOut = `cubic-bezier(0, 0, 0.25, 1)`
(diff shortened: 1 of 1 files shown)


src/theme.js
```diff
@@ -151,6 +151,17 @@ export const duration = {
   slowest: 600
 }
 
+// animation easing curves
+const easeInOut = 'cubic-bezier(0.5, 0, 0.25, 1)'
+const easeOut = 'cubic-bezier(0, 0, 0.25, 1)'
+const easeIn = 'cubic-bezier(0.5, 0, 1, 1)'
+
+const easing = {
```
jxnblk: I wonder if this should be called `timingFunctions` to stick with the CSS naming convention, e.g. `animation-timing-function: ${easeInOut}`

sqott: Good call

PR #292, opened 2018-04-23
Add defaultProps to Radio

lundgren2:
Added default theme props to Radio component to avoid undefined props.


hakimelek:
Thank you for the contribution @lundgren2, I think this should help for tests as well.

jxnblk:
Thanks! This looks great! Related to #219

lundgren2:
I've added theme to defaultProps in Flag and Checkbox component according to #219 
It maybe need some cleanup to only use colors from theme in props right?

PR #293, opened 2018-04-24
fix(Flag): allow hex to be passed in as bg prop

msafari:
problem:
if we passed in hex values directly to Flag, the UI was kinda broken.
screen shot 2018-04-24 at 9 13 33 am

fixed:
screen shot 2018-04-24 at 9 33 55 pm

### DIFF
--- a/src/Flag.js
+++ b/src/Flag.js
@@ -29,7 +29,8 @@ const FlagShadow = styled(Box)`
 
 const FlagRight = styled(Box)`
   flex: none;
-  background-color: ${props => theme(`colors.${props.color}`)(props)};
+  background-color: ${props =>
+    theme(`colors.${props.color}`)(props) || props.color};
   border-radius: 0 ${theme('radius')} ${theme('radius')} 0;
   /* for 32 x 8 triangle */
   transform: skew(-14deg);
--- a/src/Flag.js
+++ b/src/Flag.js
@@ -8,7 +8,8 @@ import Box from './Box'
 const capitalize = str => str.charAt(0).toUpperCase() + str.slice(1)
 
 const darkBorderColor = props => {
-  const darkColor = theme(`colors.dark${capitalize(props.color)}`)(props)
+  const darkColor =
+    props.darkBgColor || theme(`colors.dark${capitalize(props.color)}`)(props)
   return {
     borderTopColor: darkColor,
     borderRightColor: darkColor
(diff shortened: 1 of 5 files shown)


tornglintaffychen:
Thanks for updating.

src/Flag.js
```diff
@@ -7,28 +7,36 @@ import Box from './Box'
 
 const capitalize = str => str.charAt(0).toUpperCase() + str.slice(1)
 
-const darkBorderColor = props => {
+const shadowColor = props => {
   const darkColor = theme(`colors.dark${capitalize(props.color)}`)(props)
+
+  if (!darkColor) {
```
msafari: props to @jxnblk for your suggestion. This is much better than adding a darkBgColor prop 🙇‍♀️

src/Flag.js
```diff
@@ -7,28 +7,36 @@ import Box from './Box'
...
+      `
+    }
+  }
+
   return {
-    borderTopColor: darkColor,
-    borderRightColor: darkColor
+    backgroundImage: `linear-gradient(45deg, transparent 50%, ${darkColor} 50%)`
   }
 }
```
BeniCheni: Could these two `return` statements consolidate into one `return` statement?
```
return !darkColor
  ? { ... }
  : { ... }
```
Or the ternary goes inside the returned `backgroundImage` object key, as it seems the ternary conditions return the same `backgroundImage` object key?
```
return {
  backgroundImage: !darkColor ? `...` : `...`
   }
```

Which one would you like better, if you decide to consider consolidating?

msafari: I updated it to
```
  return {
    backgroundImage: !darkColor
      ? `
        linear-gradient(45deg, transparent 50%, rgba(0, 0, 0, 0.5) 50%),
        linear-gradient(45deg, transparent 50%, ${props.color} 50%)
      `
      : `linear-gradient(45deg, transparent 50%, ${darkColor} 50%)`
  }
```

src/Flag.js
```diff
@@ -7,28 +7,32 @@ import Box from './Box'
...
+  ${shadowColor};
   position: absolute;
   bottom: 0;
 `
 
 const FlagRight = styled(Box)`
   flex: none;
-  background-color: ${props => theme(`colors.${props.color}`)(props)};
+  background-color: ${props =>
+    theme(`colors.${props.color}`, props.color)(props)};
```
jxnblk: :+1: good use of the fallback argument here

jxnblk:
LGTM - feel like if this used an SVG instead of CSS hacks, this component would probably be a little easier to work on

src/Flag.js
msafari: lol thanks for your suggestion. I didn't know about it :D

PR #294, opened 2018-04-25
Fix for travis config

jxnblk:


jxnblk:
Going to merge this in since it's not a source code change

PR #299, opened 2018-05-01
Add transition delay

sqott:
Adds standard transition delays in ms
Also denotes transition durations in ms


jxnblk:
It looks like snapshots are failing due to the theme change. I can update and push a fix

PR #300, opened 2018-05-04
Prefix icon select

zerodom30:
Added support for select in `InputField` and as a result updated the name to something more generic `FormField`.


.storybook/FormField.js
```diff
@@ -179,6 +188,72 @@ storiesOf('InputField', module)
...
+                With a super long label that doesn't get clobbered by the
+                chevron
+              </option>
+            </Select>
+          </FormField>
+        </Box>
+      </Flex>
+    ))
+  )
+  .add(
```
BeniCheni: Nice to see your PR here. I wonder what these red colored lines indicate in the code diff. Does anyone know? (Never saw before in previous PR reviews)

.storybook/FormField.js
```diff
@@ -179,6 +188,72 @@ storiesOf('InputField', module)
...
+    'FormField composed with Select',
+    withInfo(
+      'This example demonstrates how the input field will behave with a select dropdown'
+    )(() => (
+      <Flex width={900}>
+        <Box width={1 / 3} pl={3}>
+          <Text bold pt={4} pb={2}>
+            User Types a Value
+          </Text>
+          <FormField onChange={() => {}}>
```
BeniCheni: Instead of empty function in `onChange`, perhaps the PR can take on the **Storybook Addon Actions** to "_hook up_" with the `onChange` synthetic event, such as:
```
action('onChange event is triggered in FormField')
```
https://github.com/storybooks/storybook/tree/master/addons/actions#storybook-addon-actions

.storybook/FormField.js
```diff
@@ -179,6 +188,72 @@ storiesOf('InputField', module)
...
+              name="cabinClass"
+              defaultValue="Premium Economy"
+              prefixIcon="key"
+            >
+              <option>Economy</option>
+              <option>Premium Economy</option>
+              <option>Business</option>
+              <option>First Class</option>
+              <option>
+                With a super long label that doesn't get clobbered by the
```
BeniCheni: Very trivial. Perhaps as a "_super long label_" example, the content inside the `option` tag can go for the "_extreme lorem ipsum_" with usage of a `Truncate` from design system to show interoperable composition with design system components composing each other in a Storybook example?
https://pricelinelabs.github.io/design-system/Truncate

src/FormField.js
```diff
@@ -60,38 +60,43 @@ class InputField extends React.Component {
...
     React.Children.forEach(children, (child, index) => {
       if (child) {
         if (child.type === Label) {
           LabelChild = child
         }
-        if (child.type === Input) {
-          inputPosition = index
-          InputChild = child
-          inputId = child.props.id
+        if (child.type === Input || child.type === Select) {
```
BeniCheni: Seems like the `if` condition grows a little bit with the `||` operator. Since there are these `let` variables declared atop, perhaps, the "grown" `if` logic can be "_hoisted_" out to a declared variable along with the ones above?

src/FormField.js
```diff
@@ -108,8 +113,8 @@ class InputField extends React.Component {
     if (label) {
       LabelChild = <Label>{label}</Label>
     }
-    if (!InputChild) {
-      InputChild = <Input />
+    if (!FieldChild) {
+      FieldChild = <Input />
     }
```
BeniCheni: Non-blocking, outside the scope of this PR, though the observation is about these `if` and inner `if` statements inside each `React.Children.forEach` iteration. Perhaps a new issue can be log to start a discuss about potential pattern to "_dry_" out all these `if` & inner `if` statements? 😇

src/__tests__/InputField.js
```diff
@@ -24,6 +24,105 @@ describe('InputField', () => {
...
+      .toJSON()
+    expect(json).toMatchSnapshot()
+  })
+
+  test('it renders using FormField alias with Select and Icon', () => {
+    const json = renderer
+      .create(
+        <FormField onChange={() => {}}>
+          <Icon name="email" />
+          <Select id="with-left-icon" />
```
BeniCheni: Tiny test case nit: seems the test is testing "_left icon_" & "_no options_" in the composition here, could the `id` value be something that combines the essence of two characteristics of this unit test?

BeniCheni: The other thought is that maybe the `id` of the `Select` can decouple with the context to the `Icon` and just stay simple with "with-no-options", like the `id` values in other unit tests.

src/__tests__/InputField.js
```diff
@@ -24,6 +24,105 @@ describe('InputField', () => {
...
+        </FormField>
+      )
+      .toJSON()
+    expect(json).toMatchSnapshot()
+  })
+
+  test('it renders using FormField alias with Select icons on the right side', () => {
+    const json = renderer
+      .create(
+        <FormField onChange={() => {}}>
```
BeniCheni: Is the `onChange` prop required in the `FormField` composition? If yes and it's being used in any test case, perhaps we can think of ways to test `onChange` instead of the empty function?

Another thought is that, perhaps, the PR can keep it simple and finds way to not to require `onChange` in the `FormField` composition? Seems like your code coverage is 100% for `FormField`, and it's green for CI. So probably another open issue can explore how to test the `onChange` in design system unit test, or similar issue topic? (just thoughts...)

.storybook/FormField.js
zerodom30: I'm not a 100% sure here but maybe its a change in whitespace from my ide...

src/FormField.js
zerodom30: To keep the PR small I followed the existing pattern but in terms of scalability I was thinking of creating an issue to get this component cleaned up.

src/FormField.js
zerodom30: Similar to my comment above just preserving the existing logic to keep things small but a follow up would be nice to simplify things.

src/__tests__/InputField.js
zerodom30: Sure it can... I'll work on that.

src/FormField.js
BeniCheni: Nice idea @zerodom30 about opening a new issue to discuss and/or ideate about the potential "_cleaning up_" pattern of this component. Let's try out the issue when you have time. See you on the issue. 😸

BeniCheni: #301 issue has been created and refers to this thread for the thought of having an issue to discuss more simplistic JavaScript pattern to follow up to `if` & inner `if` blocks in to to-be `FormField`. Feel frees to chime in to discuss on that issue.

.storybook/FormField.js
jxnblk: Personally, I'd rather avoid custom storybook APIs as much as possible and try to keep our examples decoupled from that sort of thing

.storybook/FormField.js
jxnblk: I don't think Truncate will work as expected inside an `<option>` element, which should automatically truncate however the OS handles it

src/FormField.js
```diff
@@ -60,38 +60,43 @@ class InputField extends React.Component {
...
-    return React.Children.toArray(children).reduce(
-      (a, child) => a || (child && child.type === Input && !!child.props.value),
-      false
-    )
+    return React.Children
+      .toArray(children)
+      .reduce(
+        (a, child) =>
+          a ||
+          (((child && child.type === Input) || child.type === Select) &&
```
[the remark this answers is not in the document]
jxnblk: This could also be a function like `isFormElement(child)`

src/FormField.js
```diff
@@ -60,38 +60,43 @@ class InputField extends React.Component {
...
     let BeforeIcon
     let AfterIcon
-    let inputId
-    let inputPlaceholder
+    let fieldId
+    let fieldPlaceholder
     let iconAdjustment
 
     React.Children.forEach(children, (child, index) => {
       if (child) {
```
jxnblk: Looks like a good opportunity for a return early, e.g. `if (!child) return`

src/__tests__/InputField.js
jxnblk: That should have been fixed in #286 – I think this onChange can be removed for most tests, but we'd want a single test to ensure it works for backwards API compatibility

src/index.js
```diff
@@ -18,7 +18,8 @@ export { default as Icon } from './Icon'
 export { default as IconButton } from './IconButton'
 export { default as Image } from './Image'
 export { default as Input } from './Input'
-export { default as InputField } from './InputField'
+export { default as InputField } from './FormField'
+export { default as FormField } from './FormField'
```
jxnblk: :+1: keeps the current API intact

.storybook/FormField.js
zerodom30: I updated my test, as well as the others in the file.

src/FormField.js
zerodom30: Nice will do.

jxnblk:
Thanks, this looks great! The one thing I think we'll also need is an update to the doc. Since the docs site is expecting the filename to be `InputField.md`, I think you can update the contents and leave the filename the same. Then we can do a follow up PR for updating the site

.storybook/FormField.js
BeniCheni: Oh thanks for educating me about this nuance of `<option>` with `Truncate`. Will keep that in mind when using `<option>`.

src/FormField.js
```diff
@@ -44,7 +44,20 @@ const getInputStyles = showLabel => {
 
 const noop = () => {}
 
-class InputField extends React.Component {
+const isFormElement = element => {
+  const typesOfElements = [Input, Select]
+  let isElement = false
```
jxnblk: Might be simpler to handle this with an object/dict and do this like:
`const isFormElement = element => !!formElements[element]`

src/FormField.js
```diff
@@ -60,43 +73,52 @@ class InputField extends React.Component {
...
-        if (child.type === Icon) {
-          if (inputPosition < 0) {
-            BeforeIcon = child
-            iconAdjustment = child.props.size - 24
-          } else {
-            AfterIcon = child
-          }
+      if (!child) {
+        return
+      }
```
jxnblk: 👍

src/FormField.js
```diff
@@ -195,7 +217,7 @@ InputField.propTypes = {
       }
     })
 
-    if (!inputCount) {
+    if (!count) {
       return new Error(
         `No 'Input' child found for '${componentName}'. Please update your component to use the compound version of this component and pass an 'Input' component as the child`
```
jxnblk: NB: might be good to add `Select` in this warning

src/FormField.js
zerodom30: Then `element` would need to be a `key` not an `object`. I could do a `map` but I'm not sure if that makes the code any better.

docs/InputField.md
```diff
@@ -1,23 +1,36 @@
-# InputField
+#FormField alias FormField
```
jxnblk: Markdown syntax requires a space after the `#`. Also should this say `FormField` then list the alias below?

docs/InputField.md
```diff
@@ -1,23 +1,36 @@
...
 
-Use `<InputField />` component to combine `<Input />`, `<Label />`, and `<Icon />` components into a flexible input field that matches the style guide.
+Use `<FormField />` component to combine `<Input />`, `<Label />`, and `<Icon />` components into a flexible input field that matches the style guide.
```
jxnblk: + `<Select />`

docs/InputField.md
zerodom30: Webstorm's `md` preview feature ignores that kind of stuff - making the fix.

docs/InputField.md
```diff
@@ -1,23 +1,36 @@
...
   />
-</InputField>
+</FormField>
+```
+```.jsx
+<FormField>
+  <Label>Select One</Label>
+  <Icon name='email' size='20' />
+  <Select
+    id='form-field-4'
```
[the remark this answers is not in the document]
zerodom30: I'll update that as well.

src/FormField.js
```diff
@@ -44,7 +44,11 @@ const getInputStyles = showLabel => {
 
 const noop = () => {}
 
-class InputField extends React.Component {
+const formElements = [Input, Select]
+
+const isFormElement = element => formElements.includes(element)
```
jxnblk: :+1: so clean!

jxnblk:
LGTM :shipit:

src/index.js
hakimelek: @jxnblk  Do you think we should plan on deprecating `InputField` part of v2?

PR #305, opened 2018-05-08
Letter spacing fix

### LINKED ISSUE #295: Change letter-spacing in Badge/Stamp/Flag
Please read discussion https://github.com/pricelinelabs/design-system/issues/287

Team discussion:
We'll pull this from styled-system in each specific component that needs this letter-spacing.
If we have more text-styles we should consider adding them to theme.

odmln:
Fixes #295 — Added letterSpacings to theme and applied new values to the Badge and Stamp components.


src/Badge.js
```diff
@@ -48,7 +48,7 @@ const Badge = styled.div`
   font-size: ${props => props.theme.fontSizes[0]}px;
   font-weight: 600;
   text-transform: uppercase;
-  letter-spacing: 0.05em;
+  letter-spacing: ${props => props.theme.letterSpacings.caps};
```
jxnblk: The styled-system `theme` getter is generally better to use because it won't throw when `theme` or `theme.letterSpacings` is undefined. It can be used like this:

```
letter-spacing: ${theme('letterSpacings.caps')};`
```

See this https://github.com/jxnblk/styled-system/tree/ab0242c981fa1d94d29f607a42a9e9a284dd452c#theme

odmln: Sweet, that sounds good, ill update.

jxnblk:
LGTM :shipit:

src/Badge.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import { space } from 'styled-system'
+import { theme as themeGet, space } from 'styled-system'
```
BeniCheni: Saw you had to use alias for the imported theme as `themeGet` and the theme utility all works. +1 Really

Non-blocking, tiny nit on the name of the alias, if you agree that looking at the alias signature, it seems to be in a form of invokable signature with  "**_nounVerb_**" (i.e. "somethingDo"), I wonder if the signature can simply swap the order of the words to follow like the traditional CS naming invokable signature to be something closer to "_**verbNoun**_" (i.e. `doSomething`)?
```
themeGet('letterSpacings.caps')
```

src/__tests__/__snapshots__/Badge.js.snap
```diff
@@ -207,14 +207,16 @@ exports[`Badge renders 1`] = `
...
+  -webkit-letter-spacing: 0.025em;
+  -moz-letter-spacing: 0.025em;
+  -ms-letter-spacing: 0.025em;
+  letter-spacing: 0.025em;
   padding-left: 8px;
   padding-right: 8px;
   padding-top: 4px;
   padding-bottom: 4px;
+  background-color: #f6f8fa;
+  color: #001833;
```
BeniCheni: Could the namespaces of colors be applied? Such as: `background-color: lightGray;` & `color: text;`?

PR #306, opened 2018-05-09
Added colorStyles to the theme

### LINKED ISSUE #291: Add color pairings to theme
Would be good to add set color pairings to the theme so that components like Badge & Banner can use them. 

An example & reference can be found in @jxnblk's [styled-system](https://github.com/jxnblk/styled-system/blob/master/README.md#complex-styles).

```
colorStyles: {
  warning: {
    color: 'text',
    backgroundColor: 'orange'
  },
  error: {
    color: 'white',
    backgroundColor: 'red'
  },
}
```

odmln:
Fixes #291 — I added colorStyles to the theme and updated the styled-system package to 2.1.0 in order to support that functionality. I will need to consult the rest of the design team to figure out which components should have this functionality and then I'll add that in a separate PR. 

Example of usage: `<Badge colors="whiteOnRed" />` or `<Badge colors="danger" />`

Going to write docs for this as well — stay tuned.


src/__tests__/__snapshots__/theme.js.snap
```diff
@@ -15,6 +15,100 @@ Object {
...
+  "colorStyles": Object {
+    "blueOnLightBlue": Object {
+      "backgroundColor": "#cdf",
+      "color": "#007aff",
+    },
+    "blueOnWhite": Object {
+      "backgroundColor": "#fff",
+      "color": "#007aff",
+    },
+    "danger": Object {
```
msafari: I don't see this danger and info in the actual source code. May need to regenerate snapshots

PR #324, opened 2018-06-06
Add contribution instructions for static docs site

unscsprt:
It's not immediately obvious how to make changes to the static site docs. This should make it easier for new people to get started.

### DIFF
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -62,14 +62,13 @@ open http://localhost:8000/
 
 ### Static Docs Site
 
-Markdown and source code for the [static docs site](https://pricelinelabs.github.io/design-system/) is located in
+Markdown and source code for the [static docs site](https://pricelinelabs.github.io/design-system/) are located in
 `docs/`. This [x0](https://www.npmjs.com/package/@compositor/x0) project requires Node 7+.
 
 To run the static docs locally:
 
 ```sh
 cd docs
-npm install -g @compositor/x0
 npm install
 npm start
 ```
--- a/docs/Contributing.md
+++ b/docs/Contributing.md
@@ -63,7 +63,7 @@ open http://localhost:8000/
 ### Static Docs Site
 
 Markdown and source code for the [static docs site](https://pricelinelabs.github.io/design-system/) are located in
-`docs/`. This [x0](https://www.npmjs.com/package/@compositor/x0) project requires Node 7+.
+`docs/`. This [x0](https://www.npmjs.com/package/@compositor/x0) project requires Node 8+.
 
 To run the static docs locally:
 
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks for this! We do have *some* docs for the docs started here, might make sense to tie the two together? https://github.com/pricelinelabs/design-system/blob/master/docs/README.md

docs/Contributing.md
```diff
@@ -60,6 +60,20 @@ npm start
...
+### Static Docs Site
+
+Markdown and source code for the [static docs site](https://pricelinelabs.github.io/design-system/) is located in
+`docs/`. This [x0](https://www.npmjs.com/package/@compositor/x0) project requires Node 7+.
+
+To run the static docs locally:
+
+```sh
+cd docs
+npm install -g @compositor/x0
```
jxnblk: This step shouldn't be necessary since this is installed locally

docs/Contributing.md
```diff
@@ -60,6 +60,20 @@ npm start
 open http://localhost:8000/
 ```
 
+### Static Docs Site
+
+Markdown and source code for the [static docs site](https://pricelinelabs.github.io/design-system/) is located in
+`docs/`. This [x0](https://www.npmjs.com/package/@compositor/x0) project requires Node 7+.
```
BeniCheni: If the `compositor` installation step is not needed, you probably won't need this sentence either.

docs/Contributing.md
unscsprt: Good call; looks like just running `npm i` is sufficient.

docs/Contributing.md
unscsprt: Even though it's installed locally, `x0` is still needed to run the docs locally. I added this sentence because using node <7 causes the following error because `async` support is missing.

```
/Users/craig/git/design-system/docs/node_modules/@compositor/x0/bin/cli.js:69
      .then(async html => {
            ^^^^^
SyntaxError: missing ) after argument list```

BeniCheni: @unscsprt , thanks for the elaboration of this Node.js version observation. If Node.js version 7+ has become a requirement then.

Question to all,
What do you guys think of using the [engines](https://docs.npmjs.com/files/package.json#engines) field in the `package.json`, to take advantage on a tiny bit of the npm `package.json` features?
> You can specify the version of node that your stuff works on:

BeniCheni: @unscsprt et al,

Seems like Node.js 7+ has become a requirement. What do you guys think of using the [engines](https://docs.npmjs.com/files/package.json#engines) field of the `package.json` to take a tiny bit of advantage on npm `package.json` features?
> You can specify the version of node that your stuff works on:

PR #328, opened 2018-06-21
add logic to form field when Label has hidden prop

abautist:
When using `Form Field`, there are some unwanted style issues if we pass a valid `hidden` prop to the Label component:

```<Label hidden>Pick-up Location</Label>```

Example:
screen shot 2018-06-21 at 12 11 00 pm

This PR should resolve this:
screen shot 2018-06-21 at 3 24 40 pm

Note: If `alwaysShowLabel` is passed as a prop to `Form Field`, Label should not have a `hidden` prop because that makes no sense.

### DIFF
--- a/.storybook/FormField.js
+++ b/.storybook/FormField.js
@@ -164,6 +164,32 @@ storiesOf('FormField', module)
             </InputField>
           </Box>
         </Flex>
+        <Flex width={900}>
+          <Box width={1 / 2} pl={3}>
+            <Text bold pt={4} pb={2}>
+              Label component has hidden prop
+            </Text>
+            <InputField
+              alwaysShowLabel
+              onChange={action('onChange event is triggered in FormField')}
+            >
+              <Label hidden>Pick-up Location</Label>
+              <Input id="form-field-3" placeholder="Same as pick-up" value="" />
+            </InputField>
+          </Box>
+          <Box width={1 / 2} pl={3}>
+            <Text bold pt={4} pb={2}>
+              Label component without hidden prop
+            </Text>
+            <InputField
+              alwaysShowLabel
+              onChange={action('onChange event is triggered in FormField')}
+            >
+              <Label>Pick-up Location</Label>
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: .storybook/FormField.js [hunk 1]
```diff
@@ -164,6 +164,31 @@ storiesOf('FormField', module)
             </InputField>
           </Box>
         </Flex>
+        <Flex width={900}>
+          <Box width={1 / 2} pl={3}>
+            <Text bold pt={4} pb={2}>
+              Label component has hidden prop
+            </Text>
+            <InputField
+              onChange={action('onChange event is triggered in FormField')}
```
BeniCheni: The `onChange` event seems to be triggered in the `InputField` field, based on looking at the prop, right?

.storybook/FormField.js
```diff
@@ -164,6 +164,31 @@ storiesOf('FormField', module)
...
+              <Input id="form-field-3" placeholder="Same as pick-up" value="" />
+            </InputField>
+          </Box>
+          <Box width={1 / 2} pl={3}>
+            <Text bold pt={4} pb={2}>
+              Label component without hidden prop
+            </Text>
+            <InputField
+              alwaysShowLabel
+              onChange={action('onChange event is triggered in FormField')}
```
BeniCheni: The `onChange` event also seemed to be trigger in `InputField` here?

src/__tests__/InputField.js
```diff
@@ -242,6 +242,20 @@ describe('InputField', () => {
     expect(input.getDOMNode().getAttribute('aria-label')).toBeFalsy()
   })
 
+  test('it does not display a label when Label has `hidden` prop', () => {
```
BeniCheni: Trivial question: wondering about your decision of using back-tick ` here inside string value? (Just curious about why your prefer back-tick over, let's say, double quote ")

jxnblk:
Thanks! LGTM! :shipit:

PR #331, opened 2018-06-26
Toggle

surabhinigam:
It is a badge like component that has a hover state as well.

screen shot 2018-06-26 at 9 56 50 am


src/ToggleBadge.js
```diff
@@ -0,0 +1,46 @@
...
+import styled from 'styled-components'
+import { theme as themeGet, space } from 'styled-system'
+import theme from './theme'
+
+const ToggleBadge = styled.div`
+  border-radius: 2px;
+  display: inline-block;
+  font-size: ${props => props.theme.fontSizes[0]}px;
+  font-weight: 600;
+  background-color: ${props =>
```
broox9: not sure if this was already discussed with the design team but should the color/bg-color be configurable?

src/ToggleBadge.js
```diff
@@ -0,0 +1,46 @@
...
+    props.selected ? props.theme.colors.lightBlue : props.theme.colors.white};
```
jes708: Methinks this would be better if it defaulted to no color rather than white.

src/ToggleBadge.js
jes708: Best approach here may be to allow any color with default props as lightBlue/darkBlue

src/ToggleBadge.js
```diff
@@ -0,0 +1,46 @@
...
+
+const numberStringOrArray = PropTypes.oneOfType([
+  PropTypes.number,
+  PropTypes.string,
+  PropTypes.array
+])
+
+ToggleBadge.propTypes = {
+  selected: PropTypes.bool,
+  px: numberStringOrArray,
```
jes708: I think you might be able to simplify the space propTypes as `...space.propTypes`

https://github.com/jxnblk/styled-system/blob/master/docs/api.md#proptypes

src/ToggleBadge.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { theme as themeGet, space } from 'styled-system'
+import theme from './theme'
+
+const ToggleBadge = styled.div`
+  border-radius: 2px;
```
jes708: You may want to use the border-radius from theme, instead of hardcoding it, even though it is the same value.  See `Card` as an example.

src/ToggleBadge.js
```diff
@@ -0,0 +1,46 @@
...
+  display: inline-block;
+  font-size: ${props => props.theme.fontSizes[0]}px;
+  font-weight: 600;
```
jes708: You should grab this value from theme as well. (600 = theme.bold)

src/ToggleBadge.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { theme as themeGet, space } from 'styled-system'
+import theme from './theme'
+
+const ToggleBadge = styled.div`
+  border-radius: 2px;
+  display: inline-block;
+  font-size: ${props => props.theme.fontSizes[0]}px;
```
jes708: I think it might make sense to also allow this to be any fontSize passed in as a prop.

src/ToggleBadge.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { theme as themeGet, space, fontSize, propTypes } from 'styled-system'
+import theme from './theme'
+
+const ToggleBadge = styled.div`
```
jxnblk: Unsure of how this is intended to be used, but if this is meant to be a clickable element, it should be a `<button>` element

jxnblk: I think this is a blocker unless this element is *not* meant to be clicked

Issue #318, opened 2018-05-30
Add 'position' prop to Box


elizfunk:
It would be nice to have a position prop for box for relative and absolute children.  Might be something good for the Card component too.

jxnblk:
Personally, I think it's better to have separate position components since it's something that should be used with caution and can help make code a little more readable. For example, Rebass includes an Absolute component, which could be used like:

```jsx
<Absolute top={0} right={0}>
  <Card />
</Absolute>
```

elizfunk:
Good point regarding absolute positioning.  What about relative positioning?  It could be a boolean prop on its own.

I'm thinking here only about divs that need relative positioning for their children.

BeniCheni:
Similar to the proposed `Absolute` component, a `Relative` positive position could be "_made_" with similar supportive props to the relation position CSS rules.

If relative positions are aimed for children element, this pseudocode should demonstrate the idea in a composition with `Absolute` & `Relative` components together. Would this work for you?
```
<Absolute top={0} right={0}>
  <Relative top={5} left={5}>
    <Card />
  </Relative>
  <Relatvie bottom={10} right={10}>
    <Box />
  </Relative>
</Absolute>
```

elizfunk:
Sounds good to me.  Should we call it `RelativeBox` and `AbsoluteBox` to make it implicit that the component has the same properties/takes the same props as the `Box` component?

Just `Absolute` and `Relative` are fine too.

Also, @jxnblk mentioned creating a `Fixed` component.

PR #333, opened 2018-06-27
add absolute component

elizfunk:
resolves issue #318

### DIFF
--- a/.storybook/Absolute.js
+++ b/.storybook/Absolute.js
@@ -1,0 +1,113 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import styled from 'styled-components'
+import { Absolute, Box, Card, Flag, Flex, Icon, Image, Text } from '../src'
+
+const RelativeBox = styled(Box)`
+  position: relative;
+`
+
+const RelativeCard = styled(Card)`
+  position: relative;
+`
+
+const TallCard = styled(RelativeCard)`
+  height: 420px;
+`
+
+const SmallAbsolute = styled(Absolute)`
+  height: 100px;
+`
+
+const MediumAbsolute = styled(Absolute)`
+  height: 200px;
+`
+
+const LargeAbsolute = styled(Absolute)`
+  height: 300px;
+`
+
+const ExtraLargeAbsolute = styled(Absolute)`
+  height: 400px;
+`
+
+storiesOf('Absolute', module)
+  .add('Over an image', () => (
+    <RelativeBox width={1 / 2}>
+      <Absolute top={8} left={0}>
+        <Flag>Hello Flag</Flag>
+      </Absolute>
+      <Image src="https://www.priceline.com/home/public/assets/images/photos/photo-aruba.jpg" />
+    </RelativeBox>
+  ))
+  .add('Positioning an icon', () => (
+    <Box p={3}>
+      <RelativeCard p={4}>
+        <Text mt="8px" align="justify">
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Quisque et
+          nisl dolor. Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+          Donec quis nisi ac est elementum consequat a eu risus. Phasellus id
+          facilisis nulla. Aliquam vel semper enim, id lobortis dolor. Morbi sed
+          leo at turpis rutrum posuere. Nullam tincidunt ex vitae mi sagittis,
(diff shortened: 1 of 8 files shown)


.storybook/Absolute.js
```diff
@@ -0,0 +1,113 @@
...
+      <Absolute top={8} left={0}>
+        <Flag>Hello Flag</Flag>
+      </Absolute>
+      <Image src="https://www.priceline.com/home/public/assets/images/photos/photo-aruba.jpg" />
+    </RelativeBox>
+  ))
+  .add('Positioning an icon', () => (
+    <Box p={3}>
+      <RelativeCard p={4}>
+        <Text mt="8px" align="justify">
```
jxnblk: This can be `mt={2}`

.storybook/Absolute.js
```diff
@@ -0,0 +1,113 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import styled from 'styled-components'
+import { Absolute, Box, Card, Flag, Flex, Icon, Image, Text } from '../src'
+
+const RelativeBox = styled(Box)`
+  position: relative;
```
jxnblk: Would be good to add a `Relative` component in a follow up PR as discussed in #318

src/__tests__/Absolute.js
```diff
@@ -0,0 +1,26 @@
...
+      .create(<Absolute top={10} left={0} zIndex={2} />)
+      .toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('top', '10px')
+    expect(json).toHaveStyleRule('left', '0px')
+    expect(json).toHaveStyleRule('z-index', '2')
+  })
+
+  test('renders with flag and Child', () => {
+    const json2 = renderer.create(
```
jxnblk: Since this is scoped in the function, the variable can be `json` here as well

.storybook/Absolute.js
```diff
@@ -0,0 +1,113 @@
...
+const ExtraLargeAbsolute = styled(Absolute)`
+  height: 400px;
+`
+
+storiesOf('Absolute', module)
+  .add('Over an image', () => (
+    <RelativeBox width={1 / 2}>
+      <Absolute top={8} left={0}>
+        <Flag>Hello Flag</Flag>
+      </Absolute>
```
jxnblk: NB/FWIW: many times negative margins can achieve the same thing as positioning with fewer chances of CSS bugs

jxnblk:
Thanks for putting this together! LGTM :shipit:

.storybook/Absolute.js
elizfunk: That's my next PR : )

src/__tests__/Absolute.js
elizfunk: Good point.  Will fix.

.storybook/Absolute.js
elizfunk: Will update.

elizfunk: Updated

src/__tests__/Absolute.js
elizfunk: Updated

docs/Absolute.md
```diff
@@ -0,0 +1,19 @@
+
+# Absolute
+
+Use `<Absolute />` component to control position.
+This component is an extension of the `<Box />` component.
+
+```.jsx
+<Card>
```
hakimelek: Maybe make it more explicit that `Card` here should be `Relative`.  I would also think that a `Relative` component should come with the `Absolute` component. The change of position here will be relative to body, which I don't know if it's a behavior that a consumer would like seeing while wrapping it in a `Card`.

src/Absolute.js
```diff
@@ -0,0 +1,58 @@
...
+  PropTypes.string,
+  PropTypes.array
+])
+
+Absolute.propTypes = {
+  top: numberStringOrArray,
+  bottom: numberStringOrArray,
+  left: numberStringOrArray,
+  right: numberStringOrArray,
+  zIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.string])
```
hakimelek: I like the API 👍

.storybook/Absolute.js
hakimelek: I would vote that `Relative` should be part of this PR.

jxnblk: We discussed in the meeting and think it's fine for Relative to be a follow up PR

PR #334, opened 2018-06-27
fix toggle badge font

surabhinigam:
changes <ToggleBadge>
- makes the font bold if it is selected
- changes the default color of the text to blue


src/ToggleBadge.js
```diff
@@ -8,7 +8,8 @@ const ToggleBadge = styled.button`
   border-radius: ${props => props.theme.radius};
   border: 0;
   display: inline-block;
-  font-weight: ${props => props.theme.bold};
+  font-weight: ${props =>
+    props.selected ? props.theme.bold : props.theme.regular};
```
jxnblk: @jes708 do you mind following up with the designer on the font weight change here?

jes708: Sure

PR #336, opened 2018-06-29
Fix toggle font

surabhinigam:
Undo the change for `<ToggleBadge>` to let it change text font-weight on selection. Now it will be bold always.

### DIFF
--- a/docs/ToggleBadge.md
+++ b/docs/ToggleBadge.md
@@ -15,7 +15,7 @@ Use the `<ToggleBadge />` component to render a primitive badge.
 
 Prop | Type | Description
 ---|---|---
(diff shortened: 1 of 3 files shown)


docs/ToggleBadge.md
```diff
@@ -15,7 +15,7 @@ Use the `<ToggleBadge />` component to render a primitive badge.
 
 Prop | Type | Description
 ---|---|---
-selected | boolean | gives different background color and makes font bold
```
BeniCheni: So it seems like the font weight in `ToggleBadge` is still using `props.theme.bold}` in the new code. Would it be correct to update the doc to remove the emphasis of making the font bold.

PR #337, opened 2018-07-05
feature - new relative component; use in Absolute story

### LINKED ISSUE #318: Add 'position' prop to Box

elizfunk:
resolves #318 -- separate relative component in addition to the absolute component


src/__tests__/__snapshots__/Icon.js.snap
```diff
@@ -697,179 +697,6 @@ exports[`Icon accessible renders 1`] = `
 </svg>
 `;
 
-exports[`Icon airplane renders 1`] = `
```
hakimelek: This might have caused the test to fail. You probably have to rebase to latest and run the tests locally to check.

elizfunk: Thanks!  I'll give it a try.

elizfunk: I checked, and my local version of Master is already up to date.  All of my tests are passing locally.  No uncommitted changes.

elizfunk:
Updated Icon snapshot

.storybook/Absolute.js
```diff
@@ -1,17 +1,9 @@
 import React from 'react'
 import { storiesOf } from '@storybook/react'
 import styled from 'styled-components'
-import { Absolute, Box, Card, Flag, Flex, Icon, Image, Text } from '../src'
+import { Absolute, Card, Flag, Icon, Image, Relative, Text } from '../src'
```
BeniCheni: 👍 for organizing the imported components alphabetically.

docs/Absolute.md
```diff
@@ -1,12 +1,16 @@
 
 # Absolute
 
-Use `<Absolute />` component to control position.
-This component is an extension of the `<Box />` component.
+Use `<Absolute />` component to control position.  Use with a `<Relative>` parent/ancestor.
+
+This component is an extension of the [`<Box />`](https://pricelinelabs.github.io/design-system/Box) component.   It inherits `<Box />` props.
```
BeniCheni: Tiny code format nit: if your Markdown line is > 80 characters, you can break to next line to start with a space to continue writing, but with "_cleaner_" code format:
```
It is really long ...,
 now it continues...
```

src/__tests__/Relative.js
```diff
@@ -0,0 +1,28 @@
...
+describe('Relative', () => {
+  test('renders with top, left and zIndex props', () => {
+    const json = renderer
+      .create(<Relative top={10} left={0} zIndex={2} />)
+      .toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('position', 'relative')
+    expect(json).toHaveStyleRule('top', '10px')
+    expect(json).toHaveStyleRule('left', '0px')
+    expect(json).toHaveStyleRule('z-index', '2')
```
BeniCheni: 👍 , nice assertions on expected styles.

src/Relative.js
```diff
@@ -0,0 +1,58 @@
...
+  PropTypes.number,
+  PropTypes.string,
+  PropTypes.array
+])
+
+Relative.propTypes = {
+  top: numberStringOrArray,
+  bottom: numberStringOrArray,
+  left: numberStringOrArray,
+  right: numberStringOrArray,
```
BeniCheni: See that these position props accept one of types from `numberStringOrArray`, but your story only provides examples of the number type. It'd be great if the story can provide some examples of the String or Array type, and the doc of `Relative` can provide a brief emphasis what expected CSS rule(s) for each of these types used in the position props?

docs/Absolute.md
jxnblk: FWIW, with GFM the space at the beginning of the line isn't needed

jxnblk:
Thanks for putting this together! This looks great! I'm sure these new components will come in super handy

docs/Absolute.md
elizfunk: Updated.  Thanks for the tip!

src/Relative.js
elizfunk: Good point.  We use `numberStringOrArray` all over design-system, but we're not explaining it.

Let's discuss how we want to convey this information across all pages and make it a separate issue.

PR #339, opened 2018-07-06
fix(breakpoints): add unit to theme and use theme.breakpointsUnit in …

msafari:
Addresses https://github.com/pricelinelabs/design-system/issues/338


src/Hide.js
```diff
@@ -3,23 +3,26 @@ import PropTypes from 'prop-types'
...
+      props.theme.breakpoints[0]
+    }${unit}) and (max-width: ${getMaxWidth(props.theme.breakpoints[1])}${unit})`,
+    md: `@media screen and (min-width: ${
+      props.theme.breakpoints[1]
+    }${unit}) and (max-width: ${getMaxWidth(props.theme.breakpoints[2])}${unit})`,
+    lg: `@media screen and (min-width: ${
+      props.theme.breakpoints[2]
+    }${unit}) and (max-width: ${getMaxWidth(props.theme.breakpoints[3])}${unit})`,
+    xl: `@media screen and (min-width: ${props.theme.breakpoints[3]}${unit})`
+  }
```
hakimelek: For some reason, I feel like adding breakpointUnit would add a bit of confusion. I would prefer having this following API:

if `breakpoints` is defined as an array on integers, e.g. [32, 40, 48, 64], it will be passed with the `em` unit.
if `breakpoints` is defined as an array pixels passed as strings, e.g. ['320px', '640px', '1024px'], it will be passed with the `px` unit. The same could be applied for %. 

Hence, getting the unit from the breakpoints array values instead as defined in the theme.

msafari: interesting... I feel like we shouldn't have to set conditions to pick the unit based on what is in the array - string or number. 
the breakpoints array should be really kept as just integers imo. 
You can also see that `Hide` uses a util function that sets `maxWidth` by doing `width - 0.01`. I think this probably happens more that someone is doing some sort of math with the breakpoints.
If we force them to pass a string. Then in order for anyone to do any math, they would have to extract the number out of the string each time and then do what they need.

src/theme.js
```diff
@@ -12,6 +12,8 @@ const addAliases = (arr, aliases) =>
 
 export const breakpoints = [32, 40, 48, 64]
 
+export const breakpointsUnit = 'em'
```
BeniCheni: Per discussed in Design System WG meeting, the `breakpointsUnit` doesn't seem to need to be exported. Please consider revert?

src/Hide.js
```diff
@@ -3,23 +3,26 @@ import PropTypes from 'prop-types'
...
-  }em) and (max-width: ${getMaxWidth(props.theme.breakpoints[1])}em)`,
-  md: `@media screen and (min-width: ${
-    props.theme.breakpoints[1]
-  }em) and (max-width: ${getMaxWidth(props.theme.breakpoints[2])}em)`,
-  lg: `@media screen and (min-width: ${
-    props.theme.breakpoints[2]
-  }em) and (max-width: ${getMaxWidth(props.theme.breakpoints[3])}em)`,
-  xl: `@media screen and (min-width: ${props.theme.breakpoints[3]}em)`
-})
+const getMaxWidth = width => width - 0.01
```
BeniCheni: The Design System WG meeting ran out of time when this topic was discussed.

To recap, the suggestion was to design this `width` prop to be flexible to support features similar to [this "width (responsive)" API in `styled-system`](https://github.com/jxnblk/styled-system/blob/master/docs/api.md#width-responsive).

My proposal is to omit the calculation from width units to em units to keep this PR simplistic, and use another open issue to discuss the robust, agreed way(s) to build similar "_responsive width_" support mentioned above, and then an engineer will implement the "_responsive width_" in `Hide` with agreed approach.

PR #340, opened 2018-07-19
Add width 100% to Select for use in FormField component

jxnblk:


.storybook/FormField.js
```diff
@@ -26,167 +16,155 @@ storiesOf('FormField', module)
         <Text bold pt={4} pb={2}>
           Hey, ya turkey! Enter a word that has 5 characters in it.
         </Text>
-        <InputFieldWithController />
+        <FormFieldWithController />
```
broox9: it seems like _most_ of what is happening in this file is renaming but and adjusting for the select width. Am I right?

Issue #301, opened 2018-05-05
`FormField` Component: Simplify `if` & inner `if` Statements


BeniCheni:
During the code review of #300 from @zerodom30 (`InputField` => `FormField` + `Select` support), a thought arose that an observation on the `if` & inner `if` statements in certain logic could use an opportunity to simplify to a more "_hopefully expressive_" JavaScript pattern.

An example of the existing `if` and inner `if` statements could be viewed `InputField` source code [here](https://github.com/pricelinelabs/design-system/blob/master/src/InputField.js#L82). The following snippet is a demonstration of the current `if` & inner `if` logic block:
```
if (child) {
  if (child.type === Label) {
    ...
    }
  if (child.type === Input) {
    ...
  }
  if (child.type === Icon) {
    ...
  }
}
```

Therefore, I'd start this new open issue to "_socialize_" the thought of the `if` and inner `if` existing pattern with the "**discussion**" & "**good second issue**" labels as a proposed, potential follow-up PR after the "_approved merge_" of #300.

BeniCheni:
@zerodom30, I'd like to collaborate with you on a draft to see if we can simplify the `if` and inner `if` blocks. Would you happen to have any preliminary idea or pseudocode that's in your 🎧 already? 😄 I am happy to think on the potential JS pattern for a little bit, before I comment on this issue to propose any draft.

BeniCheni:
🙏 discussing and reviewing #341 during WG meeting. Closing this issue as well. Can reopen if anything missed.

PR #341, opened 2018-07-20
Simplify FormField imperative if & inner if-else blocks 

BeniCheni:
Resolves issue #301


src/FormField.js
```diff
@@ -83,33 +83,49 @@ class FormField extends React.Component {
...
-          AfterIcon = child
-        }
+      switch (type) {
+        case Label:
+          parseLabel(child)
+          break
+        case Input:
+          parseInputOrSelect(child, props, index)
+          break
+        case Select:
```
jxnblk: Since these two blocks are almost the same, couldn't this be written as:

```js
case Input:
case Select:
  parseInputOrSelect(child, props, index)
  break
```

src/FormField.js
```diff
@@ -83,33 +83,49 @@ class FormField extends React.Component {
     let fieldPlaceholder
     let iconAdjustment
 
-    React.Children.forEach(children, (child, index) => {
-      if (!child) {
-        return
-      }
+    const parseLabel = child => {
+      LabelChild = child
+    }
```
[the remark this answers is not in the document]
BeniCheni: Also worth noting that I've also "_hoist_" out the `let` variables and the parsing functions outside of the `render` function for these reasons:
* Keep the `render` method "_thin_"
* `let` variables & parsing functions would not need to be re-propcessed again, when re-rendering is triggered that `render` method runs again in React lifecycle.

BeniCheni:
@jxnblk & @zerodom30,

Instead of named, pure functions, and there's no need of reusability of these functions, I've changed the update to use a simple switch statement to manage the previous if blocks with parsing logic.

Consolidated Input & Select cases in the `switch` statement to use the same code.

Hope this pattern look?

src/FormField.js
```diff
@@ -84,31 +84,29 @@ class FormField extends React.Component {
     let iconAdjustment
 
     React.Children.forEach(children, (child, index) => {
-      if (!child) {
-        return
-      }
+      if (!child) return
```
BeniCheni: A curious question, in anyone has any insight:

If the `React.Children.forEach` top-level API is being used in current code, can one assume that the `child` existence check is handled by the API from React, and this bit of "_safety exit_" code can be omitted?

broox9:
simple enough, it looks good to me.  I'm not sure if react does anything to check existence of children but my _guess_ would be that it does

src/FormField.js
jxnblk: Line 89 will through an error if the child is not an object (React element)

zerodom30:
Looks good to me.

Issue #329, opened 2018-06-22
Update test name for InputField to FormField


hakimelek:
as per #310, InputField was generalized to FormField, we would want to rename the test from `InputField.js` to `FormField.js`.

BeniCheni:
@hakimelek (open creator et al),

As promised in WG meeting that I'd would try some v2 issue. 😉 , I've also opened #342 to try this v2 issue. See you on the PR.

BeniCheni:
Think this issue can be closed per the closure of related #310. We can reopen this issue if anything missed.

PR #342, opened 2018-07-21
Update tests from InputField to FormField

BeniCheni:
Update test code for v2 task to rename the references in test code from `InputField` to `FormField`.

Resolves issue #329


src/__tests__/FormField.js
```diff
@@ -0,0 +1,350 @@
+import React from 'react'
```
jxnblk: Shouldn't this be showing the diff as a `git mv`? Did you remove the `__tests__/InputField.js` file?

BeniCheni:
@jxnblk, good catch on the previously missing removal of the legacy `InputField` test & snapshot. Updated the PR to remove accordingly.

PR #345, opened 2018-08-06
for unselected badge the color is now transparent

surabhinigam:
unselected badge has transparent background

### DIFF
--- a/src/ToggleBadge.js
+++ b/src/ToggleBadge.js
@@ -12,7 +12,7 @@ const ToggleBadge = styled.button`
   font-family: inherit;
   cursor: pointer;
   background-color: ${props =>
-    props.selected ? props.theme.colors[props.bg] : props.theme.colors.white};
+    props.selected ? props.theme.colors[props.bg] : 'transparent'};
   color: ${props => props.theme.colors[props.color]};
   ${space} ${fontSize};
   &:hover {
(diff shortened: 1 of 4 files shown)


src/ToggleBadge.js
```diff
@@ -12,7 +12,7 @@ const ToggleBadge = styled.button`
   font-family: inherit;
   cursor: pointer;
   background-color: ${props =>
-    props.selected ? props.theme.colors[props.bg] : props.theme.colors.white};
+    props.selected ? props.theme.colors[props.bg] : 'transparent'};
```
jes708: Can you check on the design-system channel to see if anyone else is using this component, and if so, would a transparent background break their design?  This could be considered a breaking change under certain use cases.

surabhinigam: thats a good point.

surabhinigam: though I doubt its being used anywhere else

BeniCheni: Think another option is to make this "_transparent_" style optional, controlled by a prop name, namely 'isTransparent` perhaps? What do you guys think, @surabhinigam & @jes708?

surabhinigam: good point @BeniCheni , I have addressed the issue by taking background color for an unselected badge as prop, and keeping `transparent` as default. How does that sound to you?

PR #347, opened 2018-08-16
Feature/text badge

### LINKED ISSUE #349: Add dark color presets to Badge
Badge does not accept dark colors as bg.  Why not?

jes708:
Adding text bg to badge.  While I'm at it, does anyone know if there's a reason why not to add the other dark colors?  Fixes #349

### DIFF
--- a/src/Badge.js
+++ b/src/Badge.js
@@ -47,7 +47,10 @@ const type = props => {
       color: props.theme.colors.text
     }
   }
-  return badgeColors[props.bg]
+  return (
+    !(props.bg && props.color) &&
+    (badgeColors[props.bg] || badgeColors.lightGray)
+  )
 }
 
 const Badge = styled.div`
@@ -57,7 +60,7 @@ const Badge = styled.div`
   font-weight: 600;
   text-transform: uppercase;
   letter-spacing: ${themeGet('letterSpacings.caps')};
-  ${space} ${color} ${type};
+  ${space} ${type} ${color};
 `
 
 Badge.displayName = 'Badge'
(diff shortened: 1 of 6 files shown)


src/Badge.js
```diff
@@ -14,6 +14,10 @@ const type = props => {
       backgroundColor: props.theme.colors.lightBlue,
       color: props.theme.colors.darkBlue
     },
+    text: {
+      backgroundColor: props.theme.colors.text,
+      color: props.theme.colors.white
```
BeniCheni: Started to identify repeating code pattern here like `color: props.theme.colors.white` is being repeated by, for instance: `text` & `green`. Is it possible to avoid repeating `color` style code shared by different values of the`bg` prop support?

jes708: Based on the discussion in #349 we've gone in a different direction!  Rather than adding more colors presets to type, we're going to allow any color combination.  `type` will be left in for now to support previous use cases.

src/Badge.js
```diff
@@ -57,7 +60,7 @@ const Badge = styled.div`
   font-weight: 600;
   text-transform: uppercase;
   letter-spacing: ${themeGet('letterSpacings.caps')};
-  ${space} ${type};
+  ${space} ${type} ${color};
```
jxnblk: NB: we can probably rethink whether or not this component extends the `Box` component in v2, which would mean all the Box style props would work on this component as well

jxnblk:
LGTM – should the docs be updated as well?

jes708:
@jxnblk Docs updated.  Still LGTY?

PR #353, opened 2018-08-30
Lerna setup

jxnblk:
This PR reorganizes the repo as a monorepo using [Lerna](https://lernajs.io) to allow consumers to only install the dependencies they need and to help keep bundle sizes to a minimum.

- The main `pcln-design-system` package is moved to `packages/core`
- A `lerna.json` config file has been added to the root directory
- Lerna run scripts have been added to the root `package.json` file
- Adds a `package-lock.json` file to the core package

Following this PR, the following packages can be split from the core:
- Icons, since this is one of the larger parts of the package
- Slider, since this isn't used everywhere and requires a dependency

- [x] Move storybook directory back into root


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
-  - 8
-script: npm test -- --runInBand
```
hakimelek: When does the test be executed here?

packages/core/package.json
```diff
@@ -0,0 +1,95 @@
+{
+  "name": "pcln-design-system",
```
hakimelek: Maybe we should keep the same naming for the folder as `pcln-design-system` or `design-system` for short if we will be transitioning to be serving `@pcln/design-system` in v2.

jxnblk: This has to be the same name as the package that’s published. AFAIK we weren’t planning on renaming the package

.travis.yml
jxnblk: Travis runs `npm test` by default. The `runInBand` flag was to try to mitigate some of the bugs we were seeing with Jest

packages/core/package.json
jxnblk: Oh, I think I misread this. Were you suggesting renaming the directory from `core` to `design-system`? My only thinking there is then you end up with a sort of redundant `design-system/packages/design-system`, but happy to name it whatever makes most sense

hakimelek: Yes, but good point on the path. Following the reasoning behind using lerna, we will end up having a set of packages that would include priceline UI resources such as date-picker so maybe the root name would have to be changed to something broader than `design system` in the future. Also, I am just thinking that it would make sense to have a set of npm packages with the `@pcln` prefix following the same patterns used on other internal utils packages. `@pcln/design-system`, `@pcln/date-picker`, `@pcln/icons` would make sense to me.

jxnblk: We weren't planning on using scope for the public npm packages, but could be worth discussing at the next meeting

jxnblk:
I've moved the storybook config back into the root directory. Running `npm start` will look for stories in `storybook` directories nested within the `packages` directory

PR #356, opened 2018-09-04
Update to styled-system v3

jxnblk:
- Uses `mapProps` HOC for Flex component props aliases
- Uses built-in style functions for `display` and positioning
- Replaces old styled-system `propTypes` module with new prop types
- Adds `width` prop to Button (to replace `fullWidth`) – this is open for discussion
- Release notes for [styled-system v3](https://github.com/jxnblk/styled-system/releases/tag/v3.0.0)
- Removes `clean-tag` dependency from Icon component
- [x] Add propTypes warnings for old Flex props
- [x] map props for `fullWidth` + propTypes warnings


src/Flex.js
```diff
@@ -28,7 +35,11 @@ Flex.propTypes = {
   ...alignItems.propTypes,
   ...justifyContent.propTypes,
   ...flexWrap.propTypes,
-  ...flexDirection.propTypes
+  ...flexDirection.propTypes,
+  // deprecated
+  wrap: deprecatedPropType('flexWrap'),
```
msafari: nice! Thanks for adding these

Issue #335, opened 2018-06-29
Fix and improve docs


hakimelek:
Little things I have noticed in the docs that could be improved:

- [x] Fix live code error in `FormField`, second example. Maybe add another doc page for `InputField` to remove the confusion between the two components.
- [x] Prevent sidebar scrolling when selecting a component.
- [x] Remove comments from live code example and move them as text in markdown.
- [x] Better examples for `Box`, maybe add a background for every Box.
- [x] Better `Banner` second example, why are we showing an icon there?
- [x] Show a compound component example for Banner using Icon and Text.
- [x] Capitalize the first letter of all the descriptions in the props table.
- [x] Add a working example for `Checkbox`, could also be a recommended practice with a code snippet.
- [ ] Syntax highlighter for live code?

hakimelek:
Let's do that.

jxnblk:
Closing in favor of #364

PR #357, opened 2018-09-04
Use Next.js for docs

jxnblk:
Updates the docs site build setup to use [Next.js](https://github.com/zeit/next.js) and [MDX](https://github.com/mdx-js/mdx), which should help with performance of the site. The deploy setup is mostly unchanged and still uses gh-pages and Travis CI.

- Markdown files now live in `docs/pages` (this is a Next.js convention)
- All other site related modules are in `docs/src`
- New pages can be added to the navigation in `docs/src/navigation.js`
- A `docs/pages/template.md` file has been added to serve as a template for component docs
- Front matter support has been removed, although it wasn't being used
- Attempts to fix some of the issues in #335


docs/src/theme.js
```diff
@@ -0,0 +1,71 @@
...
+  LayoutContainer: {
+    paddingBottom: '128px'
+  },
+
+  MenuButton: {
+    padding: '8px',
+    color: blue
+  },
+
+  LiveCode: {
```
hakimelek: Might be worth adding a quick `overflow-x: hidden` to fix the text overflow. Not sure if it's the right solution here.
image

PR #361, opened 2018-09-13
Refactor FormField

### LINKED ISSUE #226: Add 'autoHide` prop to Label for use in InputField component
Currently the InputField component has a 'alwaysShowLabel' prop that determines whether the label should be there all the time. Instead we should add an optional prop to Label that looks like this:
```
<InputField>
  <Label autohide>This one shows/hides</Label>
  <Input />
</InputField>

<InputField>
  <Label>This one is always visible</Label>
  <Input />
</InputField>
```

jxnblk:
This is an update to the FormField component mostly meant to make the component work with other form field components, such as masked inputs. It also attempts to simplify some of the child parsing logic and fix styles in the Select component when used with a Label.

- Allows Label to be used with Select, without affecting the chevron position
- Adds IconField component for grouping icons with Select and Input elements
- Adds `isField` `isLabel` and `isIcon` static properties to allow for other components to be used as FormField children – this can be used for masked inputs in upcoming common components
- Uses inline styles for applying styles dynamically to children
- Simplifies child parsing logic, but makes this component a little less fool-proof
- Replaces `alwaysShowLabel` prop with Label `autoHide` prop – fixes #226


src/FormField.js
```diff
@@ -1,235 +1,77 @@
...
-      </Root>
+  if (!label) {
+    return new Error(
+      `No label found for ${componentName}. Please include a Label as a child.`
     )
   }
 }
 
 FormField.propTypes = {
-  alwaysShowLabel: PropTypes.bool,
```
hakimelek: Note for breaking change.

package.json
```diff
@@ -75,7 +75,6 @@
       "lcov",
       "html"
     ],
-    "collectCoverage": true,
```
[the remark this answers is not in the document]
jxnblk: Maybe not for this PR, but having this on by default slows down watch mode a bit. I’ll revert this change, but we should discuss this separately

packages/core/src/FormField.js
```diff
@@ -1,235 +1,78 @@
...
+          paddingBottom: showLabel ? 8 : 14
+        }
+      })
+    }
+    return child
+  })
+
+  const styledLabel =
+    label &&
+    React.cloneElement(label, {
```
jxnblk: This should also automatically apply the `htmlFor` prop like the current FormField

packages/core/src/__tests__/IconField.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import TestRenderer from 'react-test-renderer'
+import { IconField, Icon, Input } from '..'
+
+describe('IconField', () => {
+  test('renders', () => {
+    const json = TestRenderer.create(
+      <IconField>
+        <Icon name="calendar" />
+        <Input id="test" placholder="IconField" />
```
hakimelek:
```suggestion
        <Input id="test" placeholder="IconField" />
```

jxnblk:
Since the icon package was moved out since this started, I want to add the `isIcon` static property to the new icon components before merging this in

jxnblk:
Added the `isIcon` static property in #403

PR #384, opened 2018-10-01
feat: Added no underline for visited

jon06712:
Removing underlined link for visited state.

### DIFF
--- a/src/Link.js
+++ b/src/Link.js
@@ -9,6 +9,9 @@ const Link = styled.a`
   ${color} &:hover {
     text-decoration: underline;
   }
+  &:visited {
+    text-decoration: none;
+  }
 `
 
 Link.displayName = 'Link'
--- a/src/__tests__/__snapshots__/BlockLink.js.snap
+++ b/src/__tests__/__snapshots__/BlockLink.js.snap
@@ -13,6 +13,11 @@ exports[`BlockLink renders 1`] = `
   text-decoration: underline;
 }
 
+.c1:visited {
+  -webkit-text-decoration: none;
+  text-decoration: none;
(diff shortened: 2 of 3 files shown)


jxnblk:
Thanks for the PR! Would you mind adding a little more context around this change? I'm not sure I understand why we'd want this to be the default behavior for links. Maybe someone from the design team can chime in

BeniCheni:
Would like to have design rep. to review to weigh in the visited state of link & approve. cc Isaiah Miller.

wesleymartin85:
I agree with Jackson, I need a little more understanding around the need for this and what brought about this request. In general I don’t think we should make this the default behavior.

jon06712:
Well right now you have to write styles so the visited stated is not underlined. If everyone wants to write extra code all the time cool!

lauradhahn:
Per design team, the underline shouldn’t be there for visited links. only active and possibly hover if needed.

jxnblk:
So from my understanding, the behavior that @lauradhahn mentioned *is* the current behavior without this change. (see https://pricelinelabs.github.io/design-system/Link/) 

@jon06712 are you sure there isn't some other CSS that's affecting how links are styled in your app?

> Well right now you have to write styles so the visited stated is not underlined. If everyone wants to write extra code all the time cool!

As for this comment, since it's a React component, you should only need to do this once in your application. I'm not sure I understand the issue?

jon06712:
I've had to add this class more than once as the default behavior gives an underline. Maybe this isnt the exact cause, I would have to dig further but I think I'll drop this though because its too much effort. its easier to overwrite when needed.

Issue #375, opened 2018-09-21
Should imported Icons be removed from the core package?


jxnblk:

jxnblk:
After a quick glance, it looks like the following components import the current Icon component:
- Hug
- CloseButton
- IconButton
- Select
- Radio
- Checkbox
- Banner
- FormField

- The FormField will not require the Icon import with the changes in #361 
- Because the CloseButton, Select, Radio, and Checkbox components import specific icons, they do not require the entire icon library, and can be updated to import the icons individually.
- The IconButton component, could be included in the new pcln-icons package, instead of the core design system package
- The Hug and Banner component APIs could be updated to be more composable and not import the Icon component directly

Since completely removing the Icon component would require updates to all these components and increase the scope for migration, I'd suggest postponing the complete removal until after v2 is out and slating the change for a smaller v3 to follow later, but open to hearing thoughts from the team

BeniCheni:
Resolved by v2 release.

PR #369, opened 2018-09-21
WIP pcln-icons package

jxnblk:
- Adds a separate icon components package
- Intend to help reduce app bundle sizes for perf
- Icons can be imported individually
- A backwards-compatible `Icon` component is also included for migration
- Uses rollup for the bundled `Icon` component
- Uses official [material-design-icons](https://github.com/google/material-design-icons) package for non-custom icons
- [x] This PR depends on the monorepo setup in #353 
- [ ] Verify that all icon names are properly aliased for current icon names
- [ ] Remove icons from core package
- [ ] Add docs for `Icon`

Fixes #179 #288 

Related #375

PR #385, opened 2018-10-02
Add pcln-icons package

### LINKED ISSUE #179: Add exporting single Icons in addition to the Icon component
Webpack bundle analyzer results when consuming the design system components
image

jxnblk:
This is a simpler alternative to #369 that does not attempt to use the material design icons package since many of the icon names in the design system differ from the official names.

- Adds `pcln-icons` package
- Icon components can be imported individually to help reduce bundle size
- A backwards-compatible Icon component is included for migration
- Uses `@babel/runtime` to help with bundle size
- [x] Removes icons from core package
- [x] Updates core components to use the `pcln-icons` package

Fixes #179 
Related #375


packages/core/storybook/Icon.js
```diff
@@ -1,27 +1,29 @@
 import React from 'react'
 import { storiesOf } from '@storybook/react'
-import { Icon, Flex, Box, Text } from '../src'
-import icons from '../icons.json'
+import { Icon, Flex, Box, Text, Truncate } from '../src'
+import * as icons from 'pcln-icons/lib'
 
 const keys = Object.keys(icons)
 
+console.log(keys)
```
hakimelek:
```suggestion
```

babel.config.js
```diff
@@ -1,6 +1,7 @@
 module.exports = {
   presets: ['@babel/env', '@babel/react'],
   plugins: [
+    '@babel/transform-runtime',
```
hakimelek: @jxnblk Maybe you can add a comment in the code to explain the reason behind this.

hakimelek: Thank you for the examples :) it makes more sense now, it seems to be good for code reusability so I think we should keep it.

PR #398, opened 2018-10-22
fix(rc-slider): only import range to avoid bundling rc-tooltip

msafari:
Since we're doing a named import, it's gonna bundle up everything in their `index.js` file in pcln-design-system. 
For example, one of the things that gets bundled up is rc-tooltip which is only used in `createSliderWithTooltip`. But we only need Range looks like. we can reduce the size of rc-slider we're shipping out by just changing this import

### DIFF
--- a/src/Slider.js
+++ b/src/Slider.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { Range } from 'rc-slider'
+import Range from 'rc-slider/lib/Range'
 import { space, color, theme as getTheme, propTypes } from 'styled-system'
 import theme from './theme'
 


src/Slider.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { Range } from 'rc-slider'
```
jxnblk: Shouldn't Rollup handle this sort of thing? cc @hakimelek @bertya

msafari:
I agree with @hakimelek even in v2 the import should stay like this to reduce the size for the slider package.

src/Slider.js
hakimelek: I like this better than handling it in Rollup for now. However, we should probably lock the version of `rc-slider`.

jxnblk:
:shipit:

Issue #400, opened 2018-10-26
Support styled-components v4


jxnblk:
- [x] Remove `injectGlobal` from ThemeProvider
- [x] Update peerDependencies

efd1:
Also remove `.extend` : https://www.styled-components.com/docs/api#deprecated-extend

jxnblk:
@efd1 thanks, we've already covered that change in another PR

PR #402, opened 2018-10-30
fix(ThemeProvider): Removed injectGlobal from ThemeProvider so styled…

broox9:
Removed the `injectGlobal` from the ThemeProvider.  It is being [deprecated](https://www.styled-components.com/releases#v3.4.7) as of styled-components 4.0.0.  This opens up the possibility of using styled-components >=4.0.0 as a peer dependency.

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -22,7 +22,7 @@ Follow these steps to migrate from v1 to v2
 - Ensure Heading component has appropriately set styles. If your application relied on the styles with the `Heading.h1` – `Heading.h6` semantic components, use the new `textStyle` prop to set styles (e.g. `<Heading.h1 textStyle='display6' />`)
 - Account for other typographic changes (#231)
 - Update dependencies if your application has the following in its `package.json`:
-  - `styled-components` >=3.0.0 or >=4.0.0
+  - `styled-components` >=2.0.0 || >=3.0.0 || >=4.0.0
   - `styled-system` >=3.0.0
 - Account for ThemeProvider no longer using the deprecated `injectGlobal` therby not setting `body { margin: 0 }`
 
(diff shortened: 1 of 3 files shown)


MIGRATING.md
```diff
@@ -22,6 +22,7 @@ Follow these steps to migrate from v1 to v2
 - Ensure Heading component has appropriately set styles. If your application relied on the styles with the `Heading.h1` – `Heading.h6` semantic components, use the new `textStyle` prop to set styles (e.g. `<Heading.h1 textStyle='display6' />`)
 - Account for other typographic changes (#231)
 - Update dependencies if your application has the following in its `package.json`:
-  - `styled-components` >=3.0.0
+  - `styled-components` >=3.0.0 or >=4.0.0
```
BeniCheni: The supported version & format in `packages/core/package.json` seems to be supporting v2 also & using `||` to separate supported version for v2, v3 & v4. Please try this suggested change, to align the doc with the format in `package.json` to look more "_unified_"?
```suggestion
  - `styled-components` >=2.0.0 || >=3.0.0 || >=4.0.0
```

packages/core/src/ThemeProvider.js
```diff
@@ -1,15 +1,8 @@
...
-  ThemeProvider as StyledThemeProvider,
-  injectGlobal
-} from 'styled-components'
+import styled, { ThemeProvider as StyledThemeProvider } from 'styled-components'
 import nextTheme from './theme'
 
-injectGlobal`body {
-  margin: 0;
-}`
-
```
BeniCheni: Is the removal of `margin: 0` on global injected style agreed by WG?

My hunch is the 0 margin on the global of the injected storybook style would omit any ambiguous margins outside of the component being developed. So perhaps we can confirm this removal of the injected global style w. 0 margin?

broox9: we talked about it in the last meeting, most people seemed fine with it

jxnblk:
:shipit: This LGTM. As a next step, it might be good to mention setting global styles in the *Getting Started* doc: https://pricelinelabs.github.io/design-system/

jxnblk:
This fixes #400

PR #405, opened 2018-11-02
chore: Update "prepublish" to "prepare" script

BeniCheni:
Per [the DEPRECATION NOTE npm PREPUBLISH AND PREPARE doc](https://docs.npmjs.com/misc/scripts#prepublish-and-prepare), it seems to suggest that the favor is on `prepare` over `prepublish`, assuming that backward compatibility to npm 3 is trivial.

Should we update to use `prepare` script?

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Priceline Design System",
   "main": "dist/index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


jxnblk:
:shipit:

PR #406, opened 2018-11-06
chore: dep fixes from "next-deps" for "next"

BeniCheni:
This PR takes the dep fixes form `next-deps` branch & other chore updates to ensure `next` branch has "_healthy_" development workflow.


.travis.yml
```diff
@@ -6,7 +6,7 @@ after_success:
   - npm run codecov
 before_deploy:
   - cd docs
-  - npm install
+  - npm ci
```
BeniCheni: This edit uses [`npm ci`](https://docs.npmjs.com/cli/ci) as the linked npm doc suggests:
> Configure Travis to build using npm ci instead of npm install

###### Benefits
>  It can be significantly faster than a regular npm install by skipping certain user-oriented features. It is also more strict than a regular install, which can help catch errors or inconsistencies caused by the incrementally-installed local environments of most npm users.

package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "scripts": {
     "precommit": "lint-staged",
-    "prepare": "lerna bootstrap",
+    "prepare": "npm run bootstrap",
```
BeniCheni: This edit updates npm `prepare` script to reuse the existing `bootstrap` script, in order to support the `npm ci` script.

babel.config.js
```diff
@@ -3,6 +3,14 @@ module.exports = {
   plugins: [
     '@babel/transform-runtime',
     '@babel/proposal-class-properties',
-    '@babel/proposal-object-rest-spread'
+    '@babel/proposal-object-rest-spread',
+    [
+      'module-resolver',
```
jxnblk: I think for storybook, we'll want to set up any aliases in the [webpack.config.js](https://github.com/pricelinelabs/design-system/blob/next/storybook/webpack.config.js) file instead of Babel, so that the  the Babel transpiled dist files aren't affected

.travis.yml
jxnblk: 👍 this used to have issues with npm versions on Travis, but looks like this might work now

package.json
jxnblk: I'm fine with this change, but how does using `npm ci` affect this?

packages/slider/test/__snapshots__/index.js.snap
```diff
@@ -1,8 +1,68 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Slider renders 1`] = `
+.c0 {
```
jxnblk: I've been seeing this snapshot change – might be something weird going on with Lerna bootstrap

package.json
jxnblk: Yeah, I've had other issues with the `--hoist` flag. I think we'll need to make sure it works for all use-cases before changing this, though this might be the right way to go. If it is, we should remove the `bootstrap` run script because it would be duplicative with `prepare`

babel.config.js
BeniCheni: @jxnblk, with the latest rebase w. `next-deps` branch, the `babel-plugin-module-resolver` code is not needed & it's been reverted. 👍

BeniCheni:
@jxnblk,

Merged the feature branch of this PR to take the dep fixes from `next-deps` & reverted the `babel-plugin-module-resolver` code. Tested locally w. clean npm install, `npm t`, `npm run bootstrap` & `npm start`. LGTM.

You should definitely review the diff again w. the `next-deps` code merge.

package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "scripts": {
     "precommit": "lint-staged",
-    "prepare": "lerna bootstrap",
```
jxnblk: I think we can ditch the `bootstrap` npm run script if this does the same thing, e.g. `"prepare": "lerna bootstrap --hoist"`

packages/slider/package.json
```diff
@@ -23,8 +23,8 @@
     "styled-system": "^3.1.6"
   },
   "peerDependencies": {
-    "pcln-design-system": "^1.0.1",
-    "styled-components": "^3.4.9"
+    "pcln-design-system": ">=1.x || >=2.x",
+    "styled-components": ">=3.x || >= 4.x"
   },
```
jxnblk: I think `jest-styled-components` should be removed from this package as well, since it's a dev dependency in the root

package.json
BeniCheni: Ditched `bootstrap`, and use `lerna bootstrap --hoist` as the source of `prepare`.

packages/slider/package.json
BeniCheni: Removed `jest-styled-components` as well from `package.json` of the slider package.

BeniCheni:
@jxnblk,

Per previous comments ([here](https://github.com/pricelinelabs/design-system/pull/406#discussion_r231245209) & [there](https://github.com/pricelinelabs/design-system/pull/406#discussion_r231245695)), ditched `bootstrap` & updated to use `lerna bootstrap --hoist` for `prepare` script. Removed `jest-styled-components` from `package.json` of the slider package as well. Should be ready for LGTM.

jxnblk:
:shipit:

PR #407, opened 2018-11-06
Replace enzyme with react-testing-library

jxnblk:
This is a proof-of-concept for using [react-testing-library](https://github.com/kentcdodds/react-testing-library) instead of Enzyme. The idea behind react-testing-library is to promote writing good unit tests that test the component and *not* implementation details of the component.

From the README:

> You want to write maintainable tests for your React components. As a part of this goal, you want your tests to avoid including implementation details of your components and rather focus on making your tests give you the confidence for which they are intended. As part of this, you want your testbase to be maintainable in the long run so refactors of your components (changes to implementation but not functionality) don't break your tests and slow you and your team down.

From my limited use of react-testing-library, I like it a lot more than Enzyme and think we should consider using it. This PR is meant to be a demonstration of what it would look like in comparison to Enzyme, but would love to hear thoughts on this


packages/core/test/Banner.js
```diff
@@ -96,61 +98,48 @@ describe('Banner', () => {
   })
 
   test('renders close button if onClose func is provided', () => {
-    const wrapper = shallow(<Banner onClose={() => {}} />)
-    const closeButton = wrapper.find('CloseButton')
-    expect(closeButton).toHaveLength(1)
+    const { container } = render(<Banner onClose={() => {}} />)
+    const closeButton = container.querySelector('button')
+    expect(closeButton).toBeTruthy()
```
BeniCheni: `.toBeTruthy` is indeed more "_expressive_" than previous `.toHaveLength(1)`. 👍

jxnblk:
The codecov comment from above shows no change in coverage, but I confirmed again locally. I didn't add any new tests, but there were a few that I skipped that weren't actually adding to the test coverage count and that won't be needed when we change to using a composable API for the icons. Currently there's no way to tell which icon is being rendered in the DOM, which makes me think that the icons should have a `<title>` element for descriptions

jxnblk:
Since the design system components are fairly simple/small, you won't notice a huge difference between this and enzyme, but the changes in the Banner tests are a good place to look. The new tests rely on the DOM (rather than the component's props – i.e. an implementation detail) to check that the SVG is rendered, but they can't currently check to see if the *correct* icon is rendered because there aren't any sort of accessible labels on the icons – which is something we should potentially address later. I suspect you'd notice bigger differences when using this in our apps, since those components incorporate more functionality than the styled-components here.

The author of react-testing-library @kentcdodds has written a lot about writing unit tests, and I'd recommend reading some of what he's posted, including the introductory post: https://blog.kentcdodds.com/introducing-the-react-testing-library-e3a274307e65

kentcdodds:
Here's a particularly useful post I've published recently :) https://blog.kentcdodds.com/testing-implementation-details-ccb8d269586

BeniCheni:
Merging as WG discussed.

PR #408, opened 2018-11-06
(WIP) Feature/carousel

jes708:
Carousel component.  Currently WIP.  Brought in from hotel-details, which is currently used primarily for rich content experience tiles in mobile browsers.


packages/carousel/src/ArrowController.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+// import PropTypes from 'prop-types'
```
jes708: Commented this out for now.  Was getting a warning, which thwarted storybook:

```
"export 'default' (imported as 'PropTypes') was not found in 'prop-types'
```

Anyone see this before?

jxnblk: This is probably related to what #406 is attempting to resolve

icons.json
```diff
@@ -0,0 +1 @@
...0 0 0-1.598-1.087 4.805 4.805 0 0 0-1.946-.396c-.694 0-1.34.127-1.94.383a5.1 5.1 0 0 0-1.589 1.067L5.301 9.34a6.739 6.739 0 0 1 2.114-1.403c.8-.34 1.66-.51 2.582-.51a6.602 6.602 0 0 1 4.712 1.948l-1.168 1.163zM10.012 14l-2.215-2.19a3.33 3.33 0 0 1 .998-.654c.375-.16.775-.24 1.202-.24.432 0 .835.083 1.213.248.378.167.706.392.984.68L10.012 14z","legacy":true},"arrowDirectional":{"viewBox":"0 0 20 9","path":"M19.79 4.147c-.052-.053-.079-.105-.13-.158L15.752.139a.522.522 0 0 0-.707 0c-.184.21-.184.527 0 .712l3.171 3.138H.498A.517.517 0 0 0 0 4.49c0 .264.236.501.498.501h17.72L15.045 8.13a.503.503 0 0 0 0 .713.516.516 0 0 0 .367.158.497.497 0 0 0 .34-.132l3.906-3.85.158-.158c.104-.106.183-.212.183-.343a.653.653 0 0 0-.21-.37z","legacy":true},"arrowExpander":{"viewBox":"0 0 24 24","path":"M19.4840214,7.20598808 L19.4832811,5.58543204 L15.2680867,9.85739783 C14.8742195,10.2561596 14.2364062,10.2565348 13.8429091,9.85739783 C13.4486716,9.458636 13.4486716,8.81153898 13.8421687,8.41277715 L18.0581034,4.14081137 L16.4596833,4.14081137 C15.9033088,4.1411865 15.451694,3.68352947 15.451694,3.11933589 C15.4524344,2.55551743 15.9033088,2.09823553 16.4596833,2.0978604 L20.4920107,2.0978604 C21.0483
```
jxnblk: This is probably left over from the master branch. This file isn't needed in the `next` branch's setup so this should be removed

packages/carousel/.gitignore
```diff
@@ -0,0 +1,4 @@
+node_modules
+npm-debug.log
+dist
+.DS_Store
```
jxnblk: If we have the root `.gitignore` file, I don't think this one is necessary

packages/carousel/babel.config.js
```diff
@@ -0,0 +1,3 @@
+module.exports = {
+  presets: ['@babel/env', '@babel/react']
+}
```
jxnblk: Oh, just noticed the slider package has this too, but take a look at the `babel.config.js` in core and icons, which I think is the pattern we should go with for now

packages/carousel/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "coverageReporters": [
+      "lcov",
+      "html"
+    ],
+    "testMatch": [
+      "<rootDir>/test/**/*.js"
+    ]
+  },
+  "dependencies": {
+    "react-swipeable": "4.3.0"
```
jxnblk: Use a `^` here so that this can be updated with patches and minor bumps

packages/carousel/src/ArrowController.js
```diff
@@ -0,0 +1,55 @@
...
+    props.iconDirection === 'chevronLeft'
+      ? '8px 10px 2px 6px'
+      : '8px 6px 2px 10px'};
+  height: 40px;
+  width: 40px;
+`
+
+const ArrowIcon = styled(Icon)`
+  width: 24px;
+  height: 24px;
```
jxnblk: These *should* be the defaults for the Icon, so I don't think this needs extra styling

packages/carousel/src/ArrowController.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+// import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { Icon } from 'pcln-design-system'
```
jxnblk: This should probably be imported from `pcln-icons` directly – and, if this is just arrow icons, import the single icons as opposed to the shim for backwards compatibility

packages/carousel/src/index.js
```diff
@@ -0,0 +1,228 @@
...
+        key={mapIndex}
+        mx={2}
+        mt={2}
+        bg={mapIndex === currentItemIndex ? 'blue' : 'borderGray'}
+        onClick={this.updateSelectedItem.bind(this, mapIndex)}
+      />
+    ))
+  }
+
+  componentWillReceiveProps(nextProps) {
```
jxnblk: This is a deprecated component method – generally you shouldn't need to derive state from props changes. Guessing this could be handled in the render method at first glance

packages/carousel/src/index.js
```diff
@@ -0,0 +1,228 @@
...
+
+    const { shouldShowArrows } = this.props
+
+    const currentItemIndex =
+      this.props.currentItemIndex || this.state.internalCurrentItemIndex
+
+    return (
+      <Box {...this.props}>
+        <OverflowContainer w={1} mb={3}>
+          {shouldShowArrows && (
```
jxnblk: Should this be a node type prop for composition?

packages/carousel/src/index.js
```diff
@@ -0,0 +1,228 @@
...
+            onSwipingRight={this.handleSwipingRight}
+            onSwipedLeft={this.handleSwipedLeft}
+            onSwipedRight={this.handleSwipedRight}
+            trackMouse
+          >
+            <SwiperWrap
+              translateX={-92 * currentItemIndex + currentItemTranslate + 4}
+              isSwiping={isSwiping}
+            >
+              {this.mapItems()}
```
jxnblk: Method names like this tend to suggest that this should be its own component, e.g. `ItemsList`

packages/carousel/src/index.js
```diff
@@ -0,0 +1,228 @@
...
+      </Box>
+    )
+  }
+}
+
+const OverflowContainer = styled(Relative)`
+  overflow: hidden;
+`
+
+const Dot = styled(Box)`
```
jxnblk: Should Dot/Dots be a component in core that can be used on its own? Not sure, but there might be other use-cases for this

packages/carousel/src/index.js
```diff
@@ -0,0 +1,228 @@
...
+  top: 38%;
+  z-index: 1;
+`
+
+const PositionedArrowRightController = styled(PositionedArrowLeftController)`
+  right: 0;
+`
+
+const SwiperWrap = styled.div`
+  transition-property: all;
```
jxnblk: This can be expensive to transition all properties, rather than something like `transform`

packages/carousel/src/index.js
```diff
@@ -0,0 +1,228 @@
...
+`
+
+// Carousel.propTypes = {
+//   children: PropTypes.arrayOf(PropTypes.node).isRequired,
+//   currentItemIndex: PropTypes.number,
+//   shouldShowArrows: PropTypes.bool
+// }
+
+Carousel.defaultProps = {
+  children: []
```
jxnblk: It's better to make use of `React.Children.toArray` to ensure that `props.children` is an array

packages/carousel/storybook/Carousel.js
```diff
@@ -0,0 +1,41 @@
...
+  'https://via.placeholder.com/150/abcdef/111111?text=8',
+  'https://via.placeholder.com/150/abcdef/111111?text=9',
+  'https://via.placeholder.com/150/abcdef/111111?text=10',
+  'https://via.placeholder.com/150/abcdef/111111?text=11',
+  'https://via.placeholder.com/150/abcdef/111111?text=12',
+  'https://via.placeholder.com/150/abcdef/111111?text=13',
+  'https://via.placeholder.com/150/abcdef/111111?text=14',
+  'https://via.placeholder.com/150/abcdef/111111?text=15',
+  'https://via.placeholder.com/150/abcdef/111111?text=16',
+  'https://via.placeholder.com/150/abcdef/111111?text=17'
```
jxnblk: I'm not sure about using an external URL for unit tests – can these be fake image URLs instead?

packages/carousel/src/index.js
```diff
@@ -0,0 +1,228 @@
...
+      isSwiping: false
+    }
+
+    this.handleSwipingLeft = this.handleSwipingLeft.bind(this)
+    this.handleSwipingRight = this.handleSwipingRight.bind(this)
+    this.handleSwipedLeft = this.handleSwipedLeft.bind(this)
+    this.handleSwipedRight = this.handleSwipedRight.bind(this)
+    this.updateSelectedItem = this.updateSelectedItem.bind(this)
+    this.mapItems = this.mapItems.bind(this)
+    this.mapDots = this.mapDots.bind(this)
```
unscsprt: Does design-system support class property syntax?

Issue #102, opened 2017-10-13
Use font-smoothing on Button component


jxnblk:
Add `-webkit-font-smoothing: antialiased` to the Button component per discussion with @odmln 

See https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth

hakimelek:
Should we actually add that to the [Base Component](https://github.com/pricelinelabs/design-system/blob/master/src/ThemeProvider.js#L13 ) in the `ThemeProvider`?

PR #414, opened 2018-11-16
WIP: feature / Modal

bertya:
Work in progress. Open to review. Especially how what props it should have.


packages/modal/README.md
```diff
@@ -0,0 +1,24 @@
...
+```jsx
+import Modal from 'pcln-modal'
+```
+
+```jsx
+<Modal
+  isOpen={this.state.isOpen}
+  onClose={handleClose}
+  bg="white"
+  title="Modal title"
```
jxnblk: This `title` prop looks like it should be `props.children` – might be worthwhile to expand on the README a bit more to lock down the API before jumping into the implementation

bertya: Yes, probably finding a chance with Wes and you to figure out what level of flexibility we want to give people and then knock down the API's

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,225 @@
...
+  disableCloseButton: PropTypes.bool,
+  enableOverflow: PropTypes.bool,
+  onClose: PropTypes.func,
+  bg: PropTypes.string,
+  zIndex: PropTypes.number,
+  title: PropTypes.string,
+  headerBg: PropTypes.string,
+  imgMode: PropTypes.bool,
+  className: PropTypes.string,
+  header: PropTypes.number
```
James300: Should header be `number`?

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,225 @@
...
+}) => (
+  <Transition
+    items={isOpen}
+    config={config.gentle}
+    from={{ opacity: 0, transform: 'scale(0.5)' }}
+    enter={{ opacity: 1, transform: 'scale(1)' }}
+    leave={{ opacity: 0, transform: 'scale(0.5)' }}
+  >
+    {isOpen =>
+      isOpen &&
```
rykimchi: If I'm interpreting this correctly, you can accomplish the same thing with the transition component. There is a property you can control with transition to accomplish the same thing. 

http://reactcommunity.org/react-transition-group/transition/#Transition-prop-unmountOnExit

I think we should allow the implementer to choose if the children should remain mounted or unmounted anyways.

packages/modal/README.md
```diff
@@ -0,0 +1,24 @@
+
+# pcln-slider
```
BeniCheni: This seems to be `pcln-slider`?

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,225 @@
...
+                        id="pcln-modal-close"
+                        onClick={onClose}
+                        ml="auto"
+                      />
+                    )}
+                  </StyledHeader>
+                )}
+                {header !== 2 && !disableCloseButton && (
+                  <FloatCloseButton
+                    id="pcln-modal-close"
```
rykimchi: Is there a reason you need this to be an ID? 
We should avoid using ids to make these components as resuable as possible.

BeniCheni:
Posted an issue to ask about `style.css` in [#102 in `reach-ui`](https://github.com/reach/reach-ui/issues/102).

packages/modal/src/Modal.js
bertya: @rykimchi the ids are not used by the modal component itself at all. the purpose of ids is that the modal has so many layers, simply accepting className makes style change to deeper level hard, so these ids are helping people on that. Open to other suggestion.

jxnblk:
Closing in favor of #448

PR #416, opened 2018-11-18
Remove deprecated extend

sdalonzo:
this pr removed use of the deprecated `extend` from styled-components and resolves all outstanding warnings thrown when using the design system. this should reduce the noise for users


sdalonzo:
https://github.com/pricelinelabs/design-system/issues/413

src/Banner.js
```diff
@@ -91,7 +91,7 @@ Banner.propTypes = {
   iconName: PropTypes.string,
   onClose: PropTypes.func,
   showIcon: PropTypes.bool,
-  text: PropTypes.string,
+  text: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
jxnblk: `PropTypes.node` should include text nodes (strings) as well IIRC

src/Checkbox.js
```diff
@@ -7,15 +7,15 @@ import { Icon, Box } from '..'
 const Checkbox = props => {
   const { disabled, size } = props
   return (
-    <CheckBoxWrapper disabled={disabled}>
+    <CheckBoxWrapper disabled={disabled} theme={props.theme}>
```
jxnblk: If this is a styled component, this shouldn't need the `theme` prop to be passed explicitly

src/Icon.js
```diff
@@ -44,9 +44,11 @@ const Base = ({ name, size, legacy, ...props }) => {
   const icon = getPath({ name, legacy })
   if (!icon) return false
 
+  const { theme, ..._props } = props
```
jxnblk: `theme` can be pulled out in the arguments above if it needs to be

src/Tooltip.js
```diff
@@ -110,10 +110,12 @@ const defaultProps = {
 }
 
 const Tooltip = ({ children, ...props }) => {
+  const { align, ..._props } = props
```
jxnblk: I think we should avoid destructuring props twice – any reason not to define `align` in the line above?

src/Tooltip.js
```diff
@@ -110,10 +110,12 @@ const defaultProps = {
...
+
   return (
     <div style={{ position: 'relative', zIndex: props.zIndex }}>
-      <TooltipContent p={2} mb={3} mt={2} {...props}>
-        {children}
+      <TooltipContent p={2} mb={3} mt={2} {..._props}>
+        <Text align={'center'}>{children}</Text>
```
jxnblk: Surprised that Prettier didn't pick this up
```suggestion
        <Text align="center">{children}</Text>
```

src/__tests__/Box.js
```diff
@@ -37,4 +37,9 @@ describe('Box', () => {
     expect(json).toMatchSnapshot()
     expect(json).toHaveStyleRule('background-color', theme.colors.green)
   })
+
+  test('align prop triggers warning', () => {
+    const json = renderer.create(<Box align="center" />).toJSON()
+    expect(json).toThrowErrorMatchingSnapshot()
```
jxnblk: A prop type warning shouldn't throw an error, right?

sdalonzo: https://github.com/pricelinelabs/design-system/blob/master/src/Box.js#L46 it's returning an Error instance. i needed to replace missing coverage. i'm open to suggestions

src/Checkbox.js
sdalonzo: after replacing `extend` with `styled()`, this test was failing: https://github.com/pricelinelabs/design-system/blob/master/src/__tests__/Checkbox.js#L16

src/Banner.js
sdalonzo: update to just use node

src/__tests__/Box.js
jxnblk: Right, the prop type definition function will return an Error, but React will only `console.warn` in this case, not throw, and only in development

Here's an example of mocking console in the FormField tests https://github.com/pricelinelabs/design-system/blob/2eaa75ffb651dbcd8ee364f262cfd7328b1ce9e1/src/__tests__/FormField.js#L9

src/Checkbox.js
jxnblk: I think that's probably a case where we can update the snapshot??

src/Banner.js
```diff
@@ -91,7 +91,7 @@ Banner.propTypes = {
   iconName: PropTypes.string,
   onClose: PropTypes.func,
   showIcon: PropTypes.bool,
-  text: PropTypes.string,
+  text: PropTypes.node,
```
BeniCheni: I could be missing something important, but what is the benefit / reason of this prop type change for `text` prop?

src/__tests__/Input.js
```diff
@@ -3,21 +3,23 @@ import renderer from 'react-test-renderer'
 import 'jest-styled-components'
 import { Input } from '..'
 
+const id = 'fake-test-id'
```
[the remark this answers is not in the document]
sdalonzo: PropType warnings when running the tests: id is marked isRequired
https://github.com/pricelinelabs/design-system/blob/master/src/Input.js#L53

src/Banner.js
sdalonzo: when running the tests, this line specifically https://github.com/pricelinelabs/design-system/blob/master/src/__tests__/Banner.js#L40, if throws a PropType warning when using JSX instead of a raw string

sdalonzo: Since there is an explicit test case for it, I figured it must be a use case that we intend to support.

src/Tooltip.js
```diff
@@ -109,11 +109,11 @@ const defaultProps = {
   zIndex: 'auto'
 }
 
-const Tooltip = ({ children, ...props }) => {
+const Tooltip = ({ children, align, ...props }) => {
   return (
     <div style={{ position: 'relative', zIndex: props.zIndex }}>
       <TooltipContent p={2} mb={3} mt={2} {...props}>
-        {children}
+        <Text align="center">{children}</Text>
```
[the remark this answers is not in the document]
sdalonzo: reverted this change, thanks for feedback. the important part is not passing `align` as a prop to `Box`

PR #417, opened 2018-11-18
Card should accept borderWidth={0}

sdalonzo:
https://github.com/pricelinelabs/design-system/issues/383

### DIFF
--- a/src/Card.js
+++ b/src/Card.js
@@ -23,9 +23,10 @@ const boxShadow = props => {
 }
 
 const boxBorder = props => ({
-  border: `${props.borderWidth}px solid ${
-    props.theme.colors[props.borderColor]
-  }`
+  border:
+    props.borderWidth === 0
+      ? '0'
+      : `${props.borderWidth}px solid ${props.theme.colors[props.borderColor]}`
 })
 
 const Card = styled(Box)`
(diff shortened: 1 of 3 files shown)


src/__tests__/Card.js
```diff
@@ -90,4 +90,10 @@ describe('Card', () => {
       '0 0 2px 0 rgba(0,0,0,.08),0 2px 8px 0 rgba(0,0,0,.16)'
     )
   })
+
+  test('renders border 0 without warning', () => {
+    const json = renderer.create(<Card borderWidth={0} />).toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('border', '0')
```
jxnblk: This isn't actually testing whether the console warning is being fired or not. I think we have some other places where we use `jest.fn()` to test prop types

src/Card.js
```diff
@@ -36,7 +37,7 @@ Card.propTypes = {
   boxShadowSize: PropTypes.oneOf(['sm', 'md', 'lg', 'xl']),
   borderColor: PropTypes.string,
   ...propTypes.borderRadius,
-  borderWidth: PropTypes.oneOf([1, 2])
+  borderWidth: PropTypes.oneOf([0, 1, 2])
```
BeniCheni: Is there actually a use case where a `Card` would have "_clear/transparent_" border?

(Confirming if the original code actually intended to force all `Card`s to have at least a 1px border.)

src/__tests__/Card.js
jxnblk: Are you planning to update the test here like you did in #416?

sdalonzo: yes, sorry, just been busy

sdalonzo: pushed a similar use of console.error.mock.calls

sdalonzo:
@jxnblk made that update to check the calls of the mocked console.error. I think this should be good to go now

sdalonzo:
Will consider merge at our meeting tomorrow

PR #419, opened 2018-12-02
Text line height

sdalonzo:
naive implementation of allowing lineHeight prop for Text.

@jxnblk I wasn't sure the best way to handle this using `styled-system`, since fontSize is exported with the responsive sizing but lineHeight is not

https://github.com/jxnblk/styled-system/blob/master/src/index.js#L261
vs
https://github.com/jxnblk/styled-system/blob/master/src/index.js#L304

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-design-system",
-  "version": "1.0.3",
+  "version": "1.0.4",
   "lockfileVersion": 1,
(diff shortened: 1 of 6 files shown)


sdalonzo:
https://github.com/pricelinelabs/design-system/issues/412

sdalonzo:
https://github.com/pricelinelabs/design-system/issues/322 as well

Issue #401, opened 2018-10-26
Enable es6 build exports


hakimelek:
Currently the design-system transpiles all the files separately using babel into a `dist` folder. We would like to support ES6 module builds using Rollup. This should enable treeshaking for apps using webpack. cc @bertya @jxnblk

hakimelek:
I have started a [branch](https://github.com/pricelinelabs/design-system/tree/feat/es-build) which currently include a rollup config that adds transpiling a `dist/pcln-design-system.esm.js` file to the `dist` folder and keeping the same babel build for backward compatibility. I tested the build in storybook and will have to do more testing on consuming apps.

dansternfeld1:
We need to extend Rollup and ESM export to all exported components to prevent publishing issues related to the components without those exports right now (icons, popover, autocomplete, etc)

sdalonzo:
Issue is resolved by the new build process in v3.0.0

PR #421, opened 2018-12-07
Rollup pattern

bertya:
After close review, I think Rollup still should not live by itself as a script package. Cause if we want it as a package to work with different packages (like more than DS), then devs need to config case by case. Which means we are losing the point of the script itself (deal with common configs for packages under DS).

This PR is demonstrating my suggesting pattern. Basically, the `build.js` file handles the common parts. the `bundle.config.js` under each package folder handles the specific config to that package.


hakimelek:
Should resolve #401.

tasks/build.js
```diff
@@ -0,0 +1,70 @@
...
+  const outputOptions = {
+    file: `dist/index.${format}.js`,
+    format: format
+  }
+
+  const result = await rollup.rollup(inputOptions)
+  await result.write(outputOptions)
+}
+
+async function build(bundleLabel) {
```
jxnblk: What does having a custom script do beyond what the Rollup CLI can do? Personally I'd rather see a static config file that is consumed by Rollup. Rimraf could be handled with an npm script, but I'm not seeing anything else that requires the separate node script

bertya: At this point, everything could be done with rollup config file.  Personally, I think there are 3 major benefits of this.

1. it's easier to add the global level configuration.  Like the sharable plugins.  Yes, it could be done via import a root level `rolluo.config.js` file, but it looks in each package, it still requires a lot of changes to that import. Just not clean when the bundles are getting complicated.

2. Easier to extend. Now it's just the Rimraf function in addition to rollup, but there could be more.  https://github.com/facebook/react/tree/master/scripts/rollup could show where this pattern will go.

3. With a bundle script, to a point where we are switching to another tool, it will be easier. 

However, it really depends how complex the DS will go. Since I have no idea of that, I'm using something I think will be useful in that situation.  At this point, there is no obvious gain from it. So if you think rollup config file could be good to handle most our use cases, I'm good with that too.

jxnblk: Thanks for the explanation! Might be worth digging into point #1 a little deeper – maybe in the weekly meeting.

> it really depends how complex the DS will go

As I tend to think generally in software development, I think we should solve for the problems we have now, not the problems that we *might have* in the future, if that makes sense

packages/core/package.json
```diff
@@ -2,9 +2,9 @@
   "name": "pcln-design-system",
   "version": "2.0.0-beta.5",
   "description": "Priceline Design System",
-  "main": "dist/index.js",
+  "main": "dist/index.esm.js",
```
BeniCheni: Just double checking. Would we have support concern `esm` in dist build?

tasks/build.js
bertya: definitely, since you have a much clearer vision of DS, I trust you in that

packages/core/package.json
jxnblk: ES modules have to be in dist for treeshaking to work correctly with webpack AFAIK

packages/slider/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "pcln-slider",
   "version": "2.0.0-beta.5",
   "description": "React component for range sliders",
-  "main": "dist/index.js",
+  "main": "dist/index.esm.js",
```
hakimelek: As a good practice we should keep the `main` flag for `cjs` and the `module` flag for `esm` module. 
```
  "module": "dist/index.esm.js",  // for ES modules
  "main": "dist/index.cjs.js",  // for CommonJS modules
```

Issue #420, opened 2018-12-04
Bare bones tab component


abautist:
I built a generalized tabs components that is flexible enough to be shareable across the org. Would it be appropriate to add it as an optional package associated with the design system (similar to carousel, modal..)? 

If this is a direction we'd be interested in going, I can create a PR to continue the discussion.

BeniCheni:
+1 for building tab component to `v2` packages/ workspace.

But does "bare-bones" mean no design at all? Can confirm in next WG meeting about any general design for tab component.

abautist:
The component is styled in PCLN's DS theme. Bare bones in that the only functionality is clicking the tab labels to change active tab and display active tab's content. Content can be composite components, so hopefully super flexible/customizable. I'll open a PR to solicit more feedback on the code.

Example designs:
<img width="835" alt="screen shot 2018-12-06 at 3 41 07 pm" src="https://user-images.githubusercontent.com/5533800/49611002-34104380-f96e-11e8-8b9f-d0a61c3bb69d.png">


<img width="443" alt="screen shot 2018-12-04 at 2 35 25 pm" src="https://user-images.githubusercontent.com/5533800/49611008-38d4f780-f96e-11e8-8a3a-52c118415d33.png">

PR #422, opened 2018-12-09
WIP: Tabs

abautist:
Work-in-progress Tabs component, feedback appreciated!

TODO:
- [ ] chat with Wes about design considerations
- [ ] write documentation

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "private": true,
   "scripts": {
     "precommit": "lint-staged",
-    "prepare": "lerna bootstrap",
+    "prepare": "lerna bootstrap --hoist",
     "clean": "lerna run clean",
     "changed": "lerna changed",
     "diff": "lerna diff",
--- a/packages/tabs/README.md
+++ b/packages/tabs/README.md
@@ -1,0 +1,19 @@
+
+# tabs
+
+React tabs component
+
+```sh
+npm i pcln-tabs
+```
+
+```jsx
+import Tabs from 'pcln-tabs'
+```
+
+```jsx
+
+```
+
+## Props
+
--- a/packages/tabs/src/index.js
+++ b/packages/tabs/src/index.js
@@ -1,0 +1,63 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Flex, Text } from 'pcln-design-system'
+import styled from 'styled-components'
+
+const TabTitle = styled(Text)`
+  border-bottom: ${props =>
+    props.isActive ? `4px solid ${props.activeColor}` : 'none'};
+  cursor: pointer;
+`
+
+class Tabs extends React.Component {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      activeTab: 0
+    }
+  }
+
+  render() {
+    const { activeColor, children, defaultColor, tabs } = this.props
+
+    return (
+      <React.Fragment>
+        <Flex bg="lightGray" px={1} pt={2}>
+          {tabs.map((tab, i) => (
+            <TabTitle
+              color={this.state.activeTab === i ? activeColor : defaultColor}
+              key={i}
+              mx={3}
+              py={2}
+              onClick={() => this.setState({ activeTab: i })}
+              isActive={this.state.activeTab === i}
+              {...this.props}
+            >
+              {tab.title}
+            </TabTitle>
+          ))}
+        </Flex>
+        <Flex bg="white" py={3} px={3}>
+          {children[this.state.activeTab]}
+        </Flex>
+      </React.Fragment>
+    )
+  }
+}
+
+Tabs.defaultProps = {
+  tabs: []
+}
+
+Tabs.propTypes = {
+  activeColor: PropTypes.string,
+  defaultColor: PropTypes.string,
+  tabs: PropTypes.arrayOf(
+    PropTypes.shape({
+      title: PropTypes.string.isRequired
+    })
+  ).isRequired
+}
+
+export default Tabs
--- a/packages/tabs/src/index.js
+++ b/packages/tabs/src/index.js
@@ -19,7 +19,8 @@ class Tabs extends React.Component {
   }
 
   render() {
-    const { activeColor, children, defaultColor, tabs } = this.props
+    const { activeColor, defaultColor, tabs } = this.props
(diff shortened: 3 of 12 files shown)


packages/tabs/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+              onClick={() => this.setState({ activeTab: i })}
+              isActive={this.state.activeTab === i}
+              {...this.props}
+            >
+              {tab.title}
+            </TabTitle>
+          ))}
+        </Flex>
+        <Flex bg="white" py={3} px={3}>
+          {children[this.state.activeTab]}
```
James300: Is this handled correctly if children is empty?

packages/tabs/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+import { Flex, Text } from 'pcln-design-system'
+import styled from 'styled-components'
+
+const TabTitle = styled(Text)`
+  border-bottom: ${props =>
+    props.isActive ? `4px solid ${props.activeColor}` : 'none'};
+  cursor: pointer;
+`
+
+class Tabs extends React.Component {
```
James300: This can be a pure component: https://codeburst.io/when-to-use-component-or-purecomponent-a60cfad01a81

The approach I have taken with component libraries in the past is that you use `PureComponent` at least to begin with and have people adjust their data to work with it. If we need to downgrade it to a regular component then that is not a breaking change for anyone. However, this is not true for the reverse, hence why it is better to start with `PureComponent`.

Do people agree with this approach? I haven't seen much use of `PureComponent` around but it is vitally important to get the full benefits of React.

packages/tabs/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+    return (
+      <React.Fragment>
+        <Flex bg="lightGray" px={1} pt={2}>
+          {tabs.map((tab, i) => (
+            <TabTitle
+              color={this.state.activeTab === i ? activeColor : defaultColor}
+              key={i}
+              mx={3}
+              py={2}
+              onClick={() => this.setState({ activeTab: i })}
```
James300: Binding functions in the render is bad practice, I see why you did it in this situation but my recommendation would be to make a separate `TabTitle` which could use `memo`. Binding in a stateless components render, especially a memoized one, is much better as it will rerender much less often.

Good explanation of memo: https://scotch.io/tutorials/react-166-reactmemo-for-functional-components-rendering-control

James300: Just noticed below that you are using enzyme for testing, memo will not work with the current version of enzyme but people are working on a fix https://github.com/airbnb/enzyme/pull/1914 I would recommend just making a stateless component for now and adding memo to it later, that is if people think this overall approach is a good one

package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "scripts": {
     "precommit": "lint-staged",
-    "prepare": "lerna bootstrap",
+    "prepare": "lerna bootstrap --hoist",
```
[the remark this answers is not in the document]
jxnblk: I *think* `--hoist` is needed for storybook to work correctly, but it was causing problems with CI tests – we definitely need to figure out how we want to handle this, but I suspect that we don't want to use `--hoist` when publishing packages

packages/tabs/README.md
```diff
@@ -0,0 +1,19 @@
...
+```sh
+npm i pcln-tabs
+```
+
+```jsx
+import Tabs from 'pcln-tabs'
+```
+
+```jsx
+
```
jxnblk: It'd be helpful to get an example in here first, so that people can comment on the public API and not worry about the implementation details at first

packages/tabs/storybook/Tabs.js
```diff
@@ -0,0 +1,111 @@
...
+const ManageTab = props => <Box>{props.confirmationNumber}</Box>
+
+const RoomTab = props => <Box>{props.hotelName}</Box>
+
+const ImportantInfoTab = props => <Box>{props.importantInfo}</Box>
+
+storiesOf('Tabs', module)
+  .add('with hardcoded elements', () => {
+    return (
+      <Tabs {...props}>
```
jxnblk: I'd much rather see a composable API for the tabs, this can make changing the behavior in an application a little more difficult to do

packages/tabs/src/index.js
```diff
@@ -0,0 +1,66 @@
...
+              isActive={this.state.activeTab === i}
+              {...this.props}
+            >
+              {tab.title}
+            </TabTitle>
```
jxnblk: We'll definitely need ARIA attributes for a component like this. Here's one example: http://heydonworks.com/practical_aria_examples/#tab-interface

jxnblk:
Fixes #420

wesleymartin85:
@abautist @jxnblk here is the link to the [designs](https://priceline.invisionapp.com/d/main#/projects/prototypes/16289868). If I am missing anything let me know.

package.json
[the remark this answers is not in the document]
jxnblk: I think @BeniCheni has this covered in #425 – which I'm guessing will land before this PR

jxnblk:
@wesleymartin85 can you please include screenshots as images going forward? This is an open source project and other contributors will be unable to see internal services like the link you provided

packages/tabs/src/index.js
jxnblk: Be sure to see the [official docs on `React.memo`](https://reactjs.org/docs/react-api.html#reactmemo) and the [Optimizing Performance](https://reactjs.org/docs/optimizing-performance.html) page

package.json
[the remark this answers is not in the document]
BeniCheni: FYI that #425 landed to `next` with the `--hoist` flag in `prepare` script. All set!

packages/tabs/storybook/Tabs.js
unscsprt: I've used this library in the past and its API seems pretty reasonable. Tab titles and tab content are composed using separate wrapper components.

https://github.com/reactjs/react-tabs#basic-example

abautist:
Closing this PR for now - outdated. If Tabs get picked up again - look at https://github.com/reactjs/react-tabs#basic-example

PR #444, opened 2019-01-11
Update text align example in docs for Text component

unscsprt:
Example for centered text didn't match the actual prop name.

### DIFF
--- a/docs/pages/Text.md
+++ b/docs/pages/Text.md
@@ -9,7 +9,7 @@ Use the `Text` component to control font size, weight, alignment, and color.
 ```
 
 ```.jsx
-<Text textAlign='center'>
+<Text align='center'>
   Centered Text
 </Text>
 ```


jxnblk:
Looks like I made that change when I updated the docs https://github.com/pricelinelabs/design-system/commit/151ced50c72911e7056b1e3f9a41d793cd6d7be0#diff-5d10b5d41f34f1bc85842c651f8388ecR23

Just be aware that this prop is changing in v2 https://github.com/pricelinelabs/design-system/blob/next/MIGRATING.md

unscsprt:
@jxnblk I may be missing it, but does the migration page mention that this property on `<Text />` is changing? I just see it mentions `<Box />` and `textAlign`.

jxnblk:
Oh! Looks like the `textAlign` change for Box is listed but not Text – that needs to be updated

PR #445, opened 2019-02-04
Fix for Storybook Babel bug

jxnblk:
This fixes an issue where running storybook would not load any stories by adding a `storybook/.babelrc` file without any Babel plugins


.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 npm-debug.log
 lerna-debug.log
+yarn.lock
```
jxnblk: This disappeared somewhere along the way. This is to ensure no one accidentally checks in a Yarn lock file if they're using Yarn instead of npm

.travis.yml
```diff
@@ -6,7 +6,7 @@ after_success:
   - npm run codecov
 before_deploy:
   - cd docs
-  - npm ci
+  - npm install
```
jxnblk: Since the `docs/` directory *doesn't* currently have a lock file, `npm ci` won't work – should consider adding a lock file for the docs though

package.json
```diff
@@ -15,28 +15,28 @@
   },
   "devDependencies": {
     "@babel/core": "^7.2.2",
-    "@babel/plugin-proposal-class-properties": "^7.1.0",
-    "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
-    "@babel/plugin-transform-runtime": "^7.1.0",
-    "@babel/preset-env": "^7.1.0",
+    "@babel/plugin-proposal-class-properties": "^7.3.0",
```
jxnblk: Ran `npm update` since many of the dependencies were out of date

storybook/.babelrc
```diff
@@ -0,0 +1,6 @@
+{
+  "presets": [
```
jxnblk: This will be picked up by storybook – for some reason it looks like it was trying to pick up the root `babel.config.js` and not showing stories

jxnblk: HT @bertya

PR #448, opened 2019-02-08
Pcln modal

bertya:
Could be finally reviewed for the actual code. 

Need guidance on unit-test tho.

Please refer to README.md file and storybook folder for usage cases.


packages/modal/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "scripts": {
+    "test": "jest",
+    "prepare": "rollup --config",
+    "clean": "rm -rf node_modules dist"
+  },
+  "author": "Priceline",
+  "license": "MIT",
+  "dependencies": {
+    "@reach/dialog": "^0.1.2",
+    "react-spring": "^6.1.6"
```
abautist: @bertya one thing I ran into when I forked this modal for use in our app was that `react-spring` was breaking the bamboo build on `object.entries` since we are using node6 for prod. I ended up removing this dep/animation for now - if we want to use it I think a polyfill may be required.

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,186 @@
...
+  children,
+  imgMode,
+  width,
+  className,
+  header,
+  disableCloseButton,
+  enableOverflow,
+  height
+}) => (
+  <Transition
```
dansternfeld1: @bertya I'm leaving a comment about this in Slack for you (contains a BB link, not apropos for GH)

PR #449, opened 2019-02-19
feat(hide-for-print): Add ability to hide for print media

mjbyrnes4664:
Add print prop to Hide component
Add test case, story, and doc for Hide in print
print prop will set display: none for @media print

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,3 @@
-
 # Changelog
 
 All notable changes to this project will be documented in this file.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -13,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update colors #355
 - Use Lerna to reorganize repo into a monorepo #353
 - Removes FormField `alwaysShowLabel` prop
+- Hide `print` prop
```
BeniCheni: Thanks for updating the Changelog. I think this needs a broader discussion with the design system WG in #452 about a long-term action.

PR #453, opened 2019-02-20
chore(colors): correct yellow and lightYellow hexcodes and add to theme

abautist:
* adding yellow and lightYellow colors to the theme


packages/core/src/theme.js
```diff
@@ -134,8 +134,8 @@ const orange = '#f68013'
 const darkOrange = '#f06f20'
 const lightPurple = '#f5ebfa'
 const purple = '#70b'
-const lightYellow = '#fedc2a'
-const yellow = '#fff3c0'
+const yellow = '#fedc2a'
+const lightYellow = '#fff3c0'
```
jes708: Good catch

PR #454, opened 2019-02-25
Fix wrong external in rollup

bertya:
The rollup config put `styled-system` and `prop-types` as external, however, these 2 are dependencies for all the packages instead of peer or dev. So this should be fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "private": true,
   "scripts": {
     "precommit": "lint-staged",
-    "prepare": "lerna bootstrap --hoist",
+    "prepare": "lerna bootstrap",
     "clean": "lerna run clean",
     "changed": "lerna changed",
     "diff": "lerna diff",
@@ -42,6 +42,7 @@
     "rollup-plugin-commonjs": "^9.2.0",
     "rollup-plugin-filesize": "^5.0.1",
     "rollup-plugin-json": "^3.1.0",
+    "rollup-plugin-local-resolve": "^1.0.7",
     "rollup-plugin-node-resolve": "^3.4.0",
     "styled-components": "^3.4.10"
   },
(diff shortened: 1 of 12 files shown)


packages/core/package.json
```diff
@@ -39,7 +39,8 @@
     "react": "^16.8.2",
     "react-dom": "^16.8.2",
     "react-test-renderer": "^16.8.2",
-    "stylis": "^3.4.9"
+    "stylis": "^3.4.9",
+    "pcln-icons": "^2.0.1"
```
BeniCheni: Was this rearrange of `pcln-icons` dep to `devDependencies` and `peerDependencies` done manually? (Believe if you were to "_cd_" into `packages/core`, and run `npm i -D pcln-icons`, it'd have sorted the `pcln-icons` dep alphabetically.).

Can you update the `pcln-icons` as `devDependencies` systematically? For `peerDependencies`, you can still keep the alphabetic sort intact for your manual entry.

package.json
```diff
@@ -2,7 +2,7 @@
   "private": true,
   "scripts": {
     "precommit": "lint-staged",
-    "prepare": "lerna bootstrap --hoist",
```
BeniCheni: Is there any reason why `--hoist` flag is removed? After deep diving the [Lerna Hoisting doc](https://github.com/lerna/lerna/blob/master/doc/hoist.md), seems like using the `-- hoist` flag is a more optimized setting, and a suggested practice.

packages/core/package.json
bertya: fixed

package.json
bertya: Rollup seems to have problem resolve hoisted packages.  Right now this is the only way I found how it can work properly, open to suggestions.

BeniCheni:
Ping @bertya. Since the merge of [1668ccf](https://github.com/pricelinelabs/design-system/commit/1668ccfca641e6ba660965cb04e1700eeeed6412), can you rebase and we'll quickly review and sanity test locally again, in interest to move forward with this PR?

BeniCheni:
Ping /@bertya any update?

PR #460, opened 2019-03-11
Cleanup pcln-modal

bertya:


broox9:
I'm about halway through this PR, but it looks good so far

packages/modal/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "clean": "rm -rf node_modules dist"
+  },
+  "author": "Priceline",
+  "license": "MIT",
+  "dependencies": {
+    "@reach/dialog": "^0.1.2",
+    "react-transition-group": "^2.5.3"
+  },
+  "peerDependencies": {
+    "pcln-design-system": "^1.0.1",
```
broox9: should this be `>=1.0.1 . || >=2.0.1` ?

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,226 @@
...
+`
+
+const OverlayWrapper = styled.div`
+  display: table;
+  height: 100%;
+  width: 100%;
+`
+
+const DialogWrapper = styled(Box)`
+  display: table-cell;
```
broox9: could this be achieved with Flexbox?

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,226 @@
...
+              <DialogInnerWrapper flexDirection="column">
+                {header && <HeaderWrapper>{header}</HeaderWrapper>}
+                <ContentWrapper
+                  p={imgMode ? 0 : 16}
+                  header={header}
+                  enableoverflow={enableOverflow}
+                >
+                  {enableOverflow && !disableCloseButton && (
+                    <FloatCloseButton
+                      data-testid="close"
```
broox9: not a blocker but it would be good to namespace this test-id a bit more.  You never know if a component/host app that pulls this in and does a full render/mount test would be looking for something similar to this.

packages/modal/src/Modal.js
bertya: yes

packages/modal/package.json
bertya: This thing actually has a lot of dependencies from DS, it actually works with both, but open to discussion on this

jes708: Definitely should be `>=1.0.0 || >=2.0.0` in my opinion.  If anyone uses pcln-design-system v2 in their app, I think they would see a warning that they are not using the correct version, when in fact, there's no issue.

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,226 @@
...
+        overflow-x: hidden;
+        height: 100%;
+        -webkit-overflow-scrolling: touch;
+      `
+    }
+  }}
+`
+
+const OverlayWrapper = styled.div`
+  display: table;
```
jes708: Curious why you're using `display: table`.  Might this work as a Flex instead?

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,226 @@
...
+  width: 100%;
+`
+
+const DialogWrapper = styled(Box)`
+  display: table-cell;
+  vertical-align: middle;
+  ${props =>
+    props.enableoverflow &&
+    `
+    padding-top: 24px;
```
jes708: These values are nonstandard according to our spacing guidelines.  Can they be rounded to 16 or 32?

packages/modal/src/ModalHeader.js
```diff
@@ -0,0 +1,45 @@
...
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import { Flex, CloseButton, Text } from 'pcln-design-system'
+
+const HeaderWrapper = styled(Flex)`
+  height: 40px;
+`
+
+const StyledCloseButton = styled(CloseButton)`
+  margin-right: 8px;
```
jes708: would an inline mr={2} work instead?

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,226 @@
...
+  position: absolute;
+  top: 16px;
+  right: 16px;
+  z-index: 2;
+
+  svg {
+    vertical-align: top;
+  }
+
+  &:focus {
```
jes708: This looks dangerous for our accessibility.  How would users who rely on tabbing know when they are focused on this button?

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,226 @@
...
+  flex-shrink: 0;
+`
+
+const FloatCloseButton = styled(CloseButton)`
+  position: absolute;
+  top: 16px;
+  right: 16px;
+  z-index: 2;
+
+  svg {
```
jes708: Can this be achieved with flex align instead?

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,226 @@
...
+  width,
+  className,
+  header,
+  disableCloseButton,
+  enableOverflow,
+  height
+}) => (
+  <Transition in={isOpen} unmountOnExit timeout={500}>
+    {state => (
+      <Overlay onDismiss={onClose} zindex={zIndex} transitionstate={state}>
```
jes708: I would camel case some of these props

ie
`zIndex`
`transitionState `

packages/modal/src/ModalHeader.js
```diff
@@ -0,0 +1,45 @@
...
+  svg {
+    vertical-align: top;
+  }
+  &:focus {
```
jes708: See my other comment about accessibility concerns

packages/modal/src/Modal.js
```diff
@@ -0,0 +1,226 @@
...
+
+const ContentWrapper = styled(Box)`
+  position: relative;
+  ${props => {
+    if (!props.enableoverflow) {
+      return `
+        overflow-y: scroll;
+        overflow-x: hidden;
+        height: 100%;
+        -webkit-overflow-scrolling: touch;
```
jes708: +1 Glad you're thinking about our mobile users!

packages/modal/test/Modal.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react'
+import { render } from 'react-testing-library'
+import { Modal } from '../src/index'
+import { ThemeProvider, theme } from 'pcln-design-system'
+
+const customRender = (node, ...options) => {
```
jes708: Should this customRender function be a common utility since you use it across a couple files.  Also, could you clarify it's purpose?  Not sure how it's being used.

packages/modal/src/Modal.js
bertya: no, the table here is a trick for always vertically center it, learned from airbnb modal

packages/modal/package.json
bertya: will update this one

packages/modal/src/Modal.js
bertya: this is from Wes's design, I will talk to him about it

packages/modal/src/ModalHeader.js
bertya: I thought the button doesn't accept mr, will update

packages/modal/src/Modal.js
bertya: Reach modal by default always focus the close button, even you are not intentionally focus it, so Wes and I think it looks bad, but I will revisit with Wes about it.

packages/modal/src/Modal.js
bertya: They are passed to reach-modal, I don't know the implementation, but reach modal pass them to raw dom tags, which react will suggest not using camelcase

packages/modal/test/Modal.js
bertya: @sdalonzo might have a better idea since it's part of react-testing-lib

packages/modal/src/Modal.js
bertya: will update

jes708:
We recently had a story on the PKG team to make our modals accessible. Yang Li worked on this and met the most crucial dialog accessibility guidelines simply by wrapping our existing Modal component with react-aria-modal ( https://github.com/davidtheclark/react-aria-modal ).  We may want to discuss if we should be using this module as a dependency in the DS Modal as well in our upcoming meeting

PR #463, opened 2019-03-25
add multiAirline svg icon

abautist:
Screen Shot 2019-03-25 at 1 12 26 PM


abautist:
@wesleymartin85  - do you have input on this? Should this svg be updated, and if so are you able to format it to be more consistent with the others?

abautist:
@jes708 - I updated the format of the svg thanks to @wesleymartin85

PR #465, opened 2019-04-02
Implement support for read-only mode for Input

BeniCheni:
* Add `readonly` prop support for `Input`
  * `readonly` prop for `readonly` attribute with `true` string value
* Write test

Screen Shot 2019-04-02 at 2 48 13 PM

Screen Shot 2019-04-02 at 3 03 58 PM

### DIFF
--- a/packages/core/src/Input.js
+++ b/packages/core/src/Input.js
@@ -1,4 +1,4 @@
-import styled, { keyframes } from 'styled-components'
+import styled from 'styled-components'
 import { space, themeGet } from 'styled-system'
 import PropTypes from 'prop-types'
 import defaultTheme from './theme'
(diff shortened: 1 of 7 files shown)


packages/core/src/Input.js
```diff
@@ -17,7 +17,9 @@ const borders = ({ color, theme }) => {
   }
 }
 
-const Input = styled.input`
+const Input = styled.input.attrs({
+  readonly: props => (props.readonly ? 'true' : null)
```
sdalonzo: TIL you can use props in attrs

packages/core/src/Input.js
```diff
@@ -17,7 +17,9 @@ const borders = ({ color, theme }) => {
   }
 }
 
-const Input = styled.input`
+const Input = styled.input.attrs({
```
jes708: If this is just a `styled.input` shouldn't it be able to receive readonly inherently, without needing a readonly prop to be defined explicitly?

BeniCheni:
Thanks for pointing out the `readOnly` in camel case can be applied to existing `Input`. Closing this PR.

PR #467, opened 2019-04-05
Use babel for esm for core

bertya:
WIP of tree shaking.

Currently, only affect core package. 

The structure is not final, some files are redundant. Just demonstrate how it works.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -3,9 +3,10 @@
   "version": "2.0.4",
   "description": "Priceline Design System",
   "main": "dist/index.cjs.js",
-  "module": "dist/index.esm.js",
+  "module": "dist/index.js",
+  "sideEffects": false,
   "scripts": {
-    "prepare": "rollup --config",
+    "prepare": "babel src/ --out-dir dist --ignore spec.js,story.js; rollup --config",
     "clean": "rm -rf node_modules dist",
     "test": "jest"
   },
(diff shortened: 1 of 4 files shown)


packages/core/package.json
```diff
@@ -3,9 +3,10 @@
   "version": "2.0.4",
   "description": "Priceline Design System",
   "main": "dist/index.cjs.js",
-  "module": "dist/index.esm.js",
+  "module": "dist/index.js",
+  "sideEffects": false,
   "scripts": {
-    "prepare": "rollup --config",
+    "prepare": "babel src/ --out-dir dist --ignore spec.js,story.js; rollup --config",
```
BeniCheni: Wonder why you prefer `;` over `&&`. Read a bit on topic in a somewhat [older stackexchange post](https://unix.stackexchange.com/questions/37069/what-is-the-difference-between-and-when-chaining-commands). Would `&&` work also? If for `;` what returned value from first `babel` would the second `rollup` command depend on?

bertya: I actually don't think it really matters here. I'm open to either.

BeniCheni: @bertya, think for consistency to existing chained commands in npm-run scripts, and the fact that there'd be a subtil diff in the `;` and `&&` in the unix implication, I'd prefer to align to use `&&. Proposed the edit.

```suggestion
    "prepare": "babel src/ --out-dir dist --ignore spec.js,story.js && rollup --config",
```

BeniCheni:
Confirmed w. @bertya that this PR is outdated to be closed.

Issue #481, opened 2019-04-19
lightOrange backwards compatibility is poor


jes708:
![Screen Shot 2019-04-19 at 12 40 59 PM](https://user-images.githubusercontent.com/16601510/56433882-7e042100-62a0-11e9-8bcb-ca24aad33a5b.png)
lightOrange should alias as lightYellow for improved backwards compatibility with DS v1.  lightOrange currently aliases as orange, which could seriously impact readability and accessibility.

jes708:
Resolved by #483

PR #483, opened 2019-04-19
fix(light-orange): restore lightOrange as fef2e7

jes708:
Screen Shot 2019-04-19 at 12 40 59 PM
This is a bugfix for #481 and #485 .  lightOrange should not have been aliased as orange, as this could have a serious impact on accessibility and readability. We have decided to bring lightOrange back with a new value of #FEF2E7.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2526,7 +2526,7 @@
     },
     "@storybook/podda": {
       "version": "1.2.3",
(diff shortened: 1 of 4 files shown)


jes708:
Fixes #481

sdalonzo:
LGTM

PR #484, opened 2019-04-22
fix modal's main and module in package.json

Remexllee:
made a change in modal's package.json, change `main` and add `module`

### DIFF
--- a/packages/modal/package.json
+++ b/packages/modal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-modal",
-  "version": "2.0.1",
+  "version": "2.0.0",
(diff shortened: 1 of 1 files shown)


packages/modal/package.json
```diff
@@ -1,8 +1,9 @@
 {
   "name": "pcln-modal",
-  "version": "2.0.0",
+  "version": "2.0.1",
```
jes708: We probably don't need to upgrade the version number since lerna should do that automatically.  Otherwise this PR LGTM.

jes708:
```suggestion
    "version": "2.0.0",
```

Issue #495, opened 2019-05-16
FormField inputs break when value prop is provided


craigpalermo:
# Demo
![Screen-Recording-2019-05-16-at-9 49 41-AM](https://user-images.githubusercontent.com/3260642/57859086-42d21080-77c0-11e9-969a-b007bb0994ed.gif)

# Expected behavior
Input text should be vertically centered regardless of whether or not the `value` prop is used. This bug only affects v2.

unscsprt:
It looks like the fix in #497 no longer works, so the behavior described in the original issue still persists.

craigpalermo:
Confirming that this issue is no longer present in `3.2.0`.

PR #497, opened 2019-05-17
Fix padding around input when FormField doesn't have label

unscsprt:
Addresses #495

### DIFF
--- a/packages/core/src/FormField.js
+++ b/packages/core/src/FormField.js
@@ -9,8 +9,10 @@ const FormField = props => {
   const children = React.Children.toArray(props.children)
   const [field] = children.filter(child => child.type.isField)
   const [label] = children.filter(child => child.type.isLabel)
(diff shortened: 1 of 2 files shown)


packages/core/src/FormField.js
```diff
@@ -9,8 +9,10 @@ const FormField = props => {
   const children = React.Children.toArray(props.children)
   const [field] = children.filter(child => child.type.isField)
   const [label] = children.filter(child => child.type.isLabel)
+  const valueNoLabel = !label && field && !!field.props.value
   const showLabel =
-    (label && !label.props.autoHide) || (field && !!field.props.value)
+    ((label && !label.props.autoHide) || (field && !!field.props.value)) &&
+    !valueNoLabel
```
BeniCheni: Side thought: when these imperative object key conditions grows, it seems time to explore utility function or library to make code declarative, (e.g. Ramda). Maybe worth to log an issue.

PR #509, opened 2019-05-30
Svg accessibility

adhagam:
**Acceptance Criteria**

- The host app must be able to pass in the title and/or the desc via parameters.
- If the host app wants the SVG to be hidden as a "decorative" SVG, then it must not pass in a title .
- The <title> and elements must have ids that are unique in any given document.
 - If the title and desc params are not empty, then the SVG should have properties pointing to the <title> and elements.

- For example, aria-labelledby="uniqueTitleID uniqueDescID"

 - If the title param is empty, then the SVG must have aria-hidden="true".
 
- If the title param is not empty, then the SVG must have aria-hidden="false".

 - An SVG must have attribute focussable="false" by default.

- An optional parameter should allow the host app to set focusable=true


adhagam:
> LGTM

Thanks Steve

packages/icons/plugins/svg-dynamic-title-plugin.js
```diff
@@ -0,0 +1,45 @@
...
+      if (
+        !elements.some(element =>
+          path.get('openingElement.name').isJSXIdentifier({ name: element })
+        )
+      ) {
+        return
+      }
+
+      //Remove the Default <title> and <desc> tags if exists
+      path.get('children').forEach(childPath => {
```
BeniCheni: Curious. Would [`React.Children.forEach`](https://reactjs.org/docs/react-api.html#reactchildrenforeach) do the same logic as `.forEach`?

We can move forward for better a11y, but prefer to adapt React Top-Level API `React.Children.forEach` if same logic. I'll also mention this thought in the upcoming issue.

packages/icons/plugins/svg-dynamic-title-plugin.js
```diff
@@ -0,0 +1,45 @@
...
+    JSXElement(path) {
+      if (
+        !elements.some(element =>
+          path.get('openingElement.name').isJSXIdentifier({ name: element })
+        )
+      ) {
+        return
+      }
+
+      //Remove the Default <title> and <desc> tags if exists
```
BeniCheni: Keep comment spacing consistent, such as L1.

```suggestion
      // Remove the Default <title> and <desc> tags if exists
```

packages/icons/plugins/svg-replace-plugin.js
```diff
@@ -0,0 +1,26 @@
+//Replace <svg> tag with <Svg>
+const replaceSvgPlugin = ({ types: t }) => {
+  const replaceSvg = {
+    JSXElement(path) {
+      if (!path.get('openingElement.name').isJSXIdentifier({ name: 'svg' })) {
+        return
+      }
+      const openingElementName = path.get('openingElement.name')
```
BeniCheni: Also logging in upcoming issue:

Will the path always get `openingElement.name`? If not, maybe a conditional check to be safe w. `.replaceWith` 👇?

packages/icons/src/Icon.js
```diff
@@ -1,15 +1,50 @@
...
+const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
   const Component = icons[name] || icons[upperFirst(name)]
   if (!Component) return false
-  return <Component {...props} />
+
+  /**
+   *  To support older browsers, make sure 'titleId' is passed along with 'title' props and
+   * 'descid' with 'desc' props
+   * Also, <desc> element should be followed by a <title> tag for <svg> elements
+   * */
```
BeniCheni: Consistent spacing.

```suggestion
   **/
```

packages/icons/src/Icon.js
```diff
@@ -27,7 +62,35 @@ Icon.propTypes = {
...
+        `'desc' prop supplied to '${componentName}' should be a string`
+      )
+    } else if (!!props[propName] && !props['title']) {
+      return new Error(
+        `'title' prop should be passed along with 'desc' prop to '${componentName}'`
+      )
+    } else if (!!props[propName] && !props['descId']) {
+      return new Error(
+        `'descId' prop should be passed along with 'desc' prop to '${componentName}'`
+      )
```
BeniCheni: 👍 for thinking through the these explicitly adequate errors for DX ☝️ 🙏

packages/icons/test/index.js
```diff
@@ -30,6 +31,94 @@ describe('Icon', () => {
     expect(json).toBe(null)
   })
 
+  describe('SVG Icon Accessibility', () => {
+    // Test individual icons in './lib'
+    test.each(iconList)(
+      'renders %s with accessibility attributes when title and desc is passed ',
```
BeniCheni: Consistent spacing.

```suggestion
      'renders %s with accessibility attributes when title and desc is passed',
```

PR #517, opened 2019-06-10
Fix ScrollLock bug that incorrectly positioned body

### LINKED ISSUE #516: ScrollLock sets incorrect body top when activated
**Description**
If `window.scrollY > 0`, `ScrollLock` sets `<body>`'s `top` to a positive value when it really should be negative.

Screen Shot 2019-06-10 at 10 53 07 AM

unscsprt:
Fixes #516. I also added scrollable content to the story for `ScrollLock` to make breaking changes to it more obvious in the future.

### DIFF
--- a/packages/modal/package.json
+++ b/packages/modal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-modal",
-  "version": "2.0.2",
+  "version": "2.0.3",
   "description": "React component for modal",
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
--- a/packages/modal/package.json
+++ b/packages/modal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-modal",
-  "version": "2.0.3",
+  "version": "2.0.2",
   "description": "React component for modal",
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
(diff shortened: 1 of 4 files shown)


packages/modal/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pcln-modal",
-  "version": "2.0.2",
+  "version": "2.0.3",
```
BeniCheni: @unscsprt,

Would you mind reverting the version change in package.json of modal package? The current publish workflow stamps version in master by Collaborator. (Thought behind it is that there could be other concurrent PRs that can be published in the to-be version)

packages/modal/storybook/Modal.js
```diff
@@ -59,21 +59,35 @@ storiesOf('Modal', module)
...
-          title="Modal title"
-          onClose={() => {
-            alert('This shuold handle close')
-          }}
+  .add('With ModalHeader (and ScrollLock!)', () => {
+    // Generate content to demonstrate a scrollable <body>
+    const contentLines = []
+    for (let i = 0; i < 100; i++) {
+      contentLines.push(i)
+    }
```
BeniCheni: If you need an array with 100 slots to iterate but `.map` at L72, seems that a new array with 100 empty elements, without the `for` loop to fill in the integers for all elements in array.

```js
const contentLines = new Array(100)
```

unscsprt: I actually tried that at first, but mapping over it didn't do anything. This works, though.

```js
[...Array(100).keys()]
```

PR #519, opened 2019-06-20
feat(sr-only-v1): add SrOnly for v1

### LINKED ISSUE #518: Add SrOnly to DS v1
Since `<SrOnly>` is a core component, it may make sense to add it to DS v1 so that we can continue to support backward compatibility for the time being.

jes708:
Fixes #518


src/SrOnly.js
```diff
@@ -0,0 +1,16 @@
+import styled from 'styled-components'
+
+const SrOnly = styled.span`
```
BeniCheni: Is there another use case where the screen only composition needs something other than `span`? Confirmed the [v2 version](https://github.com/pricelinelabs/design-system/blob/master/packages/core/src/SrOnly.js#L3) is also just using `styled.span`. So it’s definitely good to move forward.

jes708: This seems sufficient for the time being, especially since this is a hidden component.  If we come across a use case for other composition in the future `.withComponent` or the `as` prop may be enough to handle such a special case.

PR #524, opened 2019-06-27
Feature/no icon tab

jes708:
Add tabIndex='-1' as a defaultProp to our icons to improve accessibility for IE and Edge


packages/icons/package.json
```diff
@@ -21,6 +21,7 @@
   "devDependencies": {
     "@babel/cli": "^7.2.3",
     "@svgr/cli": "^4.1.0",
+    "jest": "^24.1.0",
```
sdalonzo: not sure about this one

packages/icons/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "lib/Icon.js",
   "scripts": {
     "prepare": "npm run clean && npm run build:svg && npm run build:index && npm run build:copy && npm run build:babel",
-    "clean": "rm -rf node_modules components lib",
+    "clean": "rm -rf components lib",
```
sdalonzo: ?

packages/icons/package.json
jes708: I wasn't able to run the tests until I added this, since jest wasn't being installed

packages/icons/package.json
jes708: So `npm i` runs `prepare`.  If you remove the node_modules during `prepare` then it won't be able to run `svgr` or `babel` since there is no node_modules folder.  This is what I found while running node 8.15.0 and npm 6.4.1

jes708: Happy to revert `package.json` if it's some sort of node or npm version issue, but curious if anyone else has seen these issues when working on pcln-icons

BeniCheni: Does `npm run prepare` run the `clean` npm script? The workflow I've been using is to run `npm run clean` to nuke all the `node_modules` and related dist folders, then run either `npm i` (with runs postinstall script to hook into `prepare` script, or run `npm run prepare`)

I think it'd be safer to revert this package.json change for pcln-icons to move forward with this PR, and have a follow-up convo / PR to adjust any nuance about pcln-icons build / publishing workflow.

BeniCheni: Does `npm run prepare` run the `clean` npm script? The workflow I've been using is to run `npm run clean` to nuke all the `node_modules` and related dist folders, then run either `npm i` (with runs postinstall script to hook into `prepare` script, or run `npm run prepare`)

I think it'd be safer to revert this package.json change for pcln-icons to move forward with this PR, and have a follow-up convo / PR to adjust any nuance about pcln-icons build / publishing workflow.

packages/icons/package.json
BeniCheni: I just did a local test. I am able to do `npm ci && npm t` and confirmed the pcln-icons tests being executed. Definitely don't need to add jest in pcln-icon to have test working.

```
→ npm t

> @ test /Users/bchen/github/design-system
> jest

 RUNS  packages/icons/test/index.js
```

packages/icons/package.json
sdalonzo: i think the disconnect is that the tests are meant to be run from the root of the monorepo, not from inside the package

jes708: I see. Yes, I had been running the tests from the icons folder

packages/icons/package.json
jes708: UPDATED . Thanks folks!

packages/icons/package.json
jes708: UPDATED . Thanks folks!

PR #528, opened 2019-07-03
Adding Theming Support

achoboter:
This is a WIP of adding theming support that I demoed in the working group. Semantic names and implementation subject to change.

-added a palette to theme, palette would be used in place of direct colors
-added a createTheme function to add a componentStyles object to the theme
-added variants to Button and replaced some styles with componentStyles
-refactored GreenButton into a Button with variant secondary
-added variants to Link and replaced some styles with componentStyles


achoboter:
Pushed some new changes including:

-added new pages Palette and Theming to docs
-updated Button and Link docs with variant examples
-added deepmerge as a dependency to core
-refactored createTheme and added createPalette, createTextStyles and createColorStyles
-refactored getVariantStyle into applyVariant, a utility function to apply the styles for the selected variant and to merge any styles passed in through theme.componentStyles
-refactored OutlineButton and RedButton
-updated storybook for Button, Colors and Link
-added sourceType to babel config (this is needed to avoid an transpilation error with deepmerge)

*Note:* I'm not sure why all the `package-lock.json` files for the packages updated, I ran `npx lerna add deepmerge packages/core` to install the new dependency and it seems to have updated the lock for all packages.

Palette doc:
docs_palette

Theming doc:
docs_Theming

Updated Button and Link docs with variants:
docs_Button
docs_Link

packages/core/src/theme.js
```diff
@@ -312,9 +163,7 @@ const theme = {
   letterSpacings,
   regular,
   bold,
-  textStyles,
   colors,
-  colorStyles,
```
mjbyrnes4664: Ensure removing these would not be a breaking change to anyone using the design system

packages/core/src/GreenButton.js
```diff
@@ -1,19 +1,7 @@
...
-  &:hover {
-    background-color: ${props =>
-      props.disabled ? null : props.theme.colors.darkGreen};
-  }
-`
-
-GreenButton.defaultProps = {
-  theme: theme
-}
+const GreenButton = props => <Button variant="secondary" {...props} />
```
mjbyrnes4664: GreenButton should be deprecated in favor of Button with a variant prop. Add a deprecation warning

achoboter:
All components are updated to support theming.

Migration guide for updating components:
localhost_3030_Migration

Final palette:
localhost_3030_palette (1)

Theming docs:
localhost_3030_Theming

packages/core/src/Truncate.js
```diff
@@ -1,15 +1,11 @@
...
 import Text from './Text'
-import theme from './theme'
 
 const Truncate = styled(Text)`
   overflow: hidden;
   white-space: nowrap;
   text-overflow: ellipsis;
 `
-Truncate.defaultProps = {
-  theme: theme
```
jes708: nonblocking: The reason most of these core components contain `theme` is because projects with unit tests would often have several failed tests due to undefined theme, and in some cases it was very difficult to get the theme into the test. I expect this problem to come up again if theme is removed, but it may be solvable in host apps by [mounting with theme](https://github.com/styled-components/jest-styled-components/issues/106#issuecomment-373151391)

packages/core/src/createTheme.js
```diff
@@ -0,0 +1,340 @@
...
+        base: darkBlue,
+        dark: '#002f6b'
+      },
+      promoPrimary: {
+        light: lightPurple,
+        base: purple,
+        dark: '#550080'
+      },
+      promoSecondary: {
+        light: '#fe649a',
```
jes708: Should these hex values have semantic color names for improved readability?

docs/pages/Migration.md
```diff
@@ -0,0 +1,391 @@
+# Migrating to Theming
```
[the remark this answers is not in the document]
achoboter: I do agree the naming can be improved, and I it does feel like it just belongs in `Theming.md`, but I worry that there's too much content to combine them.

docs/pages/Migration.md
```diff
@@ -0,0 +1,391 @@
...
+<Box color="primary" p={2} mb={1}>Primary box</Box>
+<Box color="secondary" p={2} mb={1}>Secondary box</Box>
+<Box color="error" p={2} mb={1}>Error box</Box>
+<Box color="caution" p={2}>Caution box</Box>
+```
+
+## Button
+
+- Updated to support [palette](/palette) colors with the `color` prop.
+- Updated to support two variations, `fill` (default) and `outline`.
```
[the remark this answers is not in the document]
achoboter: Will do!

docs/pages/Migration.md
```diff
@@ -0,0 +1,391 @@
...
+- Updated to support [palette](/palette) colors with the `color` prop.
+
+```.jsx
+<Checkbox color="secondary" />
+<Checkbox color="secondary" checked />
+<Checkbox color="secondary" disabled />
+<Checkbox color="secondary" disabled checked />
+```
+
+## CloseButton (WIP)
```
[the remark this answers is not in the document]
achoboter: `CloseButton` uses `Icon`, so we will have to update `pcln-icons` to support theming first.

packages/core/src/Truncate.js
achoboter: Definitely understand why it was added as a default prop, but unfortunately this change is must, otherwise we wouldn't see the correct theme always propagate (i.e. in class components that are not wrapped in `withTheme()`.

packages/core/src/createTheme.js
achoboter: Eventually all of the old `colors.[name]` will be removed and we will only have this object with hex values.

packages/core/src/ThemeProvider.js
```diff
@@ -13,23 +13,56 @@ export const Base = styled.div`
...
 
   return (
-    <StyledThemeProvider theme={theme}>
+    <StyledThemeProvider theme={mergedTheme}>
       <Base {...props} />
     </StyledThemeProvider>
   )
 }
 
+// ThemeProvider.defaultProps = {
```
achoboter: remove default props

packages/core/src/GreenButton.js
```diff
@@ -1,19 +1,7 @@
...
-  &:hover {
-    background-color: ${props =>
-      props.disabled ? null : props.theme.colors.darkGreen};
-  }
-`
-
-GreenButton.defaultProps = {
-  theme: theme
-}
+const GreenButton = props => <Button color="secondary" {...props} />
```
James300: Should we somehow add a depreciated warning on this and similar components?

packages/core/src/IconButton.js
```diff
@@ -26,10 +25,6 @@ const IconButton = ({ name, size, color, ...props }) => (
 )
 
 IconButton.displayName = 'IconButton'
-
-IconButton.defaultProps = {
-  theme: theme
-}
 IconButton.isIconButton = true
```
James300: What is this property used for and can we remove it, just seems unnecessary or potentially hiding a problem.

packages/core/package.json
```diff
@@ -50,6 +50,7 @@
     ]
   },
   "dependencies": {
+    "deepmerge": "^4.0.0",
```
achoboter: change to peerDependency

packages/core/src/IconButton.js
achoboter: this property is used in `IconField` to check the type of component (icon or button) that is passed in.

packages/core/package.json
achoboter: after discussion, will leave this for now (do not want to force every app to install deepmerge)

PR #534, opened 2019-07-12
remove outline from icons

gbellini90:


sdalonzo:
an addition to the unit test would be good to verify the stlye change - and did you run the tests at all? i'd have expected a snapshot diff

Issue #533, opened 2019-07-12
Badge component should be more configurable


unscsprt:
# Use case:
We'd like to be able to configure `Badge` to look similar to `Stamp`.

## Example:
<img width="127" alt="Screen Shot 2019-07-12 at 8 56 08 AM" src="https://user-images.githubusercontent.com/3260642/61129695-f256f680-a482-11e9-8c77-42c8fa7391f1.png">

# Proposed props:
- `borderRadius` (string or bool) - Either standard CSS `border-radius` rule, or a `bool` to use a Stamp-like border-radius.
- `fontSize` (string) - Same typography scale as used for `<Text>`

dansternfeld1:
Discussion in the room seems to favor changing Stamp to allow setting of border-color for this use case

Other thoughts?

PR #535, opened 2019-07-12
Add borderColor prop to Stamp

unscsprt:
Addresses #533

### DIFF
--- a/packages/core/src/Stamp.js
+++ b/packages/core/src/Stamp.js
@@ -14,8 +14,7 @@ const Stamp = styled.div`
   border-radius: 2px;
   border-width: 1px;
   border-style: solid;
-  border-color: ${props =>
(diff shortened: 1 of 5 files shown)


packages/core/src/Stamp.js
```diff
@@ -14,7 +14,8 @@ const Stamp = styled.div`
   border-radius: 2px;
   border-width: 1px;
   border-style: solid;
-  border-color: ${props => theme.colors.borderGray};
+  border-color: ${props =>
+    theme.colors[props.borderColor] || theme.colors.borderGray};
```
sdalonzo: imho a default prop would be cleaner here

sdalonzo: update address this. thanks @unscsprt 👍

PR #539, opened 2019-07-16
Fix remove outline from icons 2

gbellini90:
Conditionally rendering outline for icons based off of props: focusable and tabIndex.


sdalonzo:
something does not look right here. how old was your master checkout when you started?

PR #540, opened 2019-07-16
conditional outline style for icons based on focusable and tabIndex p…

gbellini90:
conditionally rendering an outline style for icons based on the props focusable and tabIndex
renders an outline when focusable is true and tabIndex is greater than or equal 0
renders no outline when focusable is false and tabIndex is less  than or equal 0

### DIFF
--- a/packages/core/storybook/Icon.js
+++ b/packages/core/storybook/Icon.js
@@ -25,3 +25,33 @@ storiesOf('Icon', module)
       <Icon color="orange" size={48} m={2} name="Cars" />
     </div>
   ))
+  .add(
+    'Renders Icon with outline when focusable and tabIndex greater than or equal to 0',
+    () => (
+      <Box p={2} color="white" bg="blue">
+        <Flex wrap>
+          {keys.map(name => (
+            <Box key={name} width={[1 / 3, 1 / 5, 1 / 6, 1 / 8]} mx={2} my={3}>
+              <Icon name={name} focusable="true" tabIndex="1" size={48} />
+              <Truncate fontSize={0}>{name}</Truncate>
+            </Box>
+          ))}
+        </Flex>
+      </Box>
+    )
+  )
+  .add(
+    'Renders Icon with no outline when not focusable and tabIndex less than 0',
+    () => (
+      <Box p={2} color="white" bg="blue">
+        <Flex wrap>
+          {keys.map(name => (
+            <Box key={name} width={[1 / 3, 1 / 5, 1 / 6, 1 / 8]} mx={2} my={3}>
+              <Icon name={name} focusable="false" tabIndex="-1" size={48} />
+              <Truncate fontSize={0}>{name}</Truncate>
+            </Box>
+          ))}
+        </Flex>
+      </Box>
+    )
+  )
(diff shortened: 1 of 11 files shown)


sdalonzo:
took a look at these logs - the issue is unit test coverage.
`Jest: "global" coverage threshold for branches (90%) not met: 78.87%
Jest: "global" coverage threshold for functions (90%) not met: 85.95%`

packages/icons/svgr.config.js
```diff
@@ -20,6 +21,19 @@ export const ${componentName} = ({
...
+const showOutline = ({ focusable, tabIndex }) => {
+  if (focusable === 'true' && parseInt(tabIndex, 10) >= 0){
+    return ''
+  }
+  else {
+    return 'outline: none;'
+  }
+}
+
+export const ${componentName} = styled(BaseComponent)\`
```
jes708: 2 Questions:

1. Why are non-alphanumeric characters being escaped here?
2. Should this code also be added to packages/icons/src/Icon.js?

sdalonzo: 1) it broke on nested backticks, this was the only thing that worked - since we are inside the template backticks that start on line 9 https://github.com/pricelinelabs/design-system/blob/Fix-outline-property-for-icons/packages/icons/svgr.config.js#L9

2) Icon uses these svgr-built components, so it should be there - https://github.com/pricelinelabs/design-system/blob/Fix-outline-property-for-icons/packages/icons/src/Icon.js#L8

sdalonzo: it's ugly - if you have a better way of rendering backticks and `${}` inside of the template, without evaluation, please suggest.

jes708: Ok, this makes sense.  I approve!

packages/core/storybook/Icon.js
```diff
@@ -25,3 +25,33 @@ storiesOf('Icon', module)
...
+              <Icon name={name} focusable="true" tabIndex="1" size={48} />
+              <Truncate fontSize={0}>{name}</Truncate>
+            </Box>
+          ))}
+        </Flex>
+      </Box>
+    )
+  )
+  .add(
+    'Renders Icon with no outline when not focusable and tabIndex less than 0',
```
dansternfeld1: 👍

packages/icons/test/index.js
```diff
@@ -119,6 +119,24 @@ describe('Icon', () => {
...
+
+    test('Outline when focusable', () => {
+      const namedJson = TestRenderer.create(
+        <AirplaneIcon focusable="true" tabIndex="1" />
+      ).toJSON()
+      const json = TestRenderer.create(
+        <Icon focusable="true" name="Airplane" tabIndex="1" />
+      ).toJSON()
+      expect(json).toEqual(namedJson)
+      expect(json).not.toHaveStyleRule('outline', 'none')
```
dansternfeld1: 👍 clever

dansternfeld1:
Yup, agreed with @sdalonzo -- it's the coverage level that's causing travis to fail:
https://travis-ci.org/pricelinelabs/design-system/builds/559543172#L1101

@gbellini90 please add the required unit test coverage

BeniCheni:
Ping @gbellini90 for requested test coverage and passing “_green_” Travis CI.

dansternfeld1:
I'm closing this as a dupe of #599

PR #547, opened 2019-08-01
Allow Icon Button in Form Field

huzefakagdi:
Currently, only an Icon is allowed inside a FormField. With these changes, we can also enable the use of an IconButton inside the FormField. 
The IconButton currently gets an outline when it is in focus in the FormField. If the changes from the other PR [https://github.com/pricelinelabs/design-system/pull/540](url) are merged, we will have an option to remove the outline form the IconButton in my understanding.

### DIFF
--- a/packages/core/src/IconField.js
+++ b/packages/core/src/IconField.js
@@ -2,18 +2,20 @@ import React from 'react'
 import Flex from './Flex'
 
 const IconField = props => {
+  const isIcon = item => item.type.isIcon || item.type.isIconButton
+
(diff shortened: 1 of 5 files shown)


packages/core/src/IconField.js
```diff
@@ -2,18 +2,20 @@ import React from 'react'
 import Flex from './Flex'
 
 const IconField = props => {
+  const isIcon = item => item.type.isIcon || item.type.isIconButton
```
BeniCheni: +1 for generalizing a function.

PR #548, opened 2019-08-08
feat: add additional props to pcln modal for custom animations

dgiraldo313:
With all the a11y work, there will be some instances of the modal that require different animations and alignment on the page in order to match existing design. This PR adds new props to allow consumers to overwrite animations for both the overlay and the dialog, as well as change the vertical alignment of the dialog on the page.

### DIFF
--- a/packages/modal/README.md
+++ b/packages/modal/README.md
@@ -15,7 +15,7 @@ import Modal from 'pcln-modal'
 ```jsx
 import { Modal } from 'pcln-modal'
 
-;<Modal
+<Modal
   isOpen={true} //boolean for control this status of modal
   onClose={someFunc} //func function for handle close the modal while click on the overlay
   bg="white" //modal background color
--- a/packages/modal/README.md
+++ b/packages/modal/README.md
@@ -70,12 +70,12 @@ For its animations, this Modal currently uses `react-transition-group`. This mea
 We can then use these states to write custom animations, like so:
 
 ```javascript
-const MY_ANIMATION = transitionstate => `
+const MY_ANIMATION = transitionState => `
   transform: scale(0.5);
   transition: transform .5s cubic-bezier(0.50, 0.00, 0.25, 1.00);
-  ${transitionstate === 'entering' ? `transform: scale(0.5);` : ''}
-  ${transitionstate === 'entered' ? `transform: scale(1);` : ''}
-  ${transitionstate === 'exiting' ? `transform: scale(0.5);` : ''}
-  ${transitionstate === 'exited' ? `transform: scale(0.5);` : ''}
+  ${transitionState === 'entering' ? `transform: scale(0.5);` : ''}
+  ${transitionState === 'entered' ? `transform: scale(1);` : ''}
+  ${transitionState === 'exiting' ? `transform: scale(0.5);` : ''}
+  ${transitionState === 'exited' ? `transform: scale(0.5);` : ''}
 `
 ```
--- a/packages/modal/README.md
+++ b/packages/modal/README.md
@@ -15,7 +15,7 @@ import Modal from 'pcln-modal'
 ```jsx
 import { Modal } from 'pcln-modal'
 
-<Modal
+;<Modal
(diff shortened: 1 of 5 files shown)


packages/modal/src/Modal.js
```diff
@@ -6,6 +6,35 @@ import { color, width, height } from 'styled-system'
 import { DialogOverlay, DialogContent } from '@reach/dialog'
 import { Box, CloseButton, Flex } from 'pcln-design-system'
 
+const OVERLAY_ANIMATION = transitionstate => `
```
sdalonzo: love this refactor

packages/modal/README.md
```diff
@@ -15,7 +15,7 @@ import Modal from 'pcln-modal'
 ```jsx
 import { Modal } from 'pcln-modal'
 
-<Modal
+;<Modal
```
BeniCheni: Remove the extra `;`?

```suggestion
<Modal
```

packages/modal/src/Modal.js
```diff
@@ -6,6 +6,35 @@ import { color, width, height } from 'styled-system'
...
+  ${transitionstate === 'exiting' ? `transform: scale(0.5);` : ''}
+  ${transitionstate === 'exited' ? `transform: scale(0.5);` : ''}
+`
+
+const getAnimation = ({
+  transitionstate,
+  defaultAnimation,
+  customAnimation = null
+}) => {
+  if (customAnimation && typeof customAnimation === 'function') {
```
BeniCheni: `typeof` check covered nullity check already. Condition could be simplified such as 👇 

```suggestion
  if (typeof customAnimation === 'function') {
```

packages/modal/src/Modal.js
```diff
@@ -6,6 +6,35 @@ import { color, width, height } from 'styled-system'
...
+    return customAnimation(transitionstate)
+  }
+  return defaultAnimation(transitionstate)
+}
```
BeniCheni: Could be simplified due to `typeof` handles nullity check already, and a `return` statement with a ternary expression will omit the `if` imperative block. Also camel case can be consistent for param `transitionState`:

```js
const getAnimation = ({
  transitionState,
  defaultAnimation,
  customAnimation = null
}) => typeof customAnimation === 'function'
  ? customAnimation(transitionstate)
  : defaultAnimation(transitionstate)
```

packages/modal/storybook/Modal.js
```diff
@@ -7,6 +7,11 @@ import styled from 'styled-components'
 const StyledModal = styled(Modal)`
   height: ${props => props.height};
 `
+const CUSTOM_ANIMATION = transitionstate => `
```
BeniCheni: Consistent camel case for variable name?

```suggestion
const CUSTOM_ANIMATION = transitionState => `
```

packages/modal/storybook/Modal.js
```diff
@@ -7,6 +7,11 @@ import styled from 'styled-components'
...
+  transform: translateY(-122%);
+  transition: transform 0.3s linear;
+  ${transitionstate === 'entered' ? `transform: translateY(0%);` : ''}
```
BeniCheni: Consistent camel case for variable name?

```suggestion
  ${transitionState === 'entered' ? `transform: translateY(0%);` : ''}
```

packages/modal/test/Modal.js
```diff
@@ -74,4 +74,34 @@ describe('Modal', () => {
...
+  test('renders with top alignment', () => {
+    const { getByText, getByTestId, container } = rerender(
+      <ThemeProvider>
+        <Modal isOpen verticalAlignment="top">
+          <div data-content>Content</div>
+        </Modal>
+      </ThemeProvider>
+    )
+    expect(getByText('Content')).toBeTruthy()
+    expect(getByTestId('pcln-modal-close')).toBeTruthy()
```
BeniCheni: +1 on tests using testing library / react. 👍

Issue #471, opened 2019-04-08
Add Modal to Docs Site


jes708:
The Modal component should be showcased on our docs site

https://pricelinelabs.github.io/design-system/

BeniCheni:
Nice suggestion, @jes708. Tagged with "help wanted" and "good first issue" labels.

unscsprt:
@jes708 Thanks for raising this issue. I'm going to address it shortly.

James300:
Issue is fixed

James300:
![](https://github.trello.services/images/mini-trello-icon.png) [Add Modal to Docs Site](https://trello.com/c/6sx4GMfy/27-add-modal-to-docs-site)

PR #551, opened 2019-08-09
Add Modal to docs site

unscsprt:
Addresses #471

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -12,6 +12,7 @@
     "is-absolute-url": "^2.1.0",
     "mdx-docs": "^1.0.0-9",
     "pcln-design-system": "^2.0.5",
+    "pcln-modal": "^2.0.4",
     "pcln-slider": "2.0.1",
     "prop-types": "^15.6.2",
     "react": "^16.4.1",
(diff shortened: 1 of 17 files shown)


docs/package.json
```diff
@@ -12,6 +12,7 @@
     "is-absolute-url": "^2.1.0",
     "mdx-docs": "^1.0.0-9",
     "pcln-design-system": "^2.0.5",
```
sdalonzo: n-b bump to latest?

jes708:
Might it make sense to rename the Components section as Core Components, give the modal its own Modal Components section, and then provide docs for each export of pcln-modal (headers, etc)?

dgiraldo313:
> Might it make sense to rename the Components section of the docs as Core Components, give the modal its own Modal Components section, and then provide docs for each export of pcln-modal (headers, etc)?

I think that's a great idea. I've seen confusion from some people thinking that the modal, popover and other standalone packages come with the DS core. I think separating these in the docs can help alleviate some of this confusion.

unscsprt:
@dgiraldo313 @jes708 Good idea; created separate section for Modal components.

Screen Shot 2019-08-09 at 12 17 29 PM

PR #554, opened 2019-08-12
Creation of a PlaceholderImage component to be used during development

tfquirk:
Added PlaceholderImage to packages/src, as well as tests, and storybook. Doc for PlaceholderImage also created


tfquirk:
Packages version will need to be bumped.

packages/core/src/PlaceholderImage.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+
+import Image from './Image'
+import theme from './theme'
+
+const RandomImage = styled(Image)`
+  height: ${props => props.height}px;
```
sdalonzo: any reason not to leave the units arbitrary? since they are strings anyway, the user could specify percents, ems, vh/vw, etc

tfquirk: @sdalonzo, All of the urls we decided to use expect the `height` and `width` attributes to be numbers alone. As written using one standard measurement will allow this to work. 

Certainly could be more creative to get around this if you want the ability to use other measurements, have separate props for number and a prop to specify px/ems/percent, etc. Lmk what you think.

sdalonzo: makes sense. thanks for the explanation!

Issue #550, opened 2019-08-09
Make Checkbox and Radio Accessible


adhagam:
Make Checkbox and Radio fields Accessible when the they are focused

dansternfeld1:
@dgiraldo313 please work with @adhagam

unscsprt:
Hey @dgiraldo313, I'll start working on Checkbox. We're including that in our team's current sprint.

dgiraldo313:
@unscsprt Ok, great. Let me know if you need any help with testing and validation. Thanks!

James300:
This issue is done.

James300:
![](https://github.trello.services/images/mini-trello-icon.png) [Make Checkbox and Radio Accessible](https://trello.com/c/IJoJIugm/26-make-checkbox-and-radio-accessible)

PR #558, opened 2019-08-16
Checkbox a11y

unscsprt:
Addresses part of #550.

**Design Spec:**
Screen Shot 2019-08-16 at 12 48 24 PM

**Demo:**
ezgif com-video-to-gif


unscsprt:
@dgiraldo313 I was able to rewrite this without using any additional state variables. Here's the updated demo. Good call on the CSS-only solution.

ezgif com-video-to-gif (2)

PR #559, opened 2019-08-16
Breadcrumbs

James300:
Currently is use as part of the packages product, seemed like a good candidate for the DS.

image


packages/core/src/Breadcrumb.js
```diff
@@ -0,0 +1,43 @@
...
+  cursor: ${props => (props.onClick ? 'pointer' : 'default')};
+`
+
+function Breadcrumb({ className, isCurrent, isCompleted, children, onClick }) {
+  const color = isCurrent || isCompleted ? 'primary' : 'text.light'
+
+  return (
+    <Container alignItems="center" onClick={onClick}>
+      {isCompleted && <CheckIcon />}
+      <Text color={color} bold={isCurrent} fontSize={1}>
```
unscsprt: Do we see any utility in making the `fontSize` configurable?

packages/core/src/Breadcrumb.js
```diff
@@ -0,0 +1,43 @@
...
+        {children}
+      </Text>
+    </Container>
+  )
+}
+
+const CheckIcon = () => (
+  <Icon name="Check" color="primary" size={16} mr={1} legacy={false} />
```
unscsprt: Can we just import `Check` from `pcln-icons`?

packages/core/src/Breadcrumb.js
```diff
@@ -0,0 +1,43 @@
...
+  className: PropTypes.string,
+  isCurrent: PropTypes.bool,
+  isCompleted: PropTypes.bool,
+  children: PropTypes.node,
+  onClick: PropTypes.func
+}
+
+Breadcrumb.defaultProps = {
+  className: ''
+}
```
unscsprt: I think you can omit the default prop for `className`. Also, don't forget to pass `className` down to `Container`.

packages/core/src/index.js
```diff
@@ -4,6 +4,8 @@ export { default as Badge } from './Badge'
 export { default as Banner } from './Banner'
 export { default as BlockLink } from './BlockLink'
 export { default as Box } from './Box'
+export { default as Breadcrumb } from './Breadcrumb'
```
unscsprt: I wonder if exposing `Breadcrumb` might encourage people to go around the official `Breadcrumbs` component and try to roll their own if they have specific design needs, which probably isn't desirable.

James300: That's a good point

packages/core/src/Breadcrumb.js
James300: Yeah that's a good idea, is there a good way to scale the size of the icon too? I feel like `width/height: themeGet('fontSizes.x')` is a little weird.

BeniCheni: I think simply adding another prop fontSize with default prop value 1 would serve the configurable suggestion.

packages/core/src/Breadcrumb.js
```diff
@@ -0,0 +1,39 @@
...
+      {isCompleted && <CheckIcon />}
+      <Text color={color} bold={isCurrent} fontSize={1}>
+        {children}
+      </Text>
+    </Container>
+  )
+}
+
+const CheckIcon = () => (
+  <Icon name="check" color="primary" size={16} mr={1} legacy={false} />
```
BeniCheni: Since pcln-icons supports named icon, I think this could become CheckIcon.

packages/core/src/Breadcrumb.js
```diff
@@ -0,0 +1,39 @@
...
+)
+
+Breadcrumb.displayName = 'Breadcrumb'
+
+Breadcrumb.propTypes = {
+  className: PropTypes.string,
```
BeniCheni: Can you explain this className prop? I wouldn’t imagine CSS-in-JS would use class name.

James300: When using styled component or any CSS-in-JS library the component has to accept a className otherwise the styles will not take effect.

packages/core/src/Breadcrumb.js
James300: I can make this work if I import from `pcln-icon/lib/Check` but not from just `pcln-icons` as it's not exported to handle this.

packages/core/src/Breadcrumb.js
James300: Yes but it could result in something like:
image
where the icon size is tiny.

I could convert the fontSize to size for the icon too but is it worth it when we should probably be keeping our breadcrumb size consistent across the site???

packages/core/src/Breadcrumb.js
BeniCheni: Cool, would you mind doing it in this PR? If you prefer to keep the Pr the same scope, would you mind opening a new issue to track for next volunteer to follow up to use names icon import?

James300: Sorry I'm not sure if you want me to change the import to `pcln-icon/lib/Check` or to update the icons package to export the icons differently.

James300:
Closing to reopen as Stepper PR.

New PR: https://github.com/pricelinelabs/design-system/pull/583

Issue #544, opened 2019-07-30
pcln-modal does not receive child styles of ThemeProvider


jes708:
pcln-modal is missing `line-height`, `box-sizing`, and other properties that are specified by the child of ThemeProvider due to the nature of the modal's DOM manipulation. We should ensure that the modal includes any import styles that may be attached to a ThemeProvider's child.

BeniCheni:
#563 landed to fix this issue.

PR #562, opened 2019-08-19
fix: modal style update

bertya:
related to ticket #544

### DIFF
--- a/packages/modal/package.json
+++ b/packages/modal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-modal",
-  "version": "2.1.0",
+  "version": "2.1.1",
   "description": "React component for modal",
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
--- a/packages/modal/package.json
+++ b/packages/modal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-modal",
-  "version": "2.1.1",
+  "version": "2.1.0",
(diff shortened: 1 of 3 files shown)


packages/modal/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pcln-modal",
-  "version": "2.1.0",
+  "version": "2.1.1",
```
jes708: Version updates can wait until the publishing process, as per the [Publishing guidelines](https://github.com/pricelinelabs/design-system/blob/master/CONTRIBUTING.md#publishing)

packages/modal/src/Modal.js
```diff
@@ -43,7 +44,12 @@ const Overlay = styled(DialogOverlay)`
   ${props => `
     z-index: ${props.zindex || 100};
     font-family: ${props.theme.font};
+    line-height: ${props => props.theme.lineHeights.standard};
```
jes708: Should we be concerned about keeping this in sync with <Base /> from ThemeProvider?

https://github.com/pricelinelabs/design-system/blob/c2ad5c588944d5713bb6c61f32ce20a81368d31b/packages/core/src/ThemeProvider.js#L6-L14

bertya: personally, I feel it's fine since it's grabbing things from the theme,

jes708: Ok, that sounds fair to me

PR #563, opened 2019-08-20
Fix/#544 modal style chaining

bertya:
related to ticket #544


unscsprt:
@bertya ✅LGTM, but I don't see an approve button anywhere.

PR #567, opened 2019-08-21
Remove Warnings and Update Readme for Theming

James300:
According to a11y, icons don’t actually require title/desc unless the icon is meant to be interacted with by the user. We currently handle this correctly by setting aria-hidden to true but these prop type warnings are not consistent with the behaviour.

Additionally, I updated the readme to the new themeable approach which is simply to allow the icon to inherit from the parent component. We currently don't have a way of passing primary or any other semantic name through to the icon but I will be bringing this up as a topic of the next DS meeting.

### DIFF
--- a/packages/icons/src/Icon.js
+++ b/packages/icons/src/Icon.js
@@ -1,55 +1,39 @@
 import React from 'react'
 import upperFirst from 'lodash.upperfirst'
 import PropTypes from 'prop-types'
-
 import * as icons from './index'
 
 const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
   const Component = icons[name] || icons[upperFirst(name)]
-  if (!Component) return false
 
-  /**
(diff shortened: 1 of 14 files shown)


packages/icons/src/Icon.js
```diff
@@ -1,97 +1,52 @@
...
   name: (props, key, componentName) => {
     const name = props[key]
     if (!icons[name] && !icons[upperFirst(name)]) {
       return new Error(
         `Unknown name prop \`${name}\` supplied to \`${componentName}\``
       )
-    } else if (!icons[name]) {
-      return new Error(
-        `Icon name prop should be uppercase.\n` +
-          `Use \`${upperFirst(name)}\` instead of ${name}.`
```
James300: I ran into a case where this warning actually caused a production bug. For some reason locally using uppercase names was fine but we found that production only accepted lowercase icon names.

PR #578, opened 2019-09-03
chore(Hide): hide able to handle inline 

### LINKED ISSUE #553: Hide should have a default display of inline

donnobot:
fixes #553

### DIFF
--- a/packages/core/src/Hide.js
+++ b/packages/core/src/Hide.js
@@ -46,5 +46,6 @@ Hide.propTypes = {
 }
 
 Hide.displayName = 'Hide'
+Hide.text = Hide.withComponent('text')
 
 export default Hide
--- a/packages/core/test/Hide.js
+++ b/packages/core/test/Hide.js
@@ -36,4 +36,9 @@ describe('Hide', () => {
     const json = rendererCreateWithTheme(<Hide print />).toJSON()
     expect(json).toMatchSnapshot()
   })
+
+  test('renders as inline', () => {
+    const json = rendererCreateWithTheme(<Hide.text />).toJSON()
+    expect(json).toMatchSnapshot()
+  })
 })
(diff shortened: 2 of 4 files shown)


packages/core/test/__snapshots__/Hide.js.snap
```diff
@@ -6,6 +6,12 @@ exports[`Hide renders 1`] = `
 />
 `;
 
+exports[`Hide renders as inline 1`] = `
+<text
```
sdalonzo: hmm

packages/core/test/Hide.js
```diff
@@ -36,4 +36,9 @@ describe('Hide', () => {
     const json = rendererCreateWithTheme(<Hide print />).toJSON()
     expect(json).toMatchSnapshot()
   })
+
+  test('renders as inline', () => {
+    const json = rendererCreateWithTheme(<Hide.text />).toJSON()
+    expect(json).toMatchSnapshot()
```
sdalonzo: perhaps a `toHaveStyleRule` test to prove out that it is in fact inline?

BeniCheni:
ping / @donnobot , any update to move forward to land this PR? Just some trivial updates per @sdalonzo's comments.

packages/core/src/Hide.js
```diff
@@ -46,5 +46,6 @@ Hide.propTypes = {
 }
 
 Hide.displayName = 'Hide'
+Hide.text = Hide.withComponent('text')
```
jes708: FYI `withComponent` is a candidate for deprecation in styled-components v4.

jes708: `withComponent('text')` is trying to use an html5 text tag, which isn't valid.  Maybe this could use `withComponent('span')` instead?

PR #581, opened 2019-09-04
fix: Select arrow in IE11

747823:
The native select arrow on IE11 was covering our custom icon, this change hides the native one. Also changed the rollup command to be run with `npx`, my terminal was not able to find the rollup executable from .bin without `npx`. Using `npx` makes it more reliable.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -5,7 +5,7 @@
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
   "scripts": {
-    "prepare": "rollup --config",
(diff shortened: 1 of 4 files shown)


packages/core/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.js",
   "scripts": {
-    "prepare": "rollup --config",
+    "prepare": "npx rollup --config",
```
BeniCheni: +1, 🙏🏻 Thanks for making the DX better!

PR #583, opened 2019-09-09
Stepper

James300:
image

### DIFF
--- a/packages/slider/package-lock.json
+++ b/packages/slider/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-slider",
-  "version": "2.0.5",
+  "version": "2.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/packages/slider/package.json
+++ b/packages/slider/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-slider",
-  "version": "2.0.5",
+  "version": "2.1.0",
   "description": "React component for range sliders",
   "main": "dist/index.cjs.js",
   "module": "dist/index.esm.js",


wesleymartin85:
@James300 I think will works. Thanks!

docs/pages/Stepper.md
```diff
@@ -0,0 +1,9 @@
+# Stepper
+
+```.jsx
+<Stepper>
+    <Stepper.Step isCompleted>First Step</Stepper.Step>
+    <Stepper.Step isCurrent>Second Step</Stepper.Step>
+    <Stepper.Step>Last Step</Stepper.Step>
```
BeniCheni: nit in the doc: 2-space indentations for these

packages/core/src/Step.js
```diff
@@ -0,0 +1,37 @@
...
+  cursor: ${props => (props.onClick ? 'pointer' : 'default')};
+`
+
+function Step({ className, isCurrent, isCompleted, children, onClick }) {
+  const color = isCurrent || isCompleted ? 'primary' : 'text.light'
+
+  return (
+    <Container className={className} alignItems="center" onClick={onClick}>
+      {isCompleted && (
+        <Icon name="check" color="primary" size={16} mr={1} legacy={false} />
```
BeniCheni: Can the `CheckIcon` be imported and used instead?

```jsx
import CheckIcon from 'pcln-icons/lib/Check'
...
<CheckIcon color="primary" size={16} mr={1} />
```

packages/core/src/Step.js
```diff
@@ -0,0 +1,37 @@
...
+  )
+}
+
+Step.displayName = 'Step'
+
+Step.propTypes = {
+  className: PropTypes.string,
+  isCurrent: PropTypes.bool,
+  isCompleted: PropTypes.bool,
+  children: PropTypes.node,
```
BeniCheni: Prop type for children can use further prop type definite with `PropTypes.element`. (See [prop-types usage doc](https://github.com/facebook/prop-types#usage) to learn more.)

packages/core/src/Step.js
```diff
@@ -0,0 +1,37 @@
...
+    </Container>
+  )
+}
+
+Step.displayName = 'Step'
+
+Step.propTypes = {
+  className: PropTypes.string,
+  isCurrent: PropTypes.bool,
+  isCompleted: PropTypes.bool,
```
BeniCheni: For these boolean prop types, I think it's fair to provide default values since boolean values are either `false` or `true`. Seems perfect candidate to adapt prop type default values.

packages/core/src/Stepper.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import Flex from './Flex'
+import Step from './Step'
+
+const Container = styled(Flex)`
```
BeniCheni: I would avoid using the name "Container" as the there's indeed a DS component called `Container`. Maybe use a name that retains the primitive `Flex` such as `FlexWrapper` or something similar?

packages/core/test/Step.js
```diff
@@ -0,0 +1,58 @@
...
+    expect(json).toHaveStyleRule('cursor', 'pointer')
+
+    expect(hasIcon(component)).toBeFalsy()
+  })
+})
+
+const hasIcon = component => {
+  const { container } = renderWithTheme(component)
+  return container.querySelector('svg')
+}
```
BeniCheni: nit: seems more logical to have the named arrow function `hasIcon` declared before the calling code `hasIcon(component)` happens.

The `hasIcon` function isn't declared as a first-class function w. signature `function hasIcon(component) {`, so interesting the calling code `hasIcon(component)` would work actually.

packages/core/src/Step.js
BeniCheni: Secondly, I think these names could be `active` and `completed`. Reasons:

- adjective prop names could omit "is" prefix, as they imply boolean values
- current step in some designs may not be the active step word that needs bold style. I think `active` as a prop name seems more expressive to avoid ambiguity. (see inspiration from the [Step API of Material-UI](https://material-ui.com/api/step/#props)

P.S.: I also think `disabled ` prop is worth of consideration for the `Step` implementation, to allow designs to render a step UI that's just for visual purpose, without any functionalities. You can consider to add `disabled` prop in this PR or log an issue to follow up.

packages/core/src/Stepper.js
```diff
@@ -0,0 +1,48 @@
...
+`
+
+function Stepper({ className, currentIndex, items, children }) {
+  return (
+    <Container className={className}>
+      {children}
+      {items &&
+        items.map((item, index) => (
+          <Step
+            key={index}
```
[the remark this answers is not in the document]
James300: The order of these isn't going to change so is it a big deal?

packages/core/src/Step.js
James300: I tried this and we ran into a problem with the docs, we looked into everything that we could think of but couldn't resolve it. Using the base icon didn't cause the issue though.

packages/core/src/Step.js
James300: I prefer `node` here as the children could be a string, highly unlikely but I don't think we should through a warning if that occurs.

BeniCheni: It’s fair to render string as children without warning. Good point. This comment could be resolved. Thanks.

packages/core/test/Step.js
James300: React doesn't care which order the functions are declared in but it's a good suggestion 👍

packages/core/src/Step.js
James300: I'll add an issue as I need this released asap.

PR #584, opened 2019-09-10
Stepper

James300:
image

I messed something up and had to made another new PR. I swear this never happens...

Old PR: https://github.com/pricelinelabs/design-system/pull/583


packages/core/src/Step.js
```diff
@@ -0,0 +1,43 @@
...
+  cursor: ${props => (props.onClick ? 'pointer' : 'default')};
+`
+
+function Step({ className, active, completed, children, onClick }) {
+  const color = active || completed ? 'primary' : 'text.light'
+
+  return (
+    <StyledFlex className={className} alignItems="center" onClick={onClick}>
+      {completed && (
+        <Icon name="check" color="primary" size={16} mr={1} legacy={false} />
```
sdalonzo: could you import CheckIcon instead, please?

sdalonzo: this will force every icon into consuming bundles, even if they have removed all uses of the bare Icon

packages/core/src/Step.js
```diff
@@ -0,0 +1,43 @@
...
+  )
+}
+
+Step.displayName = 'Step'
+
+Step.propTypes = {
+  className: PropTypes.string,
+  active: PropTypes.bool,
+  completed: PropTypes.bool,
+  children: PropTypes.node,
```
sdalonzo: i feel like children should be required. thoughts?

packages/core/src/Stepper.js
```diff
@@ -0,0 +1,48 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import Flex from './Flex'
+import Step from './Step'
+
+const StyledFlex = styled(Flex)`
+  & > :not(:last-child) {
+    margin-right: 32px;
```
sdalonzo: perhaps get the spacing from the inherited theme provider `props => props.theme.spacing[x]`

packages/core/src/Stepper.js
```diff
@@ -0,0 +1,48 @@
...
+`
+
+function Stepper({ className, currentIndex, items, children }) {
+  return (
+    <StyledFlex className={className}>
+      {children}
+      {items &&
+        items.map((item, index) => (
+          <Step
+            key={index}
```
sdalonzo: array index should never be used as a key
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318 as linked from https://reactjs.org/docs/lists-and-keys.html

packages/core/storybook/Step.js
```diff
@@ -0,0 +1,23 @@
...
+import Step from '../src/Step'
+
+const onClick = () => alert('Step Clicked')
+
+storiesOf('Step', module)
+  .add(
+    'Step component',
+    withInfo({
+      inline: true,
+      text: 'Use the <Step /> component to render a step.'
```
sdalonzo: Stepper.Step?

packages/core/storybook/Step.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { withInfo } from '@storybook/addon-info'
+import Step from '../src/Step'
+
+const onClick = () => alert('Step Clicked')
```
sdalonzo: please use storybook actions for click handlers, instead of alerts

packages/core/storybook/Stepper.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { withInfo } from '@storybook/addon-info'
+import { Stepper } from '../src'
+
+const onFirstClick = () => alert('First Step Clicked')
```
sdalonzo: same here, use action please

packages/core/src/Step.js
unscsprt: Probably not a bad idea, unless we see value in supporting an icon-only stepper, but we'd want have to make more changes in that case anyway. Seems like it should be required.

packages/core/src/Step.js
James300: Yes that's a problem. However, I tried using the specific icon but that cause an issue in the docs project where it couldn't find the `css` function from styled components which is used to handle the palette colors. Adam and I looked into this thoroughly but could not resolve the issue and said that this is a good enough approach for now and that I should add a tech debt issue once this PR is merged to look into the issue further.

packages/core/src/Step.js
James300: I think that's a fair requirement 👍

packages/core/src/Stepper.js
James300: I don't love using `index` but since the steps don't have an id and the name doesn't have to be unique (although they really should be) index all we really have and it seems fine to me in this scenario since the order of the items in a stepper will never change.

Quote from the react docs: "We don’t recommend using indexes for keys if the order of items may change."

However, all this being said if you can think of an alternative way then I will happily change it because while using index is okay in this scenario it might lead to others using it when it is not acceptable.

packages/core/storybook/Step.js
James300: That's fair, I should keep it consistent 👍

packages/core/src/Stepper.js
sdalonzo: if ever a step were to get introduce between existing steps, react would go haywire. perhaps requiring a step id is the sanest approach

packages/core/src/Step.js
sdalonzo: i suppose that would be ok if there is a commitment to fast follow on it. but i'm really hesitant to allow a single component to force all consumers of DS to bundle every icon for an issue you say is limited to the current docsite.

we should discuss in Friday's WG

packages/core/src/Stepper.js
James300: We could always remove this functionality and make it the responsibility of the consumer to map the children using `Stepper.Step`

packages/core/src/Step.js
sdalonzo: is the fast follow ticketed and will be worked in the short term? i'd rather omit this from the docsite than force bundling of all the icons if not

packages/core/src/Stepper.js
sdalonzo: that seems sensible

packages/core/src/Step.js
James300: Agreed, I think that's a better short-term solution, I'll make the change.

packages/core/src/Step.js
```diff
@@ -0,0 +1,43 @@
...
+  cursor: ${props => (props.onClick ? 'pointer' : 'default')};
+`
+
+function Step({ className, active, completed, children, onClick }) {
+  const color = active || completed ? 'primary' : 'text.light'
+
+  return (
+    <StyledFlex className={className} alignItems="center" onClick={onClick}>
+      {completed && (
+        <CheckIcon color="primary" size={16} mr={1} legacy={false} />
```
dgiraldo313: Considering this Icon is not purely decorative and it provides context to the state of the current state, we should add an `aria-label="Step Completed"` or make the label dynamic if needed, so that it can be passed as a prop. This is needed so that when Screen Reader users tab onto this item, they know that it has been completed.

sdalonzo: - [ ] add an aria-label="Step Completed" or make the label dynamic if needed

packages/core/src/Step.js
```diff
@@ -0,0 +1,65 @@
...
+
+function Step({ className, active, completed, children, onClick }) {
+  const color = active || completed ? 'primary' : 'text.light'
+
+  return (
+    <StyledFlex
+      className={className}
+      disabled={!onClick}
+      alignItems="center"
+      aria-selected={active}
```
dgiraldo313: Button does not support `aria-selected`, so we need to move this to the text inside of it.
```suggestion
```

packages/core/src/Step.js
```diff
@@ -0,0 +1,65 @@
...
+      onClick={onClick}
+    >
+      {completed && (
+        <CheckIcon
+          color="primary"
+          size={16}
+          mr={1}
+          legacy={false}
+          aria-label="Step Completed"
+        />
```
dgiraldo313: We need to overwrite the icon's default aria-hidden prop, so it can be picked up by the screen reader.
```suggestion
          aria-hidden="false"
        />
```

packages/core/src/Step.js
```diff
@@ -0,0 +1,65 @@
...
+      )}
+      <Text color={color} bold={active} fontSize={1}>
```
dgiraldo313: Moved from the button above.
```suggestion
      <Text color={color} bold={active} fontSize={1} aria-selected={active}>
```

packages/core/src/Stepper.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { themeGet } from 'styled-system'
+import Flex from './Flex'
+import Step from './Step'
+
+const StyledFlex = styled(Flex)`
```
BeniCheni: I think the name could be `StepperFlex`, since we know the `styled` API is implied and the term "StepperFlex" can relate the building block `Flex` with the purpose `Stepper` together.

PR #585, opened 2019-09-11
Added props to Icon element so the PropType check doesn't complain

bryantp:
Props Check was failing because title, titleId, desc and descId were not being passed down into the `Icon` element from the `TransparentButton` element

### DIFF
--- a/packages/core/src/IconButton.js
+++ b/packages/core/src/IconButton.js
@@ -19,7 +19,7 @@ const TransparentButton = styled(Button)`
 const IconButton = ({ name, size, color, ...props }) => (
   <TransparentButton {...props}>
     <div>
-      <Icon name={name} size={size} color={color} />
+      <Icon name={name} size={size} color={color} {...props} />
     </div>
   </TransparentButton>
 )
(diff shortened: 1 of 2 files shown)


packages/core/src/IconButton.js
```diff
@@ -19,7 +19,15 @@ const TransparentButton = styled(Button)`
 const IconButton = ({ name, size, color, ...props }) => (
   <TransparentButton {...props}>
     <div>
-      <Icon name={name} size={size} color={color} />
+      <Icon
+        name={name}
+        size={size}
+        color={color}
+        title={props.title}
```
sdalonzo: any reason not to just spread the remaining props `{...props}`?

bryantp: It caused problems when I attached an `onClick()` handler. You end up with a ghost click due to the same handler being passed to `Icon` as well as `IconButton`.

sdalonzo: for a11y reasons, there should be a button wrapped around the icon instead of a direct click handler on the button, but manually mapping the props is fine.

BeniCheni: > It caused problems when I attached an onClick() handler. You end up with a ghost click due to the same handler being passed to Icon as well as IconButton.

Interesting observation. @bryantp, can you provide an example (would be great if a CodeSandbox link) to understand more about why `...props` would fail this use case.

PR #587, opened 2019-09-13
feat(single-handle-slider): Export Slider

jes708:
Export { Slider } and RangeSlider from pcln-slider

This PR mostly moves some files around for better organization, and exports Slider, which already existed prior to this PR but wasn't being exported.  This PR also updates docs to reflect the new Slider export.


bertya:
LGTM, just need to double-check our current release process that should we update version here or after merge

jes708:
> LGTM, just need to double-check our current release process that should we update version here or after merge

I think you may be right.  I will revert the versions, and take care of that on publish.

packages/slider/src/RangeSlider.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import Range from './rc-slider/Range'
+import { space, color } from 'styled-system'
+import styleSlider from './styleSlider'
+
+const RangeSlider = styleSlider(Range)
+
+RangeSlider.defaultProps = {
+  allowCross: false,
+  color: 'blue'
```
James300: This should be `primary.base`

packages/slider/src/RangeSlider.js
```diff
@@ -0,0 +1,18 @@
...
+}
+
+RangeSlider.propTypes = {
+  ...space.propTypes,
+  ...color.propTypes
```
James300: I don't think we should be using `color` from styled-system anymore

packages/slider/src/RangeSlider.js
jes708: Can that be done in a follow-up PR? This version of pcln-slider does not require DS 2.3.1 yet.

packages/slider/src/RangeSlider.js
jes708: This version of pcln-slider does not require DS 2.3.1 yet, so it should be ok for now right?

packages/slider/src/RangeSlider.js
James300: Sounds good to me

PR #588, opened 2019-09-18
refactor: update color theme fn per "text" color & migration guide

BeniCheni:
When `color="text"` is used, `bg` prop is required to support theme. Otherwise, "text" color falls back to use original "text" from original color set.

- update line height standard pellet in theme to 1.4 per @wesleymartin85 

- a new `Box` story - Theme user case: color=text

Screen Shot 2019-09-18 at 11 33 39 AM

- "text" color use case in migration guide

```
"text" Color Use Case

- If "text" color is used (i.e. `color="text"`), the `bg` prop is required for
  theming. Palette namespace is also suggested to be provided for both `color` &
  `bg` props.

<Box mt={3} color="text.lightest" bg="background.dark">
  Theme 1: color="text.lightest" value & bg="background.dark"
</Box>
<Box mt={3} color="text" bg="background.lighted">
  Theme 2: color="text" & "bg"="background.lighted"
</Box>
```

### DIFF
--- a/docs/pages/Theming.md
+++ b/docs/pages/Theming.md
@@ -489,3 +489,18 @@ global.mountWithTheme = (node, options) => {
 ## Truncate
 
 - No changes required.
+
+## "text" Color Use Case
+
+- If "text" color is used (i.e. `color="text"`), the `bg` prop is required for
+  theming. Palette namespace is also suggested to be provided for both `color` &
+  `bg` props.
+
+```.jsx
+<Box mt={3} color="text.lightest" bg="background.dark">
+  Theme 1: color="text.lightest" value & bg="background.dark"
+</Box>
+<Box mt={3} color="text" bg="background.lighted">
+  Theme 2: color="text" & "bg"="background.lighted"
+</Box>
+```
--- a/packages/core/src/theme.js
+++ b/packages/core/src/theme.js
@@ -42,7 +42,7 @@ export const fontWeights = {
 }
 
 export const lineHeights = {
-  standard: 1.5,
+  standard: 1.4,
   display: 1.25
 }
 
--- a/packages/core/storybook/Box.js
+++ b/packages/core/storybook/Box.js
@@ -112,8 +112,8 @@ storiesOf('Box', module)
         <Box mt={3} color="text.lightest" bg="background.dark">
           Theme 1: color="text.lightest" value & bg="background.dark"
         </Box>
-        <Box mt={3} color="text" bg="background.lighted">
-          Theme 2: color="text" & "bg"="background.lighted"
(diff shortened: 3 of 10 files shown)


packages/core/src/theme.js
```diff
@@ -42,7 +42,7 @@ export const fontWeights = {
 }
 
 export const lineHeights = {
-  standard: 1.5,
+  standard: 1.4,
```
jes708: So this was never supposed to be 1.5?

jes708: Would it be a breaking change to change the line-height now?  Should this wait until 3.0.0?

packages/core/storybook/Box.js
```diff
@@ -96,3 +96,25 @@ storiesOf('Box', module)
...
+        </Box>
+        <Box mt={3} color="purple">
+          color="purple" as none "text" color w/o bg prop
+        </Box>
+      </Box>
+      <Box p={5}>
+        <Box mt={3} color="text.lightest" bg="background.dark">
+          Theme 1: color="text.lightest" value & bg="background.dark"
+        </Box>
+        <Box mt={3} color="text" bg="background.lighted">
```
jes708: `lighted` should be changed to `lightest` on 115 and 116

docs/pages/Theming.md
```diff
@@ -489,3 +489,18 @@ global.mountWithTheme = (node, options) => {
...
+
+- If "text" color is used (i.e. `color="text"`), the `bg` prop is required for
+  theming. Palette namespace is also suggested to be provided for both `color` &
+  `bg` props.
+
+```.jsx
+<Box mt={3} color="text.lightest" bg="background.dark">
+  Theme 1: color="text.lightest" value & bg="background.dark"
+</Box>
+<Box mt={3} color="text" bg="background.lighted">
```
jes708: 503 and 504 should be `lightest` rather than `lighted`

packages/core/src/theme.js
BeniCheni: @jes708, shared more offline to explain about the expect 1.5, but it's the intended update and no style effect as this standard pellet of line height is not really used by anything.

BeniCheni:
@jes708, nice catch on the typo "lighted". Corrected to "lightest". 🙏 PTAL?

BeniCheni:
@jes708 / @sdalonzo / @achoboter, was able to confirm the `Hug` side effect of green text, and in [900848f](https://github.com/pricelinelabs/design-system/pull/588/commits/900848fd0ad45cf6f69317e34f61cba8b7fc9d5d) to update `color` prop's default value from ~`secondary`~ to `text.lightest`, in order to avoid the green text side effect since theme in v2.3.1.

PR #596, opened 2019-09-25
fix: Remove-h5-from-banner-component

suyashchopra19:


packages/core/src/Banner.js
```diff
@@ -74,7 +74,7 @@ const Banner = props => {
         )}
         <Box width={1}>
           <Text textAlign={props.textAlign}>
-            <Heading.h5 textStyle="display2">{props.header}</Heading.h5>
+            <Text textStyle="display2">{props.header}</Text>
```
sdalonzo: should this maybe just accept a header prop as a node, so that headers can be used if they are wanted?

@wesleymartin85 any thoughts on whether Headings might be desirable in banners?

dgiraldo313: @sdalonzo For accessibility purposes, we don't want to include a heading inside this banner, UNLESS, there is other content inside the banner that is related to the heading.

Maybe this banner could just accept a node, to make it a bit more flexible?

Screen Shot 2019-09-26 at 11 43 13 AM

dgiraldo313: Why is this `Text` component inside another text component? Couldn't we just achieve the same by doing something like this? `<Text textAlign={props.textAlign} textStyle="display2">{props.header}</Text>`

dgiraldo313: @sdalonzo Looking at the source code more closely, it looks like there is content under the header. So I think it would make sense to accept a node for the header. Maybe we can discuss this and fix this as a follow fix, as to not block this PR?

sdalonzo: works for me

dgiraldo313:
Follow up issue: #600

PR #597, opened 2019-09-25
Hide with inline

### LINKED ISSUE #553: Hide should have a default display of inline

donnobot:
Added inline prop for hide
Fixes #553

### DIFF
--- a/packages/core/src/Hide.js
+++ b/packages/core/src/Hide.js
@@ -30,10 +30,18 @@ const hidden = key => props =>
       }
     : null
 
+const hideInline = props =>
+  props.inline
+    ? {
+        display: 'inline'
+      }
+    : null
+
 const Hide = styled(Box)`
   ${hidden('xs')} ${hidden('sm')} ${hidden('md')} ${hidden('lg')} ${hidden(
   'xl'
 )} ${hidden('print')};
+${hideInline}
 `
 
 Hide.propTypes = {
--- a/packages/core/src/Hide.js
+++ b/packages/core/src/Hide.js
@@ -30,18 +30,11 @@ const hidden = key => props =>
       }
     : null
 
-const hideInline = props =>
-  props.inline
-    ? {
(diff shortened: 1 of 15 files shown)


packages/core/test/Hide.js
```diff
@@ -36,4 +36,10 @@ describe('Hide', () => {
     const json = rendererCreateWithTheme(<Hide print />).toJSON()
     expect(json).toMatchSnapshot()
   })
+
+  test('renders with inline prop', () => {
+    const json = rendererCreateWithTheme(<Hide inline />).toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('display', 'inline')
```
BeniCheni: 👍

packages/core/src/Hide.js
```diff
@@ -34,6 +34,7 @@ const Hide = styled(Box)`
   ${hidden('xs')} ${hidden('sm')} ${hidden('md')} ${hidden('lg')} ${hidden(
   'xl'
 )} ${hidden('print')};
+  display: ${props => (props.inline ? 'inline;' : null)}
```
sdalonzo: no propType?

packages/core/storybook/Hide.js
```diff
@@ -22,5 +22,8 @@ storiesOf('Hide', module).add('Hide', () => (
     <Hide print p={2} bg="yellow">
       Hide print
     </Hide>
+    <Hide inline p={2} bg="blue">
```
sdalonzo: without a `sm`, `md`, etc, Hide is effectively a no-op

sdalonzo:
This also doesn't satisfy the requirement of the issue https://github.com/pricelinelabs/design-system/issues/553: `Hide.text` should be declared as a variation of Hide, that uses a `span` (`<Text.span>`) instead of <div>, but otherwise functions the same

PR #599, opened 2019-09-27
Icon outline property fix

gbellini90:
-Added an "outline:none" to icons.
-Added "role:img" to icons
-Updated the coveragePathIgnorePatterns in package.json to include icons and slider. 
-Previous PR with approvals, pending coverage fix: https://github.com/pricelinelabs/design-system/pull/540

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,9 @@
     ],
     "coveragePathIgnorePatterns": [
       "/node_modules/",
-      "dist"
+      "dist",
+      "icons/lib/",
+      "slider/src/rc-slider"
     ],
     "coverageThreshold": {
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -57,7 +57,9 @@
     ],
     "coveragePathIgnorePatterns": [
       "/node_modules/",
-      "dist"
+      "dist",
+      "icons/lib/",
```
sdalonzo: i suggested these edits - `icons/lib` is compiled code, and `slider/src/rc-slider` contains 3rd party code that was copied in so we could add a11y. these were affecting coverage metrics on our own source code.

PR #601, opened 2019-09-27
icon outline

gbellini90:


packages/icons/svgr.config.js
```diff
@@ -20,13 +21,18 @@ export const ${componentName} = ({
   ${jsx}
 )
 
+const ${componentName} = styled(BaseComponent)\`
+  outline: none;
+\`
```
BeniCheni:
```suggestion
`
```

packages/icons/svgr.config.js
```diff
@@ -20,13 +21,18 @@ export const ${componentName} = ({
   ${jsx}
 )
 
+const ${componentName} = styled(BaseComponent)\`
```
BeniCheni:
```suggestion
const ${componentName} = styled(BaseComponent)`
```

packages/icons/svgr.config.js
gbellini90: From my older PR: Steve's comment on this 
it broke on nested backticks, this was the only thing that worked - since we are inside the template backticks that start on line 9 https://github.com/pricelinelabs/design-system/blob/Fix-outline-property-for-icons/packages/icons/svgr.config.js#L9 - we want them rendered into the built template, not evaluated.

Icon uses these svgr-built components, so it should be there - https://github.com/pricelinelabs/design-system/blob/Fix-outline-property-for-icons/packages/icons/src/Icon.js#L8

PR #603, opened 2019-10-01
Add text to hide2

### LINKED ISSUE #553: Hide should have a default display of inline

donnobot:
Hide to accept <Hide.text> which will display inline until a specified breakpoint window, then will be hidden
Fixes #553

### DIFF
--- a/packages/core/src/Hide.js
+++ b/packages/core/src/Hide.js
@@ -49,6 +49,9 @@ Hide.propTypes = {
 Hide.displayName = 'Hide'
 Hide.text = styled(Text)`
   display: inline;
+  ${hidden('xs')} ${hidden('sm')} ${hidden('md')} ${hidden('lg')} ${hidden(
+  'xl'
+)} ${hidden('print')};
 `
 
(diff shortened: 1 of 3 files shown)


packages/core/src/Hide.js
```diff
@@ -46,5 +47,11 @@ Hide.propTypes = {
 }
 
 Hide.displayName = 'Hide'
+Hide.text = styled(Text)`
+  display: inline;
+  ${hidden('xs')} ${hidden('sm')} ${hidden('md')} ${hidden('lg')} ${hidden(
```
unscsprt: NB: Does the linter let you put each of these on its own line?

donnobot: Yes, It does let me put everything on one line. Is this favored?
```
Hide.text = styled(Text)`
  display: inline;
  ${hidden('xs')}
  ${hidden('sm')}
  ${hidden('md')}
  ${hidden('lg')}
  ${hidden('xl')}
  ${hidden('print')};
```

PR #604, opened 2019-10-07
Feature/521 input font size

### LINKED ISSUE #521: Increase default fontSize to 16px for Input
Mobile browsers will zoom in on any `<Input>` selected that has a fontSize <16px.  This can lead to undesirable and confusing zooming that the user will have to counteract if the input also brings up an accompanying overlay when selected. To prevent this behavior, and the urge to disable zooming with custom meta viewport tags, should default `<Input>` fontSize be 16px instead of 14px?

jes708:
fixes #521 

Font-size will increase at smaller breakpoints by default to prevent automatic zoom on mobile devices.
Screen Shot 2019-10-07 at 1 09 32 PM
Screen Shot 2019-10-07 at 1 09 24 PM

### DIFF
--- a/packages/autocomplete/package.json
+++ b/packages/autocomplete/package.json
@@ -13,7 +13,6 @@
   "license": "MIT",
   "dependencies": {
     "downshift": "^3.2.6",
-    "pcln-design-system": "^2.0.4",
     "prop-types": "^15.7.2",
     "styled-system": "^3.2.1"
   },
@@ -22,11 +21,15 @@
     "@babel/core": "^7.3.3",
     "@reach/component-component": "^0.1.1",
     "cat-names": "^2.0.0",
+    "pcln-design-system": "^2.5.1",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "react-test-renderer": "^16.8.6",
     "styled-components": "^3.4.10",
     "us": "^2.0.0"
+  },
+  "peerDependencies": {
+    "pcln-design-system": ">=2.0.0"
   },
   "jest": {
     "setupFilesAfterEnv": [
--- a/packages/core/src/Select.js
+++ b/packages/core/src/Select.js
@@ -30,7 +30,7 @@ const SelectBase = styled.select`
   }
 `
 SelectBase.defaultProps = {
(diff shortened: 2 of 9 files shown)


packages/core/src/Select.js
```diff
@@ -30,7 +30,7 @@ const SelectBase = styled.select`
   }
 `
 SelectBase.defaultProps = {
-  fontSize: 1,
+  fontSize: [2, null, 1],
```
unscsprt: JW, what is the significance of making `null` the second element here?

jes708: It means it will also be 2.  I tend to prefer using null instead of 2 for the sake of inheritance.  If we decide we want to use 3 instead of 2 one day we would only have to change it in one place instead of two.

sdalonzo: iirc it also means that styled-system won't generate the extra media query if it isn't needed.

sdalonzo:
LGTM pending design approval, cc @wesleymartin85

packages/autocomplete/package.json
```diff
@@ -22,12 +21,16 @@
...
     "cat-names": "^2.0.0",
+    "pcln-design-system": "^2.5.1",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "react-test-renderer": "^16.8.6",
     "styled-components": "^3.4.10",
     "us": "^2.0.0"
   },
+  "peerDependencies": {
+    "pcln-design-system": ">=2.0.0"
```
BeniCheni: Can we denote peer dependencies to be `>= 2.5.1`, since there were bugs in 2.4.0 & 2.5.0?

jes708: My thinking here was that 2.0.4 was the former dependency I had replaced here, so technically any version between 2.0.0 and 2.5.1 should still be usable, aside from 2.4.0 and 2.5.0

BeniCheni: True also for functional versions >=2.0.0 that are not 2.4.0 & 2.5.0. A discussion also happened to unpublish the broken versions, which would resolve this concern of peer dev.

Therefore LGTM. Thanks.

PR #614, opened 2019-10-11
Ecosystem

sdalonzo:
this PR is in preparation for updating the bundling of packages as discussed, so that all are handled by rollup.

- upgrade React devDep to latest
- set React peerDep on all packages
- upgrade `rollup` and ecosystem
- upgrade to latest supported `styled-components`
- resolves conflicting version resolutions to make lerna faster
- upgrades jest and `@testing-library/react`


babel.config.js
```diff
@@ -1,9 +1,13 @@
...
   plugins: [
-    '@babel/transform-runtime',
     '@babel/proposal-class-properties',
     '@babel/proposal-object-rest-spread',
     '@babel/plugin-proposal-optional-chaining'
-  ]
+  ],
+  env: {
+    test: {
+      plugins: ['@babel/transform-runtime']
```
James300: 👍

docs/package.json
```diff
@@ -16,9 +16,9 @@
     "pcln-modal": "^2.0.4",
     "pcln-slider": "^2.2.0",
     "prop-types": "^15.6.2",
-    "react": "^16.4.1",
-    "react-dom": "^16.4.1",
-    "styled-components": "^3.4.10"
+    "react": "^16.10.2",
+    "react-dom": "^16.10.2",
```
James300: Awesome! I created https://bitbucket.corp.priceline.com/projects/CC/repos/pcln-react-hooks/browse the other day, still need to work on finishing the build process / publishing but I'm hoping to see more hook usage starting.

sdalonzo: awesome. this only sets the dev dep, for now. hooks as a peer will probably have to wait for the next semver major

packages/core/src/GreenButton.js
```diff
@@ -1,7 +1,11 @@
-import styled from 'styled-components'
+import React from 'react'
 import Button from './Button'
 
-const GreenButton = styled(Button).attrs({ color: 'secondary' })``
+const GreenButton = ({ children, ...props }) => (
+  <Button color="secondary" {...props}>
+    {children}
+  </Button>
+)
```
James300: Should we add a deprecation warning, saying that this will be removed in v3?

packages/core/src/OutlineButton.js
```diff
@@ -1,7 +1,11 @@
-import styled from 'styled-components'
+import React from 'react'
 import Button from './Button'
 
-const OutlineButton = styled(Button).attrs({ variation: 'outline' })``
+const OutlineButton = ({ children, ...props }) => (
+  <Button variation="outline" {...props}>
+    {children}
+  </Button>
+)
```
James300: Same thing here for the deprecation warning.

docs/package.json
James300: For sure and I can't imagine to much use in the DS but this will encourage consuming apps to upgrade and use hooks 😄

packages/core/src/GreenButton.js
sdalonzo: absolutely - just needed to fix it for now, since it broke with styled-components 4 which is an accepted peer

packages/core/src/Image.js
```diff
@@ -1,7 +1,7 @@
 import styled from 'styled-components'
 
 const Image = styled.img.attrs(props => ({
-  height: 'auto'
+  height: props.height || 'auto'
```
dansternfeld1: This had to be affecting more than a few use cases, right?

packages/core/test/__snapshots__/Absolute.js.snap
```diff
@@ -11,6 +11,6 @@ exports[`Absolute renders with top, left and zIndex props 1`] = `
 }
 
 <div
-  className="c0 "
+  className="c0"
```
dansternfeld1: (n-b) huh? weird.

packages/modal/test/ScrollLock.js
```diff
@@ -1,5 +1,9 @@
 import { ScrollLock } from '../src/index'
 
+beforeEach(() => {
+  global.scrollTo = jest.fn()
```
dansternfeld1: 👍

packages/popover/src/Popover.js
```diff
@@ -2,7 +2,6 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { Manager, Reference } from 'react-popper'
-import { theme } from 'pcln-design-system'
```
dansternfeld1: 🆗

packages/slider/package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
-    "prop-types": "^15.5.4",
+    "prop-types": "^15.7.2",
```
dansternfeld1: (n-b) This is latest, I guess nothing in React 16+ changes proptype defs?

storybook/preview-head.html
```diff
@@ -1 +1,4 @@
-<link href="https://fonts.googleapis.com/css?family=Montserrat:400,600" rel="stylesheet">
\ No newline at end of file
+<link
+  href="https://fonts.googleapis.com/css?family=Montserrat:500,700"
```
dansternfeld1: Was this just missed with the initial v2 changes? Those font weight diffs are not new.

packages/core/src/Image.js
sdalonzo: it broke Placeholder image with styled-components v4

packages/core/test/__snapshots__/Absolute.js.snap
sdalonzo: yea a little strange

packages/slider/package.json
sdalonzo: shouldn't, and every other package was using 15.7.2 anyway - so this just resolves a warning. it was getting this version hoisted anyway

storybook/preview-head.html
sdalonzo: correct, storybook just had the wrong weights. DSv2 changed 400,600 to 500,700

PR #615, opened 2019-10-11
Fix unmapped icon handlement

### LINKED ISSUE #610: Unknown icon names can break unit tests
We have seen cases where unit tests will break with `Icon(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.`.  This is caused by improperly named icons, such as `<Icon name='moon' />` instead of `<Icon name='Overnight' />`

This issue was introduced at some point after pcln-design-system 2.3.2

BeniCheni:
Fixes #610 

Plus rearrange to use less default props for html attributes that could be inline in JSX.

##### Test to validate all existing a11y attributes and retain working icon feature

Screen Shot 2019-10-11 at 5 08 15 PM

### DIFF
--- a/packages/icons/src/Icon.js
+++ b/packages/icons/src/Icon.js
@@ -10,29 +10,35 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
   ariaLabelledBy += desc && descId ? ` ${descId}` : ''
   ariaLabelledBy = ariaLabelledBy ? ariaLabelledBy : undefined
 
-  return (
-    Component && (
-      <Component
-        title={title}
-        desc={desc}
-        titleId={titleId}
-        descId={descId}
-        aria-hidden={!!ariaLabelledBy}
-        aria-labelledby={ariaLabelledBy}
-        {...props}
-      />
+  if (!Component) {
+    if (process.env.NODE_ENV !== 'production')
+      console.trace(`icon ${iconName} does not exist`)
+    return null
+  } else {
+    return (
+      Component && (
+        <Component
+          aria-hidden="true"
+          title={title}
+          desc={desc}
+          titleId={titleId}
+          descId={descId}
+          aria-hidden={!!ariaLabelledBy || true}
+          aria-labelledby={ariaLabelledBy}
+          tabIndex={-1}
+          focusable={false}
+          {...props}
+        />
+      )
     )
-  )
+  }
 }
 
 Icon.isIcon = true
 Icon.displayName = 'Icon'
 
 Icon.defaultProps = {
-  size: 24,
-  tabIndex: -1,
-  focusable: false,
-  'aria-hidden': true
+  size: 24
 }
 
 Icon.propTypes = {
--- a/packages/icons/src/Icon.js
+++ b/packages/icons/src/Icon.js
@@ -22,7 +22,7 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
         desc={desc}
         titleId={titleId}
         descId={descId}
-        aria-hidden={!!ariaLabelledBy || true}
+        aria-hidden={!ariaLabelledBy}
         aria-labelledby={ariaLabelledBy}
         tabIndex={-1}
         focusable={false}
--- a/packages/icons/src/Icon.js
+++ b/packages/icons/src/Icon.js
@@ -17,7 +17,6 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
   } else {
     return (
       <Component
-        aria-hidden="true"
         title={title}
         desc={desc}
         titleId={titleId}
--- a/packages/icons/src/Icon.js
+++ b/packages/icons/src/Icon.js
@@ -16,20 +16,18 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
     return null
   } else {
     return (
-      Component && (
-        <Component
-          aria-hidden="true"
-          title={title}
-          desc={desc}
-          titleId={titleId}
-          descId={descId}
-          aria-hidden={!!ariaLabelledBy || true}
-          aria-labelledby={ariaLabelledBy}
-          tabIndex={-1}
-          focusable={false}
-          {...props}
-        />
-      )
+      <Component
+        aria-hidden="true"
+        title={title}
+        desc={desc}
+        titleId={titleId}
+        descId={descId}
+        aria-hidden={!!ariaLabelledBy || true}
+        aria-labelledby={ariaLabelledBy}
+        tabIndex={-1}
+        focusable={false}
+        {...props}
+      />
     )
   }
 }
--- a/packages/icons/src/Icon.js
+++ b/packages/icons/src/Icon.js
@@ -11,8 +11,9 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
   ariaLabelledBy = ariaLabelledBy ? ariaLabelledBy : undefined
 
   if (!Component) {
-    if (process.env.NODE_ENV !== 'production')
+    if (process.env.NODE_ENV !== 'production') {
       console.trace(`icon ${iconName} does not exist`)
+    }
     return null
   } else {
     return (
--- a/packages/icons/test/index.js
+++ b/packages/icons/test/index.js
@@ -82,16 +82,18 @@ describe('Icon', () => {
         />
       )
       const testInstance = testRenderer.root
-
       const title = testInstance.findAll(el => el.type == 'title')
       expect(title).toHaveLength(0)
       expect(testInstance.findByType('desc').children[0]).toBe(
         'Accessible Logo description'
       )
-      expect(testRenderer.toJSON().props['aria-hidden']).toBe(true)
-      expect(testRenderer.toJSON().props['focusable']).toBe(false)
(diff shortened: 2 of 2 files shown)


packages/icons/src/Icon.js
```diff
@@ -10,29 +10,35 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
...
-        aria-labelledby={ariaLabelledBy}
-        {...props}
-      />
+  if (!Component) {
+    if (process.env.NODE_ENV !== 'production')
+      console.trace(`icon ${iconName} does not exist`)
+    return null
+  } else {
+    return (
+      Component && (
```
jes708: This check `Component && ` shouldn't be necessary anymore, since we know from the conditional that Component does exist.

packages/icons/src/Icon.js
```diff
@@ -10,29 +10,35 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
...
+        <Component
+          aria-hidden="true"
+          title={title}
+          desc={desc}
+          titleId={titleId}
+          descId={descId}
+          aria-hidden={!!ariaLabelledBy || true}
+          aria-labelledby={ariaLabelledBy}
+          tabIndex={-1}
+          focusable={false}
```
jes708: I think this might need to be focusable='false', but if you're seeing it correctly in the DOM, then I'll stand corrected :)

jes708:
packages/icons/src/Svg.js also contains the defaultProps you adjusted, so I think an update is needed there too.

packages/icons/src/Icon.js
BeniCheni: I think React suggests to use JavaScript convention like (-1 as a number for tableIdex, which is also camel cased by React, and then these `false` / `true` value)

packages/icons/src/Icon.js
```diff
@@ -10,29 +10,33 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
...
+  } else {
+    return (
       <Component
+        aria-hidden="true"
         title={title}
         desc={desc}
         titleId={titleId}
         descId={descId}
-        aria-hidden={!!ariaLabelledBy}
+        aria-hidden={!!ariaLabelledBy || true}
```
jes708: looks like aria-hidden is specified twice in this Component

packages/icons/src/Icon.js
BeniCheni: @jes708 , good 👁 , updated.

packages/icons/src/Icon.js
```diff
@@ -10,29 +10,33 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
...
-  return (
-    Component && (
+  if (!Component) {
+    if (process.env.NODE_ENV !== 'production')
+      console.trace(`icon ${iconName} does not exist`)
+    return null
+  } else {
+    return (
       <Component
+        aria-hidden="true"
```
jes708: Here is the first aria-hidden

packages/icons/src/Icon.js
BeniCheni: @jes708, only once... I've deleted the aria-hidden from default props, and use it as a default case here.

The thought behind it is to use less default props with weird object key quoting like 'aria-hidden', and use the aria-hidden = true as a fault back value by `||` operator, when aria labelled by value is not valid.

packages/icons/src/Icon.js
BeniCheni: Ah, yeah. Removed it.

BeniCheni:
> packages/icons/src/Svg.js also contains the defaultProps you adjusted, so I think an update is needed there too.

@jes708, looks like the latest in master corrected these already:

https://github.com/pricelinelabs/design-system/blob/39beb570538f22489424b2f3ce355dc7c7cd8b37/packages/icons/svgr.config.js#L32-L34

packages/icons/src/Icon.js
```diff
@@ -10,29 +10,32 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
...
+    return null
+  } else {
+    return (
       <Component
         title={title}
         desc={desc}
         titleId={titleId}
         descId={descId}
-        aria-hidden={!!ariaLabelledBy}
+        aria-hidden={!!ariaLabelledBy || true}
```
James300: Do you think it would make more sense to change `undefined` on line 11 to true?

sdalonzo: yea, this seems incorrect. `X || true` is always truthy, and `!!` already coerced `ariaLabelledBy` to bool

sdalonzo:
> > packages/icons/src/Svg.js also contains the defaultProps you adjusted, so I think an update is needed there too.
> 
> @jes708, looks like the latest in master corrected these already:
> 
> https://github.com/pricelinelabs/design-system/blob/39beb570538f22489424b2f3ce355dc7c7cd8b37/packages/icons/svgr.config.js#L32-L34

i think @jes708 is saying that since they are now inline in the JSX, the svgr config doesn't need them since they aren't props

packages/icons/src/Icon.js
BeniCheni: @James300 / @sdalonzo,
 
Think there's a misunderstanding. At least here what I understood, which could be an overlook. Seems that the current version is correct logic to generate HTML attribute `aria-hidden="true"`, if `ariaLabelledBy` is invalid. Otherwise, `aria-hidden="desc-id-value"`.

Confirmed the logic w. @dgiraldo313 when doing the PR to ensure the a11y requirement. (ping @dgiraldo313 to review the PR, and chime in if he has any thoughts on this logic. Let me know, if I misunderstood your comments. Happy to colla. to get to the bottom of this thread.

Here's the logic from my interpretation.

1) `ariaLabelledBy` variant is determined by`titleId`, `desc` & `descId`. If any of those doesn't validate the rules, default value `undefined` is assigned to `ariaLabelledBy` variant:

https://github.com/pricelinelabs/design-system/blob/39beb570538f22489424b2f3ce355dc7c7cd8b37/packages/icons/src/Icon.js#L9-L11

2. In this update, the PR removed the default prop `aria-hidden: true`, and apply as fall-back default value `true`, if `ariaLabelledBy` from step 1 is ever invalid w. the intentional `undefined` value. This works because `!!undefined || true` => `false || true` => `true` => generate HTML attribute `aria-hidden="true"`.

    If there's a valid `ariaLabelledBy` value, the logic would be `!!'desc-id' || true` => generate HTML attribute `aria-hidden="desc-id"` for valid a11y.

```jsx
aria-hidden={!!ariaLabelledBy || true}
```

BeniCheni:
> i think @jes708 is saying that since they are now inline in the JSX, the svgr config doesn't need them since they aren't props

@sdalonzo, spoke to @jes708 offline before his LGTM about the svgr config. His comment was wondering if the HTML attribute values should be all string quoted, such as `'aria-hidden': 'true',` instead, which was a version before the previous previous before the PR. It seems a confusing thread, but we agreed that for React syntax (similar to camel case instead of the normal HTML attributes in lower case), these HTML attribute values could use JavaScript instead of string quoting.

Per your thought about revoming, I tried it out locally and the Icon unit tests started to fail. I wouldn't want to take too much risk. The currently inline attributes over in `Icon` was to replace its own default prop values, but the scope doesn't seem to cover svgr config.

(If you'd like, you can pull the branch of `unkown-icon-handlement` and try removing and see if any tests fail? Maybe colla on a follow PR, while this PR can move forward to fix the unmapped `name` prop in `Icon`.)

BeniCheni:
> mind using curly braces here? this one is a little tricky to read.

@sdalonzo, 😸 why not. [27bc50d](https://github.com/pricelinelabs/design-system/pull/615/commits/27bc50d78e5cafd5ea4826b7ad04c442073c68a9) updated.

packages/icons/src/Icon.js
sdalonzo: @BeniCheni except that the `!!` already coerced the `aria-labelled-by` to a bool:

image

BeniCheni:
> yea, this seems incorrect. X || true is always truthy, and !! already coerced ariaLabelledBy to bool

@sdalonzo, nice 🎣 Updated the aria-hidden logic to do something along the line such as: `'desc-id' || true`. PTAL?

packages/icons/src/Icon.js
```diff
@@ -10,29 +10,33 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
...
+    return null
+  } else {
+    return (
       <Component
         title={title}
         desc={desc}
         titleId={titleId}
         descId={descId}
-        aria-hidden={!!ariaLabelledBy}
+        aria-hidden={Boolean(ariaLabelledBy)}
```
jes708: Actually, this logic seems backwards to me.  Shouldn't it be `aria-hidden={!ariaLabelledBy}?`  If there is a label, hidden should be false, because we want the user to find the icon and have its label read by the screen reader.  Otherwise, if there is no label, do hide it as there is nothing for the screen reader to say

BeniCheni:
> Approving in anticipation of the discussed !ariaLabelledBy fix

@jes708 , updated to have final version `aria-hidden={!ariaLabelledBy}`.

PR #616, opened 2019-10-13
fix Icon references in docsite

sdalonzo:
most of these just clear warnings with the initial capital - the one that was actually broken was in `Layout`, it referred to `hotel` when the icon is called `Hotels`

### DIFF
--- a/docs/pages/Layout.md
+++ b/docs/pages/Layout.md
@@ -218,7 +218,7 @@ In a real application, these would likely be links and buttons.
   color='white'
   bg='blue'>
   <Icon
-    name='hotel'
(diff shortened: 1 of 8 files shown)


docs/pages/Layout.md
```diff
@@ -218,7 +218,7 @@ In a real application, these would likely be links and buttons.
   color='white'
   bg='blue'>
   <Icon
-    name='hotel'
+    name='Hotels'
```
sdalonzo: these two are the ones that broke the docsite, since `hotel` doesn't exist (it is `Hotels`)

PR #617, opened 2019-10-13
Fix input semantic colors

unixchad:


packages/core/src/Input.js
```diff
@@ -40,7 +40,7 @@ const Input = styled.input`
   margin: 0;
 
   ::placeholder {
-    color: ${themeGet('colors.gray')};
+    color: ${themeGet('palette.text.light')};
```
jes708: Is there any difference between `themeGet` vs `getPaletteColor`?

unixchad: Fixed to use `getPaletteColor`

PR #619, opened 2019-10-14
v2.2.1 for pcln-icons

BeniCheni:
v2.2.1 for pcln-icons


jes708:
Should there be a package-lock.json update as well?  Did you use `npm version patch`?

BeniCheni:
> Should there be a package-lock.json update as well? Did you use npm version patch?

There's no package lock file in "packages/icons". Always use `npm version patch` to systematically update version attribute in files.

PR #620, opened 2019-10-15
Fixed undefined bug on IconName var

gabebravo:
using a ternary to key off of the name passed in as a prop. tested 3 scenarios: blank string, wrong icon name, no prop at all. all worked and gave the props message.

### DIFF
--- a/packages/icons/src/Icon.js
+++ b/packages/icons/src/Icon.js
@@ -12,7 +12,7 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
 
   if (!Component) {
     if (process.env.NODE_ENV !== 'production') {
-      console.trace(`icon ${iconName} does not exist`)
+      console.trace(`icon${name ? ' ' + name + ' ' : ' '}does not exist`)
     }
     return null
   } else {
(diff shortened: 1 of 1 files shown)


packages/icons/src/Icon.js
```diff
@@ -12,7 +12,7 @@ const Icon = ({ name, title, desc, titleId, descId, ...props }) => {
 
   if (!Component) {
     if (process.env.NODE_ENV !== 'production') {
-      console.trace(`icon ${iconName} does not exist`)
+      console.trace(`icon${name ? ' ' + name + ' ' : ' '}does not exist`)
```
jes708: Could be made a little simpler if you add a space before the template literal.

```console.trace(`icon ${name ? name + ' ' : ''}does not exist`)```

jes708: Alternatively, you could do this as well: ```console.trace(name ? `icon ${name} does not exist` : 'icon is missing name prop')```

gabebravo: Went with the second version. Liked the clarity of the message. Very useful. Went ahead and commited/pushed the changes, so the PR should be up-to-date with the changes.

PR #630, opened 2019-10-18
make package-lock use registry.npmjs.org

unixchad:


unixchad:
I removed the `package-lock` file then did an `npm install` and also an `npm audit fix`

PR #633, opened 2019-10-24
Fix broken Popover by changing innerRefs to refs

qualitymanifest:
Closes https://github.com/pricelinelabs/design-system/issues/631

Tests pass, Popover works as expected. See my comment on that issue for more details. Particularly of note: I was never able to reproduce the scroll reset bug.

### DIFF
--- a/packages/popover/src/Arrow.js
+++ b/packages/popover/src/Arrow.js
@@ -12,7 +12,7 @@ const PopoverArrow = ({
 }) => (
   <Arrow
     className={className}
-    innerRef={arrowProps.ref}
+    ref={arrowProps.ref}
     style={arrowProps.style}
     data-placement={placement}
     theme={theme}
--- a/packages/popover/src/Arrow.js
+++ b/packages/popover/src/Arrow.js
@@ -12,6 +12,7 @@ const PopoverArrow = ({
 }) => (
   <Arrow
     className={className}
+    innerRef={arrowProps.ref}
     ref={arrowProps.ref}
     style={arrowProps.style}
     data-placement={placement}
--- a/packages/popover/src/Popover.js
+++ b/packages/popover/src/Popover.js
@@ -59,11 +59,13 @@ class Popover extends Component {
         <Reference>
           {({ ref }) => (
             // Need to be a native element, because of ref forwarding limitations with DS functional components
-            <InlineContainer ref={ref}>
+            // Styled-Components v4 uses regular refs, keep innerRef for v3 peer dependency
+            <InlineContainer innerRef={ref} ref={ref}>
               {// Clone element to pass down toggle event so it can be used directly from children as needed
               React.cloneElement(this.props.children, {
                 'aria-label': 'Click to open popover with more information',
                 onClick: evt => this.handleToggle(evt, isPopoverOpen),
+                innerRef: this.triggerRef,
                 ref: this.triggerRef
               })}
             </InlineContainer>
--- a/packages/popover/src/Popover.js
+++ b/packages/popover/src/Popover.js
@@ -59,13 +59,12 @@ class Popover extends Component {
         <Reference>
           {({ ref }) => (
             // Need to be a native element, because of ref forwarding limitations with DS functional components
-            <InlineContainer innerRef={ref}>
+            <InlineContainer ref={ref}>
               {// Clone element to pass down toggle event so it can be used directly from children as needed
               React.cloneElement(this.props.children, {
                 'aria-label': 'Click to open popover with more information',
                 onClick: evt => this.handleToggle(evt, isPopoverOpen),
-                innerRef: this.triggerRef, // Need to use inner ref for DS core components using styled components v4
-                ref: this.triggerRef // Additionally need to add ref, in case native element is used as trigger
+                ref: this.triggerRef
               })}
             </InlineContainer>
           )}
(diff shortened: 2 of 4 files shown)


packages/popover/src/Arrow.js
```diff
@@ -12,7 +12,7 @@ const PopoverArrow = ({
 }) => (
   <Arrow
     className={className}
-    innerRef={arrowProps.ref}
+    ref={arrowProps.ref}
```
unscsprt: Hey @qualitymanifest, could you leave the old `innerRef` in addition to `ref`? Someone can correct me if I'm wrong, but we still need the old once since we support `styled-components@^3` as a peer, so we have to keep backward compatibility.

qualitymanifest: Sure, I'll add these back in a second

sdalonzo: actually, i don't think that is necessary. react 16.3 is already our floor and it supports forwardRefs

packages/popover/src/Popover.js
```diff
@@ -59,13 +59,14 @@ class Popover extends Component {
         <Reference>
           {({ ref }) => (
             // Need to be a native element, because of ref forwarding limitations with DS functional components
-            <InlineContainer innerRef={ref}>
+            // Styled-Components v4 uses regular refs, keep innerRef for v3 peer dependency
```
BeniCheni: Nit: Feel that the comment could be either ✂️, as google searching the topic styled-components v4 could land on some reference.

Or, the comment could become:
`// TODO fixme: when pcln-design-system future upgrade decides to drop support of innerRef prop as React legacy support is safe to deprecate, the innerRef prop here could be removed.`

packages/popover/src/PopoverContent.js
```diff
@@ -68,13 +68,11 @@ class PopoverContent extends Component {
         >
           {({ placement, ref, style, arrowProps }) => (
             // Need to be a native element, because of ref forwarding limitations with DS functional components
+            // Styled-Components v4 uses regular refs, keep innerRef for v3 peer dependency
```
BeniCheni: Similar feedback to comment at L62.

packages/popover/test/__snapshots__/Popover.js.snap
```diff
@@ -132,8 +132,9 @@ exports[`Popover UI Positioning Bottom 1`] = `
   aria-describedby="dialog-description-1"
   aria-label="Top PopOver"
   class="c0"
+  data-placement="bottom"
   role="dialog"
-  style="position: absolute; top: 0px; left: 0px; opacity: 0; pointer-events: none;"
+  style="position: fixed; top: 0px; left: 0px; transform: translate3d(NaNpx, NaNpx, 0); will-change: transform;"
```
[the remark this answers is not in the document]
qualitymanifest: I'm not sure, but the state of the snapshot with this PR is the same as the state of the working snapshots before the breaking change.

If you look at the [history](https://github.com/priceline/design-system/commits/master/packages/popover/test/__snapshots__/Popover.js.snap) of the snapshot, the first two show the same `NaNpx` values. The last one, [ecosystem upgrades](https://github.com/priceline/design-system/commit/fade1644002a450db623069b3bcd9c5d3d0d2c03#diff-d5732357e972e7ff921ce1f89a8fb406), is the one that upgraded styled components and broke the popover, and that is what this current diff is against.

I'll do a little digging and see if I can find why it is happening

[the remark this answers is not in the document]
BeniCheni: @sdalonzo, maybe we could colla. and quickly double checking this snapshot observation in the [ecosystem upgrades](https://github.com/priceline/design-system/commit/fade1644002a450db623069b3bcd9c5d3d0d2c03#diff-d5732357e972e7ff921ce1f89a8fb406) to get to the bottom of the understanding to clarify, in interests of moving forward?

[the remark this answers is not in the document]
qualitymanifest: Found this [related issue](https://github.com/FezVrasta/react-popper/issues/238) on react-popper, tried the proposed fix (wrapping Popper children in a div), no change. There is also an [open issue](https://github.com/FezVrasta/popper.js/issues/796) with the same problem on popper.js

packages/popover/src/Popover.js
qualitymanifest: Removed the comments

packages/popover/test/__snapshots__/Popover.js.snap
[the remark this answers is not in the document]
qualitymanifest: This might only be an issue with the snapshots. If you put in some curly braces here and add a console.log(style) before returning the content, and then open a Popover in the storybook, you get correct values like {position: "fixed", transform: "translate3d(5px, 320px, 0)", top: 0, left: 0, willChange: "transform"}.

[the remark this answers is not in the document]
sdalonzo: thank you for the research. i'm good with moving forward.

BeniCheni:
Landed in [a39c161](https://github.com/priceline/design-system/commit/a39c1613956011817848e7efe15e94043451c44e). 🎊 @qualitymanifest Thanks you for your contribution, and hope to see you around. 😃

PR #639, opened 2019-11-05
apply rollup to all packages except icons

sdalonzo:
Apply rollup to packages that don't use it yet (coming in an additional PR)
Add `npm run bundle:audit` command


sdalonzo:
i vetted these in a CRA

PR #641, opened 2019-11-06
Standardize the disabled button

James300:
We want one consistent look and feel for the disabled button.

Before:
image

After:
image


packages/core/src/Button.js
```diff
@@ -87,6 +87,12 @@ const Button = mapProps(({ fullWidth, ...props }) => ({
   border-style: solid;
   ${width} ${size} ${space};
   ${applyVariations('Button', variations)}
+
+  &:disabled {
+    cursor: not-allowed;
```
dansternfeld1: Neat! https://caniuse.com/#feat=mdn-css_properties_cursor_not-allowed

PR #644, opened 2019-11-14
WIP v3.0.0 of all packages - with tree-shaking

sdalonzo:


unscsprt:
[Tree-shaking](https://trello.com/c/n6r9l61r/3-tree-shaking)

packages/core/src/Avatar.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
```
jes708: I think new core components should be added to v2 as well, so we can have backwards compatibility in v3.

packages/core/src/Breadcrumbs.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
```
jes708: Add to v2

packages/core/src/Avatar.js
James300: You could look at it as incentive to move to v3 😜

jes708: I'm looking forward to inheriting the performance wins in v3 for sure. I still strongly recommend putting these components in v2. When working in a host app that consumes several dependencies that all need to switch over from v2 to v3, backwards compatibility really helps facilitate that transition.

packages/core/src/Banner.js
```diff
@@ -2,48 +2,57 @@ import React from 'react'
...
+  applyVariations,
+  hasPaletteColor,
+  deprecatedColorValue,
+  deprecatedPropType
+} from './utils'
 
 const bannerColors = {
   green: {
     backgroundColor: 'green',
     color: 'white',
```
achoboter: should we remove the old color use?

packages/core/src/Radio.js
```diff
@@ -45,21 +46,31 @@ const RadioInput = styled.input`
...
-
   const borderAdjustedSize = size + 4
 
+  const radioIcon = getRadioIcon({ checked, size: borderAdjustedSize })
+
   return (
     <RadioWrap color={props.color} checked={checked} disabled={disabled}>
       <RadioInput type="radio" {...props} />
-      <RadioIcon name={radioIconName} size={borderAdjustedSize} />
+      {radioIcon}
```
achoboter: `getRadioIcon` could just be a component

packages/core/src/Banner.js
```diff
@@ -56,11 +65,16 @@ const Banner = props => {
     bannerColors[
       !props.bg && props.color === 'green' ? props.color : props.bg
     ] || {}
-  const icon = props.iconName || bannerColor.icon
+  const Icon = props.icon || bannerColor.icon
   const color =
     hasPaletteColor(props) || !bannerColor.color
       ? props.color
       : bannerColor.color
+  const header = React.isValidElement(props.header) ? (
```
achoboter: I feel like this should first check if the `prop.header` exists, then check if it is a valid element or if we should create one, else header can be null.

Then we can remove the check on line 90 `{props.header && header}` and just use `{header}`, and then we aren't creating an empty header when `props.header` in undefined.

packages/core/test/__snapshots__/FormField.js.snap
```diff
@@ -18,15 +18,15 @@ exports[`FormField renders 1`] = `
   color: inherit;
   background-color: transparent;
   border-radius: 2px;
-  border-width: 0px;
+  border-width: 1px;
   border-style: solid;
+  border-color: #c0cad5;
```
achoboter: I don't know if this is an issue, but in this snap there are a lot of classes that have `border-color` twice

packages/core/test/__snapshots__/IconField.js.snap
```diff
@@ -28,15 +29,15 @@ exports[`IconField renders 1`] = `
   color: inherit;
   background-color: transparent;
   border-radius: 2px;
-  border-width: 0px;
+  border-width: 1px;
   border-style: solid;
+  border-color: #c0cad5;
```
achoboter: this snap file also has `border-color` twice a few times

packages/core/test/__snapshots__/Input.js.snap
```diff
@@ -11,15 +11,15 @@ exports[`Input it renders 1`] = `
   color: inherit;
   background-color: transparent;
   border-radius: 2px;
-  border-width: 0px;
+  border-width: 1px;
   border-style: solid;
+  border-color: #c0cad5;
```
achoboter: this one too has `border-color` twice a few times

packages/core/test/__snapshots__/Select.js.snap
```diff
@@ -42,6 +40,7 @@ exports[`Select renders 1`] = `
   border-width: 1px;
   border-style: solid;
   border-color: #c0cad5;
+  border-color: #007aff;
```
achoboter: double `border-color` with different values?

packages/icons/test/index.js
```diff
@@ -13,20 +16,13 @@ test.each(iconList)('renders %s', (key, Component) => {
...
-    expect(json).toEqual(expected)
+  test('defaults title to name if not provided', () => {
+    const testRenderer = TestRenderer.create(<Icon name="Accessible" />)
+    const testInstance = testRenderer.root
+    expect(testInstance.children[0].props.title).toBe('Accessible')
   })
 
-  describe('SVG Icon Accessibility', () => {
-    // Test individual icons in './lib'
+  describe.skip('SVG Icon Accessibility', () => {
```
achoboter: skipping this test?

packages/core/src/Avatar.js
James300: Yeah that totally makes sense, thanks for bringing this up. Here is the PR for v2 Avatar https://github.com/priceline/design-system/pull/678 👍

bmansell: PR for Breadcrumbs v2 as well https://github.com/priceline/design-system/pull/677!

packages/core/src/Breadcrumbs.js
bmansell: PR created! https://github.com/priceline/design-system/pull/677

sdalonzo:
> Additional backwards compatibility with v2 is needed

@jes708 is this addresses by the PRs added by @James300 and @bmansell ?

jes708:
> > Additional backwards compatibility with v2 is needed
> 
> @jes708 is this addresses by the PRs added by @James300 and @bmansell ?

Yep!  I'd like to merge and publish another core v2 patch for my label width fix (#680) before this PR gets merged if possible.

packages/core/src/Banner.js
sdalonzo: We should but I don't think that requires breaking change, so it need not block this. @achoboter agree?

sdalonzo:
> > > Additional backwards compatibility with v2 is needed
> > 
> > 
> > @jes708 is this addresses by the PRs added by @James300 and @bmansell ?
> 
> Yep! I'd like to merge and publish another core v2 patch for my label width fix (#680) before this PR gets merged if possible.

Great, thanks. I already dropped an approval on #680 so that works for me!

packages/core/src/Radio.js
sdalonzo: @achoboter I pushed this change.

packages/core/test/__snapshots__/FormField.js.snap
sdalonzo: i have a WIP branch to fix this

packages/core/src/Banner.js
sdalonzo: cc @unscsprt

packages/core/test/__snapshots__/FormField.js.snap
sdalonzo: Addressed in https://github.com/priceline/design-system/pull/683

packages/core/test/__snapshots__/Input.js.snap
sdalonzo: Addressed in https://github.com/priceline/design-system/pull/683

packages/core/test/__snapshots__/Select.js.snap
sdalonzo: Addressed in https://github.com/priceline/design-system/pull/683

packages/core/test/__snapshots__/IconField.js.snap
sdalonzo: Addressed in https://github.com/priceline/design-system/pull/683

packages/icons/test/index.js
sdalonzo: I'll address this today. Thanks for a good catch!

sdalonzo: Addressed in https://github.com/priceline/design-system/pull/685

Issue #513, opened 2019-06-04
Link should specify hover color


jes708:
We are currently using default browser color as our hover color for Links.  Should we perhaps use our own darkBlue instead?

krevat:
hover just adds the underline, no?

James300:
![](https://github.trello.services/images/mini-trello-icon.png) [Link should specify hover color](https://trello.com/c/bUDSprhL/30-link-should-specify-hover-color)

PR #648, opened 2019-11-14
Link uses dark shade one hover

unscsprt:
Addresses #513


unscsprt:
[Hover color for link](https://trello.com/c/U9yxKkk1/7-hover-color-for-link)

PR #649, opened 2019-11-14
Remove Icon usage from Hug, Select, and Step

unscsprt:
Removed `<Icon>` from:

- Hug
- Step
- Select

### DIFF
--- a/packages/core/src/Hug.js
+++ b/packages/core/src/Hug.js
@@ -2,10 +2,8 @@ import React from 'react'
 import Box from './Box'
 import Card from './Card'
 import Flex from './Flex'
-import Icon from './Icon'
 import Text from './Text'
-import styled, { withTheme } from 'styled-components'
-import { display } from 'styled-system'
+import styled from 'styled-components'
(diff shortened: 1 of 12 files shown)


unscsprt:
[Tree-shaking](https://trello.com/c/n6r9l61r/3-tree-shaking)

packages/core/src/Hug.js
```diff
@@ -24,28 +22,25 @@ const BorderConcealer = styled(Box)`
   }
 `
 
-const HideableIcon = styled(Icon)`
-  ${display};
-`
+const Hug = ({ bg, color, p, fontSize, icon, iconDisplay, ...props }) => {
```
James300: What could iconDisplay be?

unscsprt: Apparently that's a string for the CSS display property

James300: I feel like it's overkill to have that much power without doing something with css but I guess it's an existing thing 🤷‍♂

Issue #600, opened 2019-09-27
Allow banner to accept node for the header prop


dgiraldo313:
For A11y reasons, there are instances where we want the banner to accept a specific level headings in order to meet the desired outline of the page. It would be great if we could either pass in text or a node for the header section.

This ticket is a follow up to: #549

unscsprt:
![](https://github.trello.services/images/mini-trello-icon.png) [Banner should accept node as prop](https://trello.com/c/0tUMLdYn/8-banner-should-accept-node-as-prop)

James300:
Fixed now, coming in v3.

PR #651, opened 2019-11-14
Banner accepts node as header prop

unscsprt:
Addresses #600


unscsprt:
[Banner should accept node as header prop](https://trello.com/c/0tUMLdYn/8-banner-should-accept-node-as-header-prop)

Issue #628, opened 2019-10-18
Enable Popover to trap focus in popup when open


unscsprt:
I received feedback that screen reader users may be confused if they tab out of the popover's tooltip and their focus is lost or returned to an unexpected part of the page.

To solve for this, I wrapped my `renderContent` in `<FocusLock />`, which is provided by [react-focus-lock](https://www.npmjs.com/package/react-focus-lock). This prevents the user from tabbing out of the dialog without using an input (e.g. a close button) to dismiss it.

Can we add a prop to `Popover` that optionally wraps the children of `renderContent` in `<FocusLock>`?

James300:
Fixed now so closing the issue, will be released with v3.

PR #652, opened 2019-11-14
Add optional focus lock to Popover

unscsprt:
Addresses #628


unscsprt:
[Add focus trap to popover](https://trello.com/c/oqo1jsEu/10-add-focus-trap-to-popover)

Issue #461, opened 2019-03-18
IE 11 Flex Column Img Container Bug


jes708:
The following pattern causes a bug in IE11 wherein a bunch of extra height/space is added to the Image.  The problem is caused by the width on the Image-containing Box.
```
<Flex flexDirection='column'>
  <Box width={['100px', '200px']}>
    <Image />
  </Box>
  <Box />
</Flex>
```

Should this pattern be avoided or is there a way to safeguard this appropriately in the design-system?

My proposal is to enable responsive widths for the `Image` component so there is no need to add an Image container with responsive widths.

James300:
Let's check that this is still an issue and try to fix it if so.

unscsprt:
Confirmed this is still an issue in IE 11

unscsprt:
![](https://github.trello.services/images/mini-trello-icon.png) [IE image responsiveness bug](https://trello.com/c/k7k9P8se/11-ie-image-responsiveness-bug)

PR #653, opened 2019-11-14
Add width support to Image

unscsprt:
Addresses #461

### DIFF
--- a/packages/core/storybook/Image.js
+++ b/packages/core/storybook/Image.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { withInfo } from '@storybook/addon-info'
-import { Image, Box } from '../src'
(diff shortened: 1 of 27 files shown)


unscsprt:
[IE image responsiveness bug](https://trello.com/c/k7k9P8se/11-ie-image-responsiveness-bug)

packages/core/storybook/Image.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { storiesOf } from '@storybook/react'
 import { withInfo } from '@storybook/addon-info'
-import { Image, Box } from '../src'
+import { Image, Box, Flex } from '../src'
```
James300: Do you need Flex here?

James300: And you might be able to remove Box too

unscsprt: Good catch

Issue #645, opened 2019-11-14
New Component: Avatar


James300:
This sort of thing:
![image](https://user-images.githubusercontent.com/9325039/68832010-7a3d4f00-0675-11ea-8bb1-e1a88ceab824.png)

James300:
![](https://github.trello.services/images/mini-trello-icon.png) [Avatar Component](https://trello.com/c/nbQaEioF/31-avatar-component)

James300:
Completed

PR #658, opened 2019-11-14
Avatar component

James300:
https://github.com/priceline/design-system/issues/645

Storybook example:
image

### DIFF
--- a/packages/core/src/Avatar.js
+++ b/packages/core/src/Avatar.js
@@ -1,0 +1,50 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import Flex from './Flex'
+import Heading from './Heading'
+import Text from './Text'
+import { User } from 'pcln-icons'
+
+const StyledImage = styled(Flex)`
+  border-radius: 50%;
+  width: 40px;
+  height: 40px;
+  background-size: 40px;
+  background-image: url(${props => props.src});
+`
+
+function Avatar({ className, title, subTitle, src, color }) {
+  return (
+    <Flex className={className}>
+      <StyledImage src={src} color={color} p={2}>
+        {!src && <User />}
+      </StyledImage>
+      {title && (
+        <Flex flexDirection="column" justifyContent="center" ml={3}>
+          <Heading.h5 m={0}>{title}</Heading.h5>
+          <Text color="text.light" fontSize={0}>
+            {subTitle}
+          </Text>
+        </Flex>
+      )}
+    </Flex>
+  )
+}
+
+Avatar.displayName = 'Avatar'
+
+Avatar.propTypes = {
+  className: PropTypes.string,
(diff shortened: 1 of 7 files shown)


James300:
[Avatar Component](https://trello.com/c/nbQaEioF/31-avatar-component)

packages/core/src/Avatar.js
```diff
@@ -0,0 +1,50 @@
...
+import styled from 'styled-components'
+import Flex from './Flex'
+import Heading from './Heading'
+import Text from './Text'
+import { User } from 'pcln-icons'
+
+const StyledImage = styled(Flex)`
+  border-radius: 50%;
+  width: 40px;
+  height: 40px;
```
unscsprt: Would it make sense for the size to be configurable? One use case might be a small avatar in a menu bar and a large one on a profile page.

packages/core/src/Avatar.js
```diff
@@ -0,0 +1,50 @@
...
+}
+
+Avatar.displayName = 'Avatar'
+
+Avatar.propTypes = {
+  className: PropTypes.string,
+  title: PropTypes.string,
+  subTitle: PropTypes.string,
+  src: PropTypes.string,
+  color: PropTypes.string
```
unscsprt: It'd be nice if this could display initials if `src` isn't present.

packages/core/src/Avatar.js
```diff
@@ -0,0 +1,50 @@
...
+
+const StyledImage = styled(Flex)`
+  border-radius: 50%;
+  width: 40px;
+  height: 40px;
+  background-size: 40px;
+  background-image: url(${props => props.src});
+`
+
+function Avatar({ className, title, subTitle, src, color }) {
```
unscsprt: Subtitle is one word, so it should be all lowercase.

byakimchuk:
Since this is a new component, do we want to add it to the [public docs](https://github.com/priceline/design-system/tree/master/docs)?

James300:
@byakimchuk we are hopefully going to remove the docs project today and move over to storybook docs.

packages/core/src/Avatar.js
James300: Good suggestion: 
image

James300: This is the fallback if no src and no initials:
image

packages/core/src/Avatar.js
```diff
@@ -0,0 +1,55 @@
...
+import Heading from './Heading'
+import Text from './Text'
+import { User } from 'pcln-icons'
+
+const StyledImage = styled(Flex)`
+  justify-content: center;
+  align-items: center;
+  border-radius: 50%;
+  width: ${props => props.size || 40}px;
+  height: ${props => props.size || 40}px;
```
unscsprt: NB - Want to make `40` the default prop value instead?

Issue #529, opened 2019-07-03
Input and Select should both use border


jes708:
Only Select uses border, whereas Input uses box-shadow. This creates misalignments when both elements are used together on a form.  Input should use border as Select does currently.

byakimchuk:
![](https://github.trello.services/images/mini-trello-icon.png) [Input and Select should both use border](https://trello.com/c/aHP1wH4r/14-input-and-select-should-both-use-border)

byakimchuk:
Closing with the merge of https://github.com/priceline/design-system/pull/659

PR #659, opened 2019-11-14
Swap Input and TextArea to use border over box-shadow (#529)

byakimchuk:
Input, TextArea, and Select will now each use `utils/borders` for `borders`, removing inconsistent use of `box-shadow`.

Related to https://github.com/priceline/design-system/issues/529


byakimchuk:
@unscsprt looks like a test for `AutoComplete` is causing some trouble for the build. Tests are passing locally, any idea how I can fix this?

unscsprt:
[Input and Select should both use border](https://trello.com/c/aHP1wH4r/14-input-and-select-should-both-use-border)

unscsprt:
> @unscsprt looks like a test for `AutoComplete` is causing some trouble for the build. Tests are passing locally, any idea how I can fix this?

You may have a different version of `styled-components` installed than CI. I installed the branch and was able to update the autocomplete snapshot, so we're all good now.

PR #663, opened 2019-11-15
Update Banner docs page, remove use of Icon

unscsprt:


unscsprt:
[Add header as node example for Banner to docs site](https://trello.com/c/OhSbTMdh/36-add-header-as-node-example-for-banner-to-docs-site)

packages/core/src/Banner.js
```diff
@@ -73,9 +73,9 @@ const Banner = props => {
       color={color}
     >
       <Flex justifyContent="space-between" alignItems="flex-start">
-        {!!icon && !!props.showIcon && (
-          <Icon name={icon} mr={2} size={24} mt="-2px" />
-        )}
+        {!!icon &&
+          !!props.showIcon &&
+          React.cloneElement(props.icon, { size: 24, mt: '-2px', mr: 2 })}
```
James300: I did the same thing in my PR, we might get a few conflicts but I'll resolve them, it's my bad

docs/pages/Banner.md
```diff
@@ -67,9 +67,23 @@ Uses the color from `theme.palette.error.base`.
...
+## Heading
+
+`Banner` accepts a `Heading` as its `header` prop if a specific heading level is desired.
+
+```.jsx
+<Banner
+  p={2}
+  header={<Heading.h4>H4 Heading</Heading.h4>}
+  color="caution"
+  iconName="Attention"
```
James300: Should this be `icon` now?

docs/pages/Banner.md
```diff
@@ -56,7 +56,7 @@ Uses the color from `theme.palette.primary.base`.
 <Banner
   p={2}
   color="error"
-  iconName="Warning"
+  icon={<Icon name="Warning"/>}
```
James300: Should we be passing in more specific icons like `icon={<Warning />}`

unscsprt: That's what I would think, but apparently those aren't available however the MDX is configured. Any ideas?

docs/pages/Banner.md
unscsprt: Yes, good catch.

docs/pages/Banner.md
James300: That makes sense. Maybe we should leave this till the move to storybook docs then.

MIGRATING_to_v3.md
```diff
@@ -31,8 +31,12 @@ import { ThumbsUp } from 'pcln-icons
 } />
 ```
 
-- Hug
-  - Icon prop now takes a `node` instead of an icon name
+- Components that now take a `node` instead of an icon name
```
sdalonzo: instead of `iconName`?

docs/pages/Banner.md
sdalonzo: afaik you have to wait for publish on this. @unscsprt is correct.

Issue #447, opened 2019-02-04
Add engines config to package.json, update CONTRIBUTING.md


sdalonzo:
`next` branch fails to install with node `6.3.1` and npm `5.10.0` - these are common here at priceline. we should update the engines config https://docs.npmjs.com/files/package.json#engines and the contributing doc to indicate what the minimum version of each is.

BeniCheni:
There's a WG follow-up item to validate why there's a node version requirement, which came out from the WG session as myself as the assignee. So I added self to the **Assignees** section.

James300:
![](https://github.trello.services/images/mini-trello-icon.png) [Add engines config to package.json](https://trello.com/c/adz89kHl/39-add-engines-config-to-packagejson)

PR #668, opened 2019-11-15
Set Engine Config

James300:
Requires at least node: 8.16.0 and npm: 6.4.1

https://github.com/priceline/design-system/issues/447


James300:
[Add engines config to package.json](https://trello.com/c/adz89kHl/39-add-engines-config-to-packagejson)

Issue #270, opened 2018-04-02
Tooltip using a (to-be) deprecated prop


krevat:
Tooltip is passing `align` into a styled `Box`, resulting in console warning.

hakimelek:
I don't know if we can call this a deprecation, I don't see `align` exposed publicly as a proptype (anymore?). We can just go ahead and remove the `align` from Tooltip Proptype, and mention that `Tooltip` extends `Box`.

James300:
![](https://github.trello.services/images/mini-trello-icon.png) [Tooltip using deprecated prop](https://trello.com/c/SbWI0aQC/41-tooltip-using-deprecated-prop)

PR #669, opened 2019-11-16
Remove align from tooltip

James300:
https://github.com/priceline/design-system/issues/270


James300:
[Tooltip using deprecated prop](https://trello.com/c/SbWI0aQC/41-tooltip-using-deprecated-prop)

Issue #441, opened 2019-01-08
responsive Icon size


krevat:
Can we have `<Icon />` accept an array of sizes instead of just one? Maybe `fontSize`?

I noticed [this issue](https://github.com/pricelinelabs/design-system/issues/46) was opened and closed but it was just docs related. Is that because the capability was shot down / removed or because the capability didn't exist?

bertya:
After discussion, we want to move forward with this. We want to use `size` to accept an array.

dansternfeld1:
@adhagam -- please work with @sdalonzo and @gbellini90 to see how they did similar work on outline states

James300:
![](https://github.trello.services/images/mini-trello-icon.png) [Responsive Icon Size](https://trello.com/c/RUkMU6IT/43-responsive-icon-size)

PR #670, opened 2019-11-16
Responsive Icon Widths

James300:
https://github.com/priceline/design-system/issues/441

### DIFF
--- a/packages/icons/src/Svg.js
+++ b/packages/icons/src/Svg.js
@@ -65,6 +65,7 @@ const Svg = styled('svg')(
 
 Svg.propTypes = {
   ...space.propTypes,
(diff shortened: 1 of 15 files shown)


James300:
[Responsive Icon Size](https://trello.com/c/RUkMU6IT/43-responsive-icon-size)

packages/icons/src/Svg.js
```diff
@@ -58,6 +58,7 @@ const Svg = styled('svg')(
     flex: 'none'
   },
   space,
+  width,
```
byakimchuk: Should we include the propTypes for width since we're using it now?

`...width.propTypes`

PR #675, opened 2019-11-20
Utilities docs and unit tests

achoboter:
- adding documentation for utilities borders, color, getPaletteColor and getTextColorOn
- added unit tests for all utilities

### DIFF
--- a/docs/pages/utilities/color.md
+++ b/docs/pages/utilities/color.md
@@ -45,3 +45,5 @@ Adding a `color` prop to your component will set the background color to that co
 ```
 
 You can also supply a `bg` prop to force a specified text color and background color, ignoring the recommended text color.
+
+**Note:** The `bg` prop is marked to be deprecated in `v3`.
--- a/docs/pages/utilities/color.md
+++ b/docs/pages/utilities/color.md
@@ -1,0 +1,47 @@
+# color
+
+A utility function to apply text and background color to a component.
+
+## Syntax
+
+```js
+color(props)
+```
+
+## Usage
+
+```jsx
+const Box = styled.div`
+  ${color}
+`
+```
+
+## Arguments
+
+| Argument | Type   | Description                                                     |
+| -------- | ------ | --------------------------------------------------------------- |
+| `props`  | Object | The props passed to the component, must contain a `theme` prop. |
+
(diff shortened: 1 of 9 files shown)


docs/pages/utilities/color.md
```diff
@@ -0,0 +1,47 @@
...
+<Box color="primary">Primary color</Box>
+```
+
+Adding a `color` prop to your component will set the background color to that color and set the text color to either `text.lightest` or `text.base` depending on the contrast ratio with the background color.
+
+```.jsx
+<Box color="primary" bg="background">Primary on background color</Box>
+```
+
+You can also supply a `bg` prop to force a specified text color and background color, ignoring the recommended text color.
```
byakimchuk: It might be worth nothing that the `bg` prop is marked as deprecated in DS v3.

docs/pages/utilities/getPaletteColor.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+```.jsx
+<ShadeOtherDemo bgColor="primary">Primary</ShadeOtherDemo>
+<ShadeOtherDemo bgColor="primary.dark">Primary dark</ShadeOtherDemo>
+<ShadeOtherDemo bgColor="orange">Legacy orange</ShadeOtherDemo>
+<ShadeOtherDemo bgColor="#007aff">Hex color</ShadeOtherDemo>
+```
+
+You can also apply the utility to use a different prop to drive `color`, or in the above example, `background-color`.
```
James300: I added docs for getPaletteColor to the `themeing` section but these look better, wanna delete my docs too?

achoboter: this is pointed to merge into `master` which doesn't have those docs yet :P

docs/pages/utilities/color.md
achoboter: Have we vetted out how removing the `bg` prop will affect our applications? I know that it is used a ton, but I do agree that most of the current use should just be replace by `color` only.

sdalonzo: the plan is just to mark it deprecated, not to disable it until v4 for exactly that reason

PR #678, opened 2019-11-22
Avatar v2

James300:
V2 version of: https://github.com/priceline/design-system/issues/645

### DIFF
--- a/packages/core/src/Avatar.js
+++ b/packages/core/src/Avatar.js
@@ -1,0 +1,55 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import Flex from './Flex'
+import Heading from './Heading'
+import Text from './Text'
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -3396,7 +3396,7 @@
 		},
 		"@storybook/addon-a11y": {
 			"version": "4.1.18",
-			"resolved": "https://registry.npmjs.org/@storybook/addon-a11y/-/addon-a11y-4.1.18.tgz",
+			"resolved": "https://npm.prod.pcln.com/repository/npm-all/@storybook/addon-a11y/-/addon-a11y-4.1.18.tgz",
```
bmansell: Do we want to keep the changes in this file for this PR? Might be better to do the lock update separately.

James300: You're right, this also made it pull from the internal repo 😬

PR #679, opened 2019-11-29
Link - set rel noopener when target is blank

bmansell:
Wanted to make it easier to ensure safe cross-origin navigation using the Link component.

- Added conditional `rel` value `noopener` when `target='_blank'`
- Added new test to assert value is added
- Update snapshots

Sources for reading more about this topic and specific attribute value:
- https://mathiasbynens.github.io/rel-noopener/
- https://developers.google.com/web/tools/lighthouse/audits/noopener

### DIFF
--- a/packages/core/src/Link.js
+++ b/packages/core/src/Link.js
@@ -1,7 +1,9 @@
 import styled from 'styled-components'
 import { applyVariations, getPaletteColor, deprecatedColorValue } from './utils'
 
-const Link = styled.a`
+const Link = styled.a.attrs(props => ({
+  rel: props.target === '_blank' ? 'noopener' : null
+}))`
   cursor: pointer;
   text-decoration: none;
   color: ${getPaletteColor('base')};
--- a/packages/core/test/__snapshots__/Breadcrumbs.js.snap
+++ b/packages/core/test/__snapshots__/Breadcrumbs.js.snap
@@ -58,6 +58,7 @@ exports[`Breadcrumbs renders basic 1`] = `
       className="c2"
       color="text.light"
       href="https://www.priceline.com"
+      rel={null}
     >
       Home
     </a>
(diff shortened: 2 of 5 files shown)


packages/core/src/Link.js
```diff
@@ -13,6 +14,12 @@ const Link = styled.a`
   ${applyVariations('Link')}
 `
 
+const Link = props => {
+  const rel = props.target === '_blank' ? 'noopener' : null
```
James300: nitpick: destruct the props to `{ target }`.
Not so nitpick: pass remaining props to the component like `<StyledLink rel={rel} {...props} />` otherwise all props will be dropped and links probably won't work.

packages/core/test/__snapshots__/Breadcrumbs.js.snap
```diff
@@ -73,11 +57,8 @@ exports[`Breadcrumbs renders basic 1`] = `
   >
     <a
       className="c2"
-      color="text.light"
-      href="https://www.priceline.com/flights/"
```
James300: Here is an example of the carnage that would ensue if my other comment isn't addressed 😛

packages/core/src/Link.js
bmansell: Oof, my bad, definitely should have noticed that with the updated snapshots. Fixed and fixed!

packages/core/test/__snapshots__/Breadcrumbs.js.snap
bmansell: Haha, that'll do it 😅 Fixed!

packages/core/src/Link.js
```diff
@@ -13,6 +14,13 @@ const Link = styled.a`
   ${applyVariations('Link')}
 `
 
+const Link = props => {
```
achoboter: nb: you can also represent this like:
```
const Link = styled.a.attrs(props => ({
  rel: props.target === '_blank' ? 'noopener' : null
}))`
  cursor: pointer;
  text-decoration: none;
  ${applyVariations('Link')}
`
```

bmansell: Oh that's way cleaner, thank you! Replacement has been pushed up.

PR #680, opened 2019-12-02
fix(label-width): prevent label width overflow

jes708:
Before:
Screen Shot 2019-12-02 at 9 53 38 AM

Screen Shot 2019-12-02 at 9 47 42 AM

After:
Screen Shot 2019-12-02 at 9 53 12 AM

Screen Shot 2019-12-02 at 9 52 27 AM

### DIFF
--- a/packages/core/src/FormField.js
+++ b/packages/core/src/FormField.js
@@ -32,12 +32,14 @@ const FormField = props => {
     return child
   })
 
+  const ml = iconBefore ? '40px' : '12px'
+
   const styledLabel =
     label &&
     React.cloneElement(label, {
       htmlFor: label.props.htmlFor || id,
       fontSize: 10,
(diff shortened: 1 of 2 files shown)


packages/core/src/FormField.js
```diff
@@ -47,7 +49,8 @@ const FormField = props => {
         transitionDuration: '.1s',
         opacity: showLabel ? 1 : 0,
         pointerEvents: 'none',
-        position: 'relative'
+        position: 'relative',
+        width: `calc(100% - ${ml})`
```
dansternfeld1: (n-b) Any perf issues from this calc? We render many form fields on some pages.

jes708: I looked into this concern a bit. This will not have a noticeable impact on performance, even when used in bulk.

PR #685, opened 2019-12-02
Next v3 icon a11y tests

sdalonzo:


packages/icons/svgr.config.js
```diff
@@ -48,6 +54,12 @@ module.exports = {
     viewBox: '0 0 24 24',
     height: '{size}',
     width: '{size}',
+    title: '{title}',
+    titleId: '{titleId}',
+    desc: '{desc}',
+    descId: '{descId}',
+    'aria-labelledby': '{ariaLabelledBy}',
```
byakimchuk: Just double checking that the `aria` keys should be wrapped in quotes?

sdalonzo: In this case and most others, yes - it's a key in an object literal - the hyphen would make it invalid syntax otherwise.

PR #687, opened 2019-12-03
v3 of all packages

sdalonzo:
This is a rebase of https://github.com/priceline/design-system/pull/644


packages/icons/README.md
```diff
@@ -10,11 +10,12 @@ npm i pcln-icons
 
 ```js
 import React from 'react'
-import FlightsIcon from 'pcln-icons/lib/Flights'
+import { Flights as FlightsIcon } from 'pcln-icons'
 
 export default props => <FlightsIcon mr={2} />
 ```
 
+// TODO: revisit this
```
unscsprt: Do we want to leave this TODO?

packages/core/src/Banner.js
```diff
@@ -2,48 +2,56 @@ import React from 'react'
...
-import Heading from './Heading'
 import PropTypes from 'prop-types'
 import styled, { withTheme } from 'styled-components'
-import { applyVariations, hasPaletteColor, deprecatedColorValue } from './utils'
+import {
+  Attention as AttentionIcon,
+  Information as InformationIcon,
+  Success as SuccessIcon,
+  Warning as WarningIcon
+} from 'pcln-icons'
```
jes708: nb: This doesn't need to be changed but was wondering for the sake of backwards compatibility, is it still possible to import from `pcln-icons/lib`?

packages/popover/package.json
```diff
@@ -1,20 +1,33 @@
 {
   "name": "pcln-popover",
-  "version": "2.0.5",
+  "version": "3.0.0",
   "description": "React component for popover",
-  "main": "dist/index.cjs.js",
-  "module": "dist/index.esm.js",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "sideEffects": false,
```
unscsprt: Is the `sideEffects` key still needed now that we removed rollup?

packages/slider/package.json
```diff
@@ -1,19 +1,32 @@
 {
   "name": "pcln-slider",
-  "version": "2.2.0",
+  "version": "3.0.0",
   "description": "React component for range sliders",
-  "main": "dist/index.cjs.js",
-  "module": "dist/index.esm.js",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "sideEffects": false,
```
unscsprt: Same comment as above

plop-templates/package/package.json.hbs
```diff
@@ -1,12 +1,21 @@
 {
   "name": "pcln-{{kebabCase componentName}}",
   "version": "0.0.1",
-  "main": "dist/index.js",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "sideEffects": false,
```
unscsprt: `sideEffects` still needed?

plop-templates/package/package.json.hbs
```diff
@@ -1,12 +1,21 @@
...
   "scripts": {
-    "prepare": "babel src -d dist",
-    "clean": "rm -rf node_modules dist",
+    "prepare": "npm run clean && npm run build",
```
unscsprt: N-b; could be `npm-run-all clean build`

packages/core/test/Hug.js
```diff
@@ -44,7 +45,7 @@ describe('Hug', () => {
 
   test('renders text when array of nodes is provided', () => {
     const json = rendererCreateWithTheme(
-      <Hug icon="thumbsUp" text={[text, text]}>
+      <Hug icon={<ThumbsUp />} text={[text, text]}>
```
jes708: Is there a method for adding an icon to Hug that will work for both v2 and v3?

MIGRATING_to_v3.md
```diff
@@ -0,0 +1,70 @@
...
+/>
+```
+
+- Components that now take an `icon` node instead of an `iconName` string
+
+  - Icon
+  - Hug
+  - Banner
+  - Step
+  - Select
```
jes708: Passing in an icon node should also be possible in v2 for most of these components, otherwise I'm not sure v3 will be backward compatible with v2.

jes708: I think IconButton can be added to this list as well

MIGRATING_to_v3.md
```diff
@@ -0,0 +1,70 @@
...
+- Image now supports responsive widths via `width` prop
+- Link now uses dark shade of `color` prop on hover
+- Banner now supports a node as its `header` prop in addition to strings
+- Popover now supports trapping focus inside of the popup via `trapFocus` prop
+- Input and TextArea now use `borders` instead of `box-shadow`, to prevent misalignments with other components like Select
+
+## New Components
+
+- Avatar
+- Breadcrumbs
```
jes708: These aren't actually new anymore since they're in v2

MIGRATING_to_v3.md
```diff
@@ -0,0 +1,70 @@
...
+
+- Text
+  - Added `textShadowSize` and `enableTextShadow` props
+  - Changed default fontSize behavior of Heading to be more inline with Heading styles
+
+### Icons (`pcln-icons`)
+
+- `main` and `module` fields in `package.json` now point to CommonJS (CJS) and ES Modules (ESM) builds, respectively.
+- Usage: `import { TrendingUp, Timer } from 'pcln-icons'`
+  - Imports from `pcln-icons/lib/` are no longer available. This previously contained a CJS build that was incompatible with tree-shaking and dead code elimination in bundlers.
```
jes708: Ah this answers my question from before.  So if a host app consumes two common components, one which uses DS v2, and one which uses DS v3, what is the proper way to use pcln-icons?

packages/core/src/Banner.js
James300: I think that would destroy a lot of the tree shaking benefits

packages/core/test/Hug.js
James300: I don't think so, this is our major breaking change that the whole v3 release was designed around.

packages/core/src/Banner.js
jes708: I found the note in the migration docs.  See my comment there for further discussion.

MIGRATING_to_v3.md
James300: The goal isn't for it to be backwards compatible. Unfortunately some of the changes required were breaking for v2, which is why this needed to be a v3 bump and couldn't be a bunch of minor bumps.

MIGRATING_to_v3.md
James300: Ideally, the common components shouldn't have the DS as a dependency but as a peerDependency which means the host app's version will be the only one included. However, this does mean that common components will have to update to use the new methods of v3 before the host app can update.

packages/popover/package.json
James300: I think so because the host app's webpack will look at `sideEffects` while building to determine whether or not the module can be tree shaken.

packages/slider/package.json
James300: Same as above too 😝

MIGRATING_to_v3.md
jes708: If there is no longer a pcln-icons/lib, could pcln-icons v3 have a lib folder that is an actually just an alias that points back to the root?

James300: That definitely seems feasible but at what point do we stop doing that? With a v4 release do we remove lib and now people have to do update everything or do we force people to update now???

packages/core/test/Hug.js
sdalonzo: There's no reason you couldn't pass `<Icon name="ThumbsUp">`, it just won't be tree-shakable.

MIGRATING_to_v3.md
sdalonzo: You should definitely be able to pass an icon node.

MIGRATING_to_v3.md
sdalonzo: Due to the breaking changes, I don't think there is a way to use both in one application.

sdalonzo: To @James300 's point, maybe it would be better not to create a `/lib` in DS itself, but instead use Webpack aliases or a similar mechanism to point to `/cjs` instead.

packages/icons/README.md
sdalonzo: Nope, I will update the documentation here. I added this stub before we were sure that `Icon` in core would be removed.

packages/popover/package.json
sdalonzo: My understanding is yes, for exactly the reason @James300 stated.

plop-templates/package/package.json.hbs
sdalonzo: Yes. See above.

packages/icons/README.md
sdalonzo: I've pushed a commit that updates this section. Please re-review, and resolve this conversation if it is satisfactory.

plop-templates/package/package.json.hbs
sdalonzo: I've pushed this change. Please resolve if it is satisfactory.

packages/core/test/Hug.js
jes708: If you're passing an icon node that means you're on v3 of design-system, so there would be no need to pass Icon at that point.

MIGRATING_to_v3.md
sdalonzo: I removed this section since it no longer applies.

packages/core/test/Hug.js
sdalonzo: If we decide it's worth the effort, perhaps those changes can be backported - so the components in question will prefer the `icon` prop if provided but fall back to `iconName`.

packages/core/src/Icon.js
```diff
@@ -1,3 +1,3 @@
-import Icon from 'pcln-icons'
```
abautist: I thought `<Icon />` is being deprecated in v3?

packages/icons/README.md
```diff
@@ -10,22 +10,22 @@ npm i pcln-icons
...
 export default props => <FlightsIcon mr={2} />
 ```
 
 For a complete list of all icons, see: [the iconography docs](https://priceline.github.io/design-system/iconography)
 
-For backwards compatibility, the `Icon` component can be used in the same way as in the Design System v1.
-
-**Note:** The `Icon` component will be deprecated. Prefer using individual imports as shown above.
+**BREAKING CHANGE:** The `Icon` component has been removed from `core` (`pcln-design-system`) in v3.0.0. The optimal
+approach is to use named imports from `pcln-icons`. If you need an `<Icon>` component that supports a `name`
```
abautist: Related to my previous comment - are we still technically allowing consumers to use `<Icon />` after v3 is published? Why do we want to support both? It feels a bit confusing.

sdalonzo: @abautist I agree, but it needs to exist for cases like amenities where the names of the icons come from GraphQL. `<Icon />` will only exist in `pcln-icons`, it has been removed from `core`

sdalonzo: It breaks tree-shaking Icons for the app, so it is less than ideal, but until teams have a chance to work around that (perhaps importing a limited set of icons), I didn't want to make the migration effort to v3 any more complicated than it needs to be.

Issue #690, opened 2019-12-06
Add link style variation to Button component


abautist:
We would like to standardize a button variation that is _styled_ like a link and keep the accessibility standards of a button.

abautist:
We also would want a Link variation that is styled like a button https://github.com/priceline/design-system/issues/691

abautist:
Resolved by https://github.com/priceline/design-system/pull/692

PR #692, opened 2019-12-06
feat(buttonlink): add link variation to button component

abautist:
* button variation styled as a link
* addresses https://github.com/priceline/design-system/issues/690


packages/core/src/Button.js
```diff
@@ -65,6 +65,17 @@ const variations = {
         box-shadow: inset 0 0 0 2px ${getPaletteColor('dark')(props)};
       `}
     }
+  `,
+  link: css`
+    color: ${getPaletteColor('base')};
+    line-height: ${props => props.theme.lineHeights.standard};
```
abautist: i added this only for this variation because `line-height` for `Button` was set to 1.5. I wonder if we can set the `Button` line-height to 1.4 which would affect all variations.

sdalonzo: @wesleymartin85 what do you think about Auggie's suggestion of changing the `line-height` on `Button`?

docs/pages/Button.md
```diff
@@ -47,6 +47,12 @@ Buttons make common actions immediately visible and easy to perform with one cli
 <Button variation="outline">Outline Button</Button>
 ```
 
+### link
+
+```.jsx
+<Button variation="link">Outline Button</Button>
```
sdalonzo: `Link Button`

packages/core/test/Button.js
```diff
@@ -54,6 +56,28 @@ describe('Button', () => {
     expect(getByTitle(title)).toHaveAttribute('aria-label', title)
   })
 
+  describe('variations', () => {
+    describe('link variation', () => {
```
sdalonzo: very nice!

packages/core/src/Button.js
wesleymartin85: @sdalonzo @abautist this could be helpful for when text as a button is inline with copy. But I know there are more use cases to think about. I started to look into this for all (most) scenarios. I think the next step is to review in a larger effort with a few designers and some devs to make sure we are all aligned and educated on the differences in buttons and links.

docs/pages/Button.md
abautist: Updated.

packages/core/src/Button.js
abautist: Ok cool, I've kept the line height change as is so as not to impact other Button variations for now and we can discuss as a group for long-term alignment on line-height for buttons.

PR #707, opened 2020-01-14
Replace lifecycle as unsafe

suyashchopra19:
This pr attempts to mark the LifeCycleMethod `componentWillReceiveProps` as `UNFASE_componentWillReceiveProps`

### DIFF
--- a/packages/popover/package.json
+++ b/packages/popover/package.json
@@ -12,18 +12,23 @@
   "author": "Priceline",
   "license": "MIT",
   "jest": {
-    "coverageReporters": ["lcov", "html"],
-    "setupFilesAfterEnv": ["../../test-setup.js"],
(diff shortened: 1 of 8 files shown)


packages/popover/package.json
```diff
@@ -12,25 +12,31 @@
   "author": "Priceline",
   "license": "MIT",
   "jest": {
-    "coverageReporters": ["lcov", "html"],
-    "setupFilesAfterEnv": ["../../test-setup.js"],
-    "testMatch": ["<rootDir>/test/**/*.js"]
-  },
-  "dependencies": {
-    "react-popper": "^1.3.3"
+    "coverageReporters": [
```
sdalonzo: there shouldn't be any change in this file

PR #712, opened 2020-01-21
Add parallax variation to BackgroundImage

bmansell:
- Add variation for parallax to the component `BackgroundImage`
- Update util function `applyVariations` to allow for variations other than color
- Add a test and storybook example, update snapshot

##### Storybook
background-image

### DIFF
--- a/packages/core/src/BackgroundImage.js
+++ b/packages/core/src/BackgroundImage.js
@@ -1,30 +1,47 @@
-import styled from 'styled-components'
+import styled, { css } from 'styled-components'
 import PropTypes from 'prop-types'
 import { width } from 'styled-system'
-import { getPaletteColor } from './utils'
+import { applyVariations, getPaletteColor, mapProps } from './utils'
+
+const variations = {
+  parallax: css`
+    background-attachment: fixed;
+  `,
+  static: css`
+    background-attachment: scroll;
+  `
+}
 
 const image = props =>
   props.image ? { backgroundImage: `url(${props.image})` } : null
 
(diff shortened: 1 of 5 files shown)


packages/core/src/BackgroundImage.js
```diff
@@ -1,30 +1,46 @@
...
+  ${applyVariations('BackgroundImage', variations)}
+`)
+
+BackgroundImage.defaultProps = {
+  variation: 'static'
+}
 
 BackgroundImage.propTypes = {
-  /** background-image url */
+  backgroundImage: PropTypes.string,
```
sdalonzo: what's this for?

packages/core/test/BackgroundImage.js
```diff
@@ -20,4 +20,16 @@ describe('BackgroundImage', () => {
...
   })
+
+  describe('variations', () => {
+    describe('parallax variation', () => {
+      test('default render', () => {
+        const json = rendererCreateWithTheme(
+          <BackgroundImage variation="parallax" image="hello.png" />
+        ).toJSON()
+        expect(json).toMatchSnapshot()
+        expect(json).toHaveStyleRule('background-attachment', 'fixed')
```
sdalonzo: 👍

packages/core/src/BackgroundImage.js
bmansell: I inferred from the comment that this snippet was missing a proptype for `backgroundImage` (used above in this file) but I misread it. Not sure what the comment is for but I added it back and added a prop type for `height` instead!

PR #727, opened 2020-03-18
Fix style bug in ModalHeader

craigpalermo:
Noticed this when installing `pcln-modal` in a host app using `styled-components@4` and `design-system@2`.

Before:
Screen Shot 2020-03-18 at 11 19 16 AM

After:
Screen Shot 2020-03-18 at 11 24 24 AM

Tested in Storybook


craigpalermo:
@steve-dalonzo reverted changes to `package-lock`

PR #729, opened 2020-03-20
Document breaking changes to `<Hide>`

sdalonzo:
Adding the `xxl` breakpoint changed the behavior of `<Hide xl>` and needs to be documented.


bertya:
LGTM

PR #731, opened 2020-03-23
Fix compatibility with SC3 and SC4

craigpalermo:
`popover` wasn't previously compatible with SC >3, so this enables components that use it to upgrade to SC4.

### DIFF
--- a/packages/popover/package.json
+++ b/packages/popover/package.json
@@ -28,7 +28,8 @@
       "../../test-setup.js"
     ],
     "testMatch": [
-      "<rootDir>/test/**/*.js"
+      "<rootDir>/test/**/*.js",
+      "<rootDir>/src/**/*.spec.js"
     ]
   },
   "dependencies": {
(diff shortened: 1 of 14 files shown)


packages/popover/src/Overlay.js
```diff
@@ -33,6 +32,7 @@ const Overlay = ({
       popoverOpen={popoverOpen}
       overlayOpacity={overlayOpacity}
       zIndex={zIndex}
+      bg={'background.darkest'}
```
craigpalermo: @wesleymartin85 do we have a replacement for `theme.colors.black` in the palette? `background.darkest` is close, but not identical.

packages/popover/package.json
```diff
@@ -28,7 +28,8 @@
       "../../test-setup.js"
     ],
     "testMatch": [
-      "<rootDir>/test/**/*.js"
+      "<rootDir>/test/**/*.js",
+      "<rootDir>/src/**/*.spec.js"
```
sdalonzo: 👍

packages/popover/src/Overlay.js
sdalonzo: n-b don't need curly braces for strings

PR #752, opened 2020-04-17
Small Badge and Rename Stamp Variations and Size

### LINKED ISSUE #738: New Variation: create "small" badge variation
here is a link to the [InVision Project](https://priceline.invisionapp.com/share/R2WMAJ03ABU)

Screen Shot 2020-04-03 at 12 40 43 PM

pcln-james:
Closes #738 

- Added Badge small size
- Renamed Stamp regular size to medium
- Renamed Stamp regular variation to outline to be consistent with Button
- Renamed Stamp solid variation to fill to be consistent with Button
- Fixed border-color issue with fill variation of the Stamp

- Added Cartesian Storybook for Badge
image

- Added Cartesian Storybook for Stamp
image

### DIFF
--- a/packages/core/src/Stamp.js
+++ b/packages/core/src/Stamp.js
@@ -43,6 +43,12 @@ const variations = {
     color: ${getTextColorOn('base')};
     background-color: ${getPaletteColor('base')};
     border-color: ${getPaletteColor('base')};
+  `,
+  // todo: remove this copy of the fill variation in v4 as its name does not follow conventions
+  solid: css`
+    color: ${getTextColorOn('base')};
+    background-color: ${getPaletteColor('base')};
+    border-color: ${getPaletteColor('base')};
   `
 }
 
--- a/packages/core/src/Stamp.js
+++ b/packages/core/src/Stamp.js
@@ -19,7 +19,7 @@ const sizes = {
       height: 12px;
     }
   `,
-  regular: css`
+  medium: css`
     ${themeGet('textStyles.display0')}
     line-height: ${themeGet('lineHeights.standard')};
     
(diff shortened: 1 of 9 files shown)


packages/core/storybook/Stamp.js
```diff
@@ -12,51 +34,18 @@ storiesOf('Stamp', module)
...
-      </Stamp>
-      <Stamp color="primary" mr={2}>
+    <Cartesian
+      component={Stamp}
+      color={Object.keys(colors)}
+      variation={Object.keys(variations)}
+      size={Object.keys(sizes)}
+      m={3}
+    >
+      <>
```
craigpalermo, line 45: Is the fragment needed here?

packages/core/src/Stamp.js
```diff
@@ -31,17 +31,18 @@ const sizes = {
...
+  outline: css`
     color: ${getPaletteColor('base')};
     border-color: ${props => getPaletteColor(props.borderColor, 'base')(props)};
     ${props =>
       props.bg
         ? `background-color: ${getPaletteColor(props.bg, 'base')(props)};`
         : ''}
   `,
-  solid: css`
+  fill: css`
```
craigpalermo, line 42: Would this be a breaking change for existing usages of `Stamp`?

packages/core/storybook/Stamp.js
pcln-james: Yes, I guess Cardinal splits all the children into separate examples which means to have the examples as a single component you need just one child.

packages/core/src/Stamp.js
pcln-james: Unfortunately yes, but we only added the version the other day, we could remove that release?

craigpalermo: I think unpublishing that release would probably be the best course of action, rather than immediately making another major. @sdalonzo, thoughts?

sdalonzo: could we alias it to the old name temporarily, and remove it in v4?

PR #753, opened 2020-04-20
Add button variations to the link component

### LINKED ISSUE #691: Add button style variation to Link component
We would like to standardize a Link variation that is _styled_ like a button and keep the accessibility standards of a link.

pcln-james:
Closes #691 

image

### DIFF
--- a/packages/core/src/Link.js
+++ b/packages/core/src/Link.js
@@ -1,29 +1,56 @@
-import styled from 'styled-components'
-import { applyVariations, getPaletteColor, deprecatedColorValue } from './utils'
+import PropTypes from 'prop-types'
+import styled, { css } from 'styled-components'
+import { width, space } from 'styled-system'
+import {
+  mapProps,
+  applyVariations,
+  getPaletteColor,
+  deprecatedColorValue
+} from './utils'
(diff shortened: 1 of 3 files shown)


packages/core/src/Link.js
```diff
@@ -1,29 +1,56 @@
...
+}
 
-  ${applyVariations('Link')}
-`
+const Link = mapProps(({ target, ...props }) => ({
+  rel: target === '_blank' ? 'noopener' : null,
+  ...props
+}))(styled.a`
+  ${width} ${space};
+  ${buttonStyles}
```
pcln-james, line 40: How does everybody feel about export the button styles, including variations and including them here?

craigpalermo: My only concern is that it makes Link dependent upon another core component, but I'm not sure what implications that might have.

pcln-james: True but is it worse to have duplicated code which might become unsynced

craigpalermo: Good point. I think this would make sense then. I'm generally in favor of DRY; if we run into issues, we could always pull the shared styles into a module and import them from Button/Link, but I doubt we'll need to.

PR #756, opened 2020-04-21
Forward refs to DS core components

### LINKED ISSUE #479: Add Ref Forwarding to core components
It looks like we are currently not forwarding refs on our core components. This can be problematic for accessibility purposes, since it won't allow us to do things like restore focus to the element by using refs.

React introduced `forwardRef` API in v16.3, so this should make it pretty easy to update our core components. 

https://reactjs.org/blog/2018/03/29/react-v-16-3.html#forwardref-api

```js
const FancyButton = React.forwardRef((props, ref) => (
  <button ref={ref} className="FancyButton">
    {props.children}
  </button>
));

// You can now get a ref directly to the DOM button:
const ref = React.createRef();
<FancyButton ref={ref}>Click me!</FancyButton>;
```

The thing we should probably discuss as a group is which core components should receive this update and why?

I think for the sake of accessibility, the following components should be forwarding refs:
- Button
- Link
- Input
- Radio
- Select
- Slider
- Tooltip

Any others we can think of?

craigpalermo:
Resolves https://github.com/priceline/design-system/issues/479

Testing done:
- Dogfooded in projects using beta tag and SC3/4


packages/core/src/IconButton.js
```diff
@@ -20,16 +22,24 @@ const TransparentButton = styled(Button)`
...
-const IconButton = ({ icon, ...props }) => (
-  <TransparentButton {...props}>
-    <div>{icon}</div>
-  </TransparentButton>
-)
+const IconButton = ({ icon, ...props }) => {
+  // TODO remove once migrated to SC4
+  const spreadProps = props
+  spreadProps[getSCMigrationRef()] = props.dsRef
+  delete spreadProps.dsRef
```
pcln-james, line 29: Would this be nicer?
```
const spreadProps = {
  ...props,
  [getSCMigrationRef()] = props.dsRef,
}
```
And is there any reason to delete dsRef?

packages/core/src/utils/getSCMigrationRef/index.js
```diff
@@ -0,0 +1,2 @@
+import getSCMigrationRef from './getSCMigrationRef'
+export default getSCMigrationRef
```
pcln-james, line 2: How do you feel about `export { default } from './getSCMigrationRef'`?

packages/core/src/utils/refPropType/index.js
```diff
@@ -0,0 +1,2 @@
+import refPropType from './refPropType.js'
+export default refPropType
```
pcln-james, line 2: Same as above

packages/core/src/utils/getSCMigrationRef/index.js
craigpalermo: Oh is that available? Wasn't sure if we could use `export default from`

packages/core/src/IconButton.js
craigpalermo: It didn't seem like it, but I wanted to avoid confusing people by passing the same value as different props.

packages/core/src/BlockLink.js
```diff
@@ -1,12 +1,22 @@
 import styled from 'styled-components'
 import Link from './Link'
+import { mapProps } from './utils'
+import getSCMigrationRef from './utils/getSCMigrationRef'
```
sdalonzo, line 4: why not export all of these from utils/index?

packages/core/src/utils/getSCMigrationRef/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './getSCMigrationRef'
```
sdalonzo, line 1: why not attach these to `utils/index.js` so you can use one destructured import for utils?

packages/core/test/__snapshots__/BlockLink.js.snap
```diff
@@ -5,23 +5,19 @@ exports[`BlockLink renders 1`] = `
   cursor: pointer;
   -webkit-text-decoration: none;
   text-decoration: none;
-  color: #007aff;
```
sdalonzo, line 8: ⁉️

packages/core/src/utils/getSCMigrationRef/index.js
craigpalermo: I don't feel strongly about doing it this way, but at some point we might want to consider breaking the utils out into separate modules if that file gets too bloated. I'll move the new helper in there to stick to the current pattern.

packages/core/test/__snapshots__/BlockLink.js.snap
craigpalermo: Not sure exactly why this changed, I'll have to look into it.

packages/core/src/BlockLink.js
craigpalermo: Updated, agreed that importing from the same utils looks better

packages/core/test/__snapshots__/BlockLink.js.snap
craigpalermo: @sdalonzo If I'm reading this correctly, `color: inherit;` at line 21 would have already been overriding this color line, so this deletion shouldn't have any real impact. I verified `Link` and `BlockLink` again in Storybook and everything looks as expected.

craigpalermo:
@sdalonzo FYI the snapshot changes to Link and related components are a result of updating with @James300's button variation changes.

sdalonzo (approved):

dgiraldo313 (approved):

PR #758, opened 2020-04-21
Modal Timeout Prop

### LINKED ISSUE #757: Reduce / Remove Delayed Modal Open
The Priceline Modal opens slowly, I suspect it's due to the timeout prop that exists on the transition component. We should test that this hypothesis is correct and if so create a timeout prop with a default of 500 (the current value) that can be changed to reduce this delay.

pcln-james:
Closes #757 

Tested and confirmed that the timeout on the Transition component was causing the open delay.

### DIFF
--- a/packages/modal/src/Modal.js
+++ b/packages/modal/src/Modal.js
@@ -145,11 +145,12 @@ const Modal = ({
   disableCloseButton,
   enableOverflow,
   height,
+  timeout,
   dialogAnimation,
   overlayAnimation,
   verticalAlignment
 }) => (
-  <Transition in={isOpen} unmountOnExit timeout={500}>
+  <Transition in={isOpen} unmountOnExit timeout={timeout}>
     {state => (
       <Overlay
         onDismiss={onClose}
@@ -206,6 +207,7 @@ Modal.defaultProps = {
   header: null,
   bg: 'white',
   height: 420,
+  timeout: 500,
   overlayAnimation: null,
   dialogAnimation: null,
   verticalAlignment: 'middle'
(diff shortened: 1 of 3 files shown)


packages/modal/storybook/Modal.js
```diff
@@ -115,3 +115,10 @@ storiesOf('Modal', module)
       verticalAlignment="top"
     />
   ))
+  .add('Zero Timeout', () => (
```
craigpalermo, line 118: If the point of this is to remove the animation of the transition, we may want to make the story more descriptive, like, "Use 0 timeout to open/close dialog without animated transition"

packages/modal/src/Modal.js
```diff
@@ -225,6 +227,7 @@ Modal.propTypes = {
   imgMode: PropTypes.bool,
   className: PropTypes.string,
   header: PropTypes.any,
+  timeout: PropTypes.number,
```
craigpalermo, line 230: I think this would be more appropriately named something like `animationDuration`; `timeout` makes me think the modal closes after this amount of time, or something.

packages/modal/storybook/Modal.js
pcln-james: The animation still happens but there is no delay before it starts, that's what timeout does

packages/modal/src/Modal.js
pcln-james: Timeout is a little confusing, it doesn't control the duration of the animation but the initial delay before the animation begins

craigpalermo: Ohhh gotcha, this makes sense then. Can you add a doc comment for that prop to avoid confusion?

PR #760, opened 2020-04-28
fix: Re-add target prop to link

bmansell:
- The `target` prop was not getting set on the `Link` component, was revealed when updating the snapshots for tests
- Added `target` back and updated snapshots
- Tweaked import for `buttonStyles` in `Link` so that storybook builds successfully (see screenshot for error)

Screen Shot 2020-04-28 at 10 51 50 AM


bmansell:
Duplicate.

PR #761, opened 2020-04-28
fix(720): Modify popover aria label

### LINKED ISSUE #720: Pcln-popover aria-label hard coded
Line 66 of Pcln-popover hard codes the aria-label:
Screen Shot 2020-03-02 at 11 12 36 AM

Ideally, we would allow a more flexible aria-label to be passed to avoid accessibility errors such is this currently being called out in our storybook:
Screen Shot 2020-03-02 at 11 12 59 AM

This would allow the popover to truly wrap any child content and avoid this accessibility issue.

bmansell:
- Move `aria-label` prop to be set on the button instead of the content
- Add assertion for popover aria label text in test
- Change spelling of `PopOver` to `Popover`

Closes #720


packages/popover/storybook/Popover.js
```diff
@@ -118,7 +118,7 @@ const Playground = ({ children }) => (
 const InnerContent = ({ handleClose }) => (
   <ThemeProvider>
     <Box p={4}>
-      <Absolute color={'blue'} top={25} right={25}>
+      <Absolute color='blue' top={25} right={25}>
```
sdalonzo, line 121: semantic name please?

bmansell: Sure, fixed.

PR #770, opened 2020-05-04
Lazy Loading by Default

### LINKED ISSUE #142: Consider add image optimization for the Image and ImageBackground component
A couple of points we should consider working on:
- Make Images take an array of sources for different screen size
- Make Images progressive
- Support lazy loading

pcln-james:
Closes #142 

https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading

Chrome is already defaulting to lazy load so this doesn't change existing behaviour but it adds the ability to change it if desired.

image

### DIFF
--- a/packages/core/src/Image.js
+++ b/packages/core/src/Image.js
@@ -2,7 +2,8 @@ import styled from 'styled-components'
 import { width } from 'styled-system'
 
 const Image = styled.img.attrs(props => ({
-  height: props.height || 'auto'
+  height: props.height || 'auto',
+  loading: props.loading || 'lazy'
 }))`
   display: block;
   max-width: 100%;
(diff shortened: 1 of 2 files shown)


craigpalermo:
I think this is fine as a progressive enhancement for a minor version; the only argument against that would be that _technically_ this changes the default behavior for browsers that support the `loading` attribute. To stick closer to semver rules, should we make this opt-in?

craigpalermo:
> @craigpalermo Chrome, Firefox and Edge have it enabled by default (`loading="lazy"`) so this actually does stick to semver but if we want to stick even closer to it we could have the default value be null and then the attribute would be avoided and use the browser's default as it currently does.

I like that idea; if it seems like it's doing well then we could probably default it to use `loading` in V4.

pcln-james:
> > @craigpalermo Chrome, Firefox and Edge have it enabled by default (`loading="lazy"`) so this actually does stick to semver but if we want to stick even closer to it we could have the default value be null and then the attribute would be avoided and use the browser's default as it currently does.
> 
> @pcln-james I like that idea; if it seems like it's doing well then we could probably default it to use `loading` in V4.

@craigpalermo Turns out this can just be closed as the current implementation supports this lol

PR #771, opened 2020-05-05
fix(721): Remove hardcoded popover theme

### LINKED ISSUE #721: pcln-popover hardcoded theme
Using the imported theme from DS causing the popover to not inherit themes provided upstream causing users of the popover to wrapper an additional theme provider. We should remove all uses of `theme` from Popover and move towards using `color` instead of `bg`/`background` to ensure proper semantic naming

bmansell:
- Remove all instances of hardcoded popover theme
- Remove prop `bg`/`background` from being passed into styles for `Arrow` and other inner components
- Substitute instances of `bg`/`background` with `color`
- Allow for specifying `borderColor` but will default to `border.base` if none is provided
- Added storybook example of theme colored Popovers
- Updated snapshots

##### New `Colors` storybook example
Screen Shot 2020-05-06 at 11 28 01 AM

Screen Shot 2020-05-06 at 11 28 10 AM

Screen Shot 2020-05-06 at 11 21 40 AM

Screen Shot 2020-05-06 at 11 21 51 AM

##### Playbook (pre-existing storybook example)
Screen Shot 2020-05-06 at 11 18 48 AM

Closes #721

### DIFF
--- a/packages/popover/src/Arrow.js
+++ b/packages/popover/src/Arrow.js
@@ -1,22 +1,21 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import { deprecatedPropType, theme } from 'pcln-design-system'
+import { deprecatedPropType, getPaletteColor } from 'pcln-design-system'
 
 const PopoverArrow = ({
   arrowProps,
   placement,
-  background,
   borderColor,
+  color,
   className
 }) => (
   <Arrow
     className={className}
     style={arrowProps.style}
     data-placement={placement}
-    theme={theme}
-    background={background}
     borderColor={borderColor}
+    color={color}
     aria-hidden="true"
   >
     <div ref={arrowProps.ref} />
@@ -53,9 +52,9 @@ const ArrowAlignment = () =>
     }
   `
 
-const ArrowPlacement = ({ background, borderColor }) => {
-  const bgColor = theme.colors[background]
-  const brColor = theme.colors[borderColor]
+const ArrowPlacement = ({ color, borderColor, ...props }) => {
+  const bgColor = getPaletteColor(color, 'base')(props)
+  const brColor = getPaletteColor(borderColor, 'base')(props)
 
   return `
     &[data-placement*="right"]::before {
@@ -82,6 +81,7 @@ const ArrowPlacement = ({ background, borderColor }) => {
     }
   `
 }
+
 const Arrow = styled.span`
   position: absolute;
   ${ArrowAlignment}
@@ -94,14 +94,13 @@ PopoverArrow.propTypes = {
     style: PropTypes.object
   }).isRequired,
   className: PropTypes.string,
+  color: PropTypes.string,
   bg: deprecatedPropType('color'),
   borderColor: PropTypes.string,
   placement: PropTypes.string
 }
 
 PopoverArrow.defaultProps = {
-  bg: 'white',
-  borderColor: 'borderGray',
   placement: 'top'
 }
 
--- a/packages/popover/src/PopoverContent.js
+++ b/packages/popover/src/PopoverContent.js
@@ -2,8 +2,9 @@ import React, { Component } from 'react'
 import ReactDOM from 'react-dom'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
+import { themeGet } from 'styled-system'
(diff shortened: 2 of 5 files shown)


pcln-james:
Based on the screenshots, the filled color popovers (success, alert, error) look a bit weird with the gray border, maybe we should make that the same color or just get rid of the border?

packages/popover/src/Arrow.js
```diff
@@ -53,9 +52,10 @@ const ArrowAlignment = () =>
     }
   `
 
-const ArrowPlacement = ({ background, borderColor }) => {
-  const bgColor = theme.colors[background]
-  const brColor = theme.colors[borderColor]
+const ArrowPlacement = ({ color, borderColor, ...props }) => {
+  let backgroundColorName = color.indexOf('.') !== -1 ? color : `${color}.base`
```
pcln-james, line 56: so this is to prevent somebody from doing `error.dark`?

pcln-james: never mind, I got my ternary backwards 🤦

bmansell: No, it's when something like `error` is passed as a color. The `getPaletteColor` usage in here isn't able to grab the corresponding background color from the palette to match the `Box` background color unless it's filled in manually here.

packages/popover/src/PopoverContent.js
```diff
@@ -131,12 +140,11 @@ const PopperGuide = styled(Box)`
   box-sizing: border-box;
 `
 const ContentContainer = styled.section`
-  box-shadow: 0 0 0 1px ${({ theme, borderColor }) => theme.colors[borderColor]},
+  box-shadow: 0 0 0 1px ${props => getPaletteColor(props.borderColor)},
```
pcln-james, line 143: I think this needs to be `${props => getPaletteColor(props.borderColor)(props)}`

bmansell:
> Based on the screenshots, the filled color popovers (success, alert, error) look a bit weird with the gray border, maybe we should make that the same color or just get rid of the border?

The thing is, the border is created with the box shadow style and it looks a bit strange without that, so I'm not too sure what to change there.

packages/popover/src/Arrow.js
bmansell: I can flip the condition around if it makes it easier to read though!

achoboter, line 56: instead you can do this:
```
getPaletteColor(color, 'base')(props)
```
https://priceline.github.io/design-system/utilities/getPaletteColor#palette-color-driven-from-other-prop

bmansell: Oh awesome, thank you!

bmansell: Fixed!

packages/popover/src/PopoverContent.js
bmansell: Fixed!

pcln-james:
> @pcln-james
> 
> > Based on the screenshots, the filled color popovers (success, alert, error) look a bit weird with the gray border, maybe we should make that the same color or just get rid of the border?
> 
> The thing is, the border is created with the box shadow style and it looks a bit strange without that, so I'm not too sure what to change there.
> 
> _Edit_: Updated the storybook example to set the `borderColor` prop to the same as the `color` and it looks a lot better:
> Screen Shot 2020-05-05 at 4 21 12 PM
> Screen Shot 2020-05-05 at 4 21 23 PM
> Screen Shot 2020-05-05 at 4 23 48 PM
> 
> Just not sure if having to set both is counter-intuitive to the purpose of `color` or not.
> 
> @sdalonzo what are your thoughts on this?

We could have the border color default to the value of the color prop if color is specified and border isn't?

bmansell:
> We could have the border color default to the value of the color prop if color is specified and border isn't?

@pcln-james Good point! I've tweaked the changes to the following:

- Remove the default props for `color` and `borderColor`
- Added a function to handle the logic for deciding which border color to use (`getBorderColorName`). 
- Changed the innermost `Box` component to be a styled component with a border radius value as the `ContainerComponent`'s border radius was getting obscured by the new color prop on `Box`. 

Screenshots have been updated in the original description! If you and @craigpalermo could take another peek that would be awesome, please and thanks!

PR #772, opened 2020-05-05
Create package for type declarations

### LINKED ISSUE #767: Technical design for hand-written TS declarations for core project
Before we start writing type declarations, we need to figure out how to structure them. Open questions from DS meeting 5/1:
- Do we use inheritance or some other strategy to structure our declarations?

### LINKED ISSUE #784: Define static properties for all core components
In order to expose the type declarations for the whole `pcln-design-system` package, even if we only have prop types defined for a handful of components, we'll need to declare all exports and their static properties. Otherwise, the TS compiler complains about missing properties:

```
ERROR in /Users/cpalermo/priceline-web/pcln-web/react-components/ts-react-component-demo/src/HorseFarm.tsx
./src/HorseFarm.tsx
[tsl] ERROR in /Users/cpalermo/priceline-web/pcln-web/react-components/ts-react-component-demo/src/HorseFarm.tsx(31,18)
      TS2339: Property 'h2' does not exist on type 'typeof Heading'.
```

If we leave out the export for a given component entirely, we get this error from TSC:

```
ERROR in /Users/cpalermo/priceline-web/pcln-web/react-components/ts-react-component-demo/src/HorseFarm.tsx
./src/HorseFarm.tsx
[tsl] ERROR in /Users/cpalermo/priceline-web/pcln-web/react-components/ts-react-component-demo/src/HorseFarm.tsx(3,25)
      TS2305: Module '"../../../../../../../Users/cpalermo/priceline-web/design-system/packages/core/dist/types"' has no exported member 'Text'.
```

craigpalermo:
This PR creates a new package to house type declarations for `pcln-design-system`, and eventually the non-core modules. During testing in a host app, I noticed that TSC gives failures for DS components that don't have a declaration, so we'll have to leave out `"types": "dist/types/index.d.ts" from `pcln-design-system` until types for all exports are fully defined.

Here's an incomplete demo of the kind of workflow enhancements enabled by this change demonstrated with `<Text />`:
design-system typescript demo 2020-05-25 17_28_01

Resolves #767, Resolves #784


sdalonzo:
> This pull request **introduces 2 alerts** when merging [5f78433](https://github.com/priceline/design-system/commit/5f78433b6e4b0d11048e7a46128c395bcf80fbfc) into [bc9f0d1](https://github.com/priceline/design-system/commit/bc9f0d19f23248c357eb7cb440d02faa5e276dab) - [view on LGTM.com](https://lgtm.com/projects/g/priceline/design-system/rev/pr-bd3faa6259ebc16aeb2122c62b9ce354bf48cc10)
> 
> **new alerts:**
> 
>     * 2 for Unused variable, import, function or class

@craigpalermo can we address this please?

craigpalermo:
> > This pull request **introduces 2 alerts** when merging [5f78433](https://github.com/priceline/design-system/commit/5f78433b6e4b0d11048e7a46128c395bcf80fbfc) into [bc9f0d1](https://github.com/priceline/design-system/commit/bc9f0d19f23248c357eb7cb440d02faa5e276dab) - [view on LGTM.com](https://lgtm.com/projects/g/priceline/design-system/rev/pr-bd3faa6259ebc16aeb2122c62b9ce354bf48cc10)
> > **new alerts:**
> > ```
> > * 2 for Unused variable, import, function or class
> > ```
> 
> @craigpalermo can we address this please?

Those are coming from the test file for types. I tried to model it on examples from the `DefinitelyTyped` where the tests aren't actually executed, just type-checked. We'll probably want to confirm that this is the testing strategy we want to take, then I'll look into fixing the warnings.

packages/core/package.json
```diff
@@ -68,6 +70,7 @@
   "dependencies": {
     "deepmerge": "^4.0.0",
     "hoist-non-react-statics": "^3.3.1",
+    "pcln-types": "^0.0.1",
```
huzefakagdi, line 73: I am assuming `pcln-types` is a project that will pulled in optionally separately by each project to get type declarations. So why do we need it as a dependency here? Should it not be a `devDependency` instead?

xifeng-jin-pcln: ☝️

SrikanthDarma: Why should each project separately install pcln-types, if this is a dependency here, then installing pcln-types is not required in other projects. @huzefakagdi @craigpalermo  Correct me if I understood this differently.

huzefakagdi: @SrikanthDarma types are always imported conditionally for all packages that do not export types directly on build time. Also, if you see the `prepublish` script task above, we are copying the types declarations directly in the `pcln-design-system` package. Which seems one more reason IMHO that this package may not be required as a `dependency`.

jeff-pcln: i think you want the option to install them because you may not have a typescript project and you may not care about getting the types. also, it's nice to a have a central place to view all your types imo.

huzefakagdi: That is my what I am thinking as well Jeff

craigpalermo: Yeah, I'm not sure if `pcln-types` should be a dependency or devDep of `pcln-design-system`. My thinking here was that the types would be installed in `node_modules`, then on `prepublish` we'd copy them into `dist/types` and reference them via the `"types"` property in `package.json`.

 I _think_ it'd just need to be a devDep because they're only needed when publishing; consumers of `pcln-design-system` would get the `types/` dir along with `cjs`, `esm`, and `es` in `dist/` and the `"types"` property in `package.json` would point to that.

jeff-pcln: +1 for `devDep`

packages/types/package.json
```diff
@@ -0,0 +1,39 @@
...
+  ],
+  "files": [
+    "types/index.d.ts"
+  ],
+  "homepage": "https://github.com/priceline/design-system",
+  "license": "MIT",
+  "types": "types",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/priceline/design-system.git"
```
sdalonzo, line 22:
```suggestion
    "url": "github:priceline/design-system",
    "directory": "packages/types"
```

packages/core/package.json
```diff
@@ -1,16 +1,18 @@
 {
   "name": "pcln-design-system",
-  "version": "3.4.1",
+  "version": "3.5.0",
```
craigpalermo, line 3: No version bump needed yet

packages/types/types/index.d.ts
```diff
@@ -0,0 +1,118 @@
...
+ * the component uses
+ */
+export interface BoxProps
+  extends LayoutProps,
+    ColorProps,
+    BackgroundColorProps,
+    SpaceProps {}
+
+/**
+ * Extend primitive Design System components to create interfaces more complex
```
craigpalermo, line 30: `interfaces for`

packages/types/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "scripts": {
+    "test": "dtslint types"
+  },
+  "bugs": {
+    "url": "https://github.com/priceline/design-system/issues"
+  },
+  "dependencies": {
+    "@types/react": "^16.9.34",
+    "@types/styled-system": "ts3.8"
```
pcln-james, line 33: Didn't know this was a thing

jeff-pcln:
@craigpalermo just a heads up, i discovered that `theme-get` was broken out into a separate package, at v5 i think, and that function wasn't included in the types i used. not really an issue, just more an fyi if you ever run into it.

PR #796, opened 2020-05-27
Responsive boxshadows card

amyhlee:
- **What does this PR do?**
Updated boxShadow function to take an array in addition to a string, for responsive views

- **Any background context you want to provide?**
- modified boxShadow function to take in element from array or string
- updated boxShadowSize propType to take in string or element from array
- updated Card.js in Storybook to reflect updated changes

- **Screenshot(s)**

### DIFF
--- a/packages/core/src/Card.js
+++ b/packages/core/src/Card.js
@@ -20,20 +20,20 @@ const boxShadow = props => {
   }
 
   return `
-    @media only screen and (min-width: 1200px)  {
-      box-shadow: ${boxShadows[props.boxShadowSize[3] || 'xl']}
+    @media only screen and (min-width: 520px) {
+      box-shadow: ${boxShadows[props.boxShadowSize[0] || 'sm']}
     }
-    ​
-    @media only screen and (max-width: 992px) {
-      box-shadow: ${boxShadows[props.boxShadowSize[2] || 'lg']}
-    }
-    ​
-    @media only screen and (max-width: 768px) {
+​
+    @media only screen and (min-width: 768px) {
       box-shadow: ${boxShadows[props.boxShadowSize[1] || 'md']}
     }
-
-    @media only screen and (max-width: 520px) {
-      box-shadow: ${boxShadows[props.boxShadowSize[0] || 'sm']}
+​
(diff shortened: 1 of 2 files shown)


packages/core/src/Card.js
```diff
@@ -6,20 +6,35 @@ import { applyVariations, getPaletteColor, deprecatedColorValue } from './utils'
...
+    @media only screen and (min-width: 768px) {
+      box-shadow: ${boxShadows[props.boxShadowSize[1] || 'md']}
+    }
+​
+    @media only screen and (min-width: 992px) {
+      box-shadow: ${boxShadows[props.boxShadowSize[2] || 'lg']}
+    }
+​
+    @media only screen and (min-width: 1200px)  {
+      box-shadow: ${boxShadows[props.boxShadowSize[3] || 'xl']}
```
pcln-james, line 35: If I provide the prop `boxShadowSize={['md', 'sm']}` I would expect medium boxShadows on the smallest breakpoint and the small boxShadows on all other breakpoints, I don't think this behaviour would occur with the current implementation. Unfortunately I don't immediately have an idea for how to write this in a clean way.

Additionally, I think we would want to use the theme breakpoints rather than hardcoded breakpoints.

sdalonzo: I agree with @James300 - hardcoded breakpoints are a non-starter

sdalonzo: or @pcln-james as he is now known

craigpalermo: Just realized this relates to #844; whatever approach we take will probably apply to both `Card` and `Box`, but we might want to see if Styled System has any utility for this. I know it has a `boxShadow` prop already that supports the responsive array, but I'm not sure how that fits with the `boxShadows` values on our theme.

craigpalermo:
Hey @amyhlee, I'm going to close this PR for now to avoid confusion. I linked it to the issue we have open for responsive box shadows, so we can revisit the implementation once we have a plan that we can apply to both `Box` and `Card`. I need to look more closely at how `styled-system` handles box shadows.

PR #797, opened 2020-05-27
colocate src, test and stories

### LINKED ISSUE #765: Restructure packages to colocate tests and stories with their components
This should make it easier to find tests when working on components, and could also simplify Jest and other configs. We can apply this pattern to all the projects, not just `core`.

**Before**
```
core/
  src/
    A.js
  test/
    A.spec.js
  storybook/
    A.story.js
```

**After**
```
core/
  src/
    A/
      A.js
      A.spec.js
      A.story.js
```

sdalonzo:
- **What does this PR do?**

- Organizes src, stories and tests to be colocated per component in `core`
- adds robust `testing-library` RTL wrapper

- **Any background context you want to provide?**

As discussed in WG

fixes #765


packages/core/src/index.js
```diff
@@ -1,57 +1,59 @@
...
+export { Avatar } from './Avatar'
+export { BackgroundImage } from './BackgroundImage'
+export { Badge } from './Badge'
+export { Banner } from './Banner'
+export { BlockLink } from './BlockLink'
 
-export { default as theme } from './theme'
-export { default as ThemeProvider } from './ThemeProvider'
-export { default as createTheme } from './createTheme'
+export * from './Box'
```
pcln-james, line 8: Why is this different from the others?

sdalonzo: it's not anymore. this was leftover from an earlier iteration and i missed it

PR #801, opened 2020-05-28
Create continuous-integration.yml

sdalonzo:
Prototype of what this CI might look like. Can be merged now, it is not used as a merge check yet but seeing how it runs against PRs will be instructive.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -26,6 +26,4 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm install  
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,29 @@
...
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 14.x]
```
craigpalermo, line 20: Did not know this was a thing 👍

PR #815, opened 2020-05-28
Add prop types for image components

### LINKED ISSUE #790: Define prop types for image components
Define prop types for the following core components:

- Image
- BackgroundImage
- PlaceholderImage

craigpalermo:
Resolves #790

### DIFF
--- a/packages/types/types/index.d.ts
+++ b/packages/types/types/index.d.ts
@@ -13,6 +13,7 @@ import {
   FlexWrapProps,
   FontSizeProps,
   FontWeightProps,
+  HeightProps,
   JustifyContentProps,
   LeftProps,
   LineHeightProps,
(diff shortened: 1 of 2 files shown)


packages/types/types/index.d.ts
```diff
@@ -48,6 +49,13 @@ type RefPropType = (() => any) | ReactRefObject
  */
 type AbsoluteProps = TopRightBottomLeft
 
+export interface BackgroundImageProps extends WidthProps {
+  height?: string
+  /** URL of background image */
```
pcln-james, line 54: Is this comment necessary?

craigpalermo: I figured it couldn't hurt to be explicit because it's not a standard name like `src`

Issue #768, opened 2020-05-01
Add aria-label for DialogContent inside of pcln-modal


hakimelek:
In development I see the following warning when using the Modal component:

```
Warning: Failed prop type: A <DialogContent> must have either an `aria-label` or `aria-labelledby` prop.
      
See https://www.w3.org/TR/wai-aria/#aria-label for details.
    in DialogContent (created by Styled(DialogContent))
    in Styled(DialogContent) (created by r)
    in div (created by Box)
    in Box (created by Styled(Box))
    in Styled(Box) (created by r)
    in div (created by styled.div)
    in styled.div (created by r)
    in div (created by ForwardRef)
    in div (created by ForwardRef)
    in ForwardRef (created by ForwardRef)
    in ForwardRef (created by ForwardRef)
    in div (created by ForwardRef)
    in ForwardRef (created by ForwardRef)
    in ForwardRef (created by ForwardRef)
    in ForwardRef (created by DialogOverlay)
    in Portal (created by DialogOverlay)
    in DialogOverlay (created by Styled(DialogOverlay))
    in Styled(DialogOverlay) (created by r)
    in r (created by yt)
    in yt (at ExpressDealModal/index.tsx:12)
    in ExpressDealModal (at SearchForm/index.tsx:74)
    ...
```

PR #816, opened 2020-05-28
Add aria-label / aria-lablledby props to Modal

pcln-barry:
Resolves https://github.com/priceline/design-system/issues/768

Looks like the pre-commit hook linted a few things, I can revert that if it was incorrect.

### DIFF
--- a/packages/modal/package.json
+++ b/packages/modal/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-modal",
-  "version": "3.3.0",
+  "version": "3.4.0",
   "description": "React component for modal",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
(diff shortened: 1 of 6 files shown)


packages/modal/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pcln-modal",
-  "version": "3.3.0",
+  "version": "3.4.0",
```
craigpalermo, line 3: Should we hold off on version bumps until the hackathon is over?

craigpalermo:
LGTM pending a decision on bumping the version as part of this PR

packages/modal/package.json
pcln-james: No I think it's fine unless it's a major 😛

pcln-james: Unless we want to group all our bumps into one

craigpalermo: I guess we also don't need to publish each time we bump the version

PR #818, opened 2020-05-28
Define prop types for icon & buttons

### LINKED ISSUE #789: Define prop types for icon components
Define prop types for the following core components:

- Icon
- IconButton
- IconField

### LINKED ISSUE #794: Define prop types for button components
Define prop types for the following core components:

- Button
- CloseButton

bmansell:
- Add types for `Icon` and components using `Icon`
- Adding types for `IconButton` required types for `Button`, so did that and `CloseButton`

Closes #789 
Closes #794

### DIFF
--- a/packages/types/types/index.d.ts
+++ b/packages/types/types/index.d.ts
@@ -26,6 +26,7 @@ import {
   TopProps,
   WidthProps,
   ZIndexProps,
+  SizeProps,
 } from 'styled-system'
 
 /**
@@ -62,6 +63,17 @@ export interface BoxProps
     SpaceProps,
     TextAlignProps {}
 
+export interface ButtonProps
+  extends ColorProps,
+  SpaceProps,
+  WidthProps {
+    disabled?: boolean,
+    dsRef?: RefPropType
+    onClick?: React.MouseEventHandler<HTMLElement>
+    size?: 'small' | 'medium' | 'large' | number
+    variation?: 'fill' | 'outline' | 'link'
+}
+
 export interface FlexProps
   extends BoxProps,
     SpaceProps,
@@ -82,6 +94,21 @@ export interface HideProps extends BoxProps {
   print?: boolean
 }
 
+export interface IconProps
+  extends ColorProps,
+  SizeProps {
+  name?: string
+  title?: string
+  titleId?: string
+  desc?: string
+  descId?: string
+}
+
+export interface IconButtonProps extends ButtonProps {
+  icon?: React.ComponentElement<IconProps, any>
+  dsRef?: RefPropType
+}
+
 export interface CardProps
   extends BoxProps,
     BorderRadiusProps,
@@ -89,6 +116,13 @@ export interface CardProps
     TextColorProps {
   boxShadowSize?: 'sm' | 'md' | 'lg' | 'xl'
   borderWidth?: 0 | 1 | 2
+}
+
+export interface CloseButtonProps
+  extends IconButtonProps,
+  ButtonProps {
+  size?: number
+  title?: string
 }
 
 export interface ImageProps extends WidthProps {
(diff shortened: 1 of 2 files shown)


packages/types/types/test.tsx
```diff
@@ -62,6 +68,53 @@ const myBlockLink = (
     dsRef={() => 'ima ref'}
   />
 )
+// $ExpectType Element
+const myButton = (
+  <Button
+    variation={'fill'}
```
sdalonzo, line 74: n-b curly braces around strings are unnecessary

packages/types/types/index.d.ts
```diff
@@ -74,6 +85,24 @@ export interface HideProps extends BoxProps {
...
+  descId?: string
+}
+
+export interface IconButtonProps extends ButtonProps {
+  icon?: React.ComponentElement<IconProps, any>
+  onClick?: Function
+  dsRef?: RefPropType
+}
+
+export interface IconFieldProps extends FlexProps {}
```
craigpalermo, line 104: Can you try running `npm run test` from the types project? I think this will complain about `no-empty-interface`. In this case, you should just be able to use `FlexProps` as the prop types for `IconField` since they use the same interface.

packages/types/types/index.d.ts
```diff
@@ -74,6 +85,24 @@ export interface HideProps extends BoxProps {
...
+  name?: string
+  title?: string
+  titleId?: string
+  desc?: string
+  descId?: string
+}
+
+export interface IconButtonProps extends ButtonProps {
+  icon?: React.ComponentElement<IconProps, any>
+  onClick?: Function
```
craigpalermo, line 100: I think `onClick` could probably go on `ButtonProps` instead since it's needed there too

packages/types/types/index.d.ts
```diff
@@ -83,6 +112,12 @@ export interface CardProps
   borderWidth?: 0 | 1 | 2
 }
 
+export interface CloseButtonProps
+  extends IconButtonProps,
+  ButtonProps {
+  title?: string
```
craigpalermo, line 118: I think `CloseButton` also needs `size?: number`

Side note: that's weird, because it also passes that size prop along to `IconButton`, so it seems like `size` is being used in two different ways by that component.

packages/types/types/test.tsx
```diff
@@ -62,6 +68,53 @@ const myBlockLink = (
...
+const myIconButton = (
+  <IconButton
+    dsRef={() => 'ima ref'}
+    icon={<Icon name='Calendar' title='Choose date' />}
+    onClick={() => 'ima function'}
+  />
+)
+// $ExpectType Element
+const myIconField = (
+  <IconField>
```
craigpalermo, line 113: Also doesn't hurt to add some props to `IconField` for extra validation

packages/types/types/index.d.ts
bmansell: Yeah you're right, `CloseButton` uses its `size` prop to control the Close icon's size. It's unfortunate because I have to add `number` as a type to the `ButtonProps` interface otherwise it conflicts with the inherited type.

```
Interface 'CloseButtonProps' incorrectly extends interface 'ButtonProps'.
  Types of property 'size' are incompatible.
    Type 'number | undefined' is not assignable to type '"small" | "medium" | "large" | undefined'.
      Type 'number' is not assignable to type '"small" | "medium" | "large" | undefined'.ts(2430)
```

Perhaps we should add an issue to refactor `CloseButton`'s size prop to be more specific like `iconSize` so that it's less muddy?

packages/types/types/test.tsx
bmansell: Will fix! Wasn't sure as it seemed like the standard in this file. Thanks!

packages/types/types/index.d.ts
bmansell: Makes sense, moved into `ButtonProps` and also added a more specific type (`React.MouseEventHandler<HTMLElement>`) based on linter feedback!

packages/types/types/test.tsx
bmansell: Added!

packages/types/types/index.d.ts
bmansell: Yup you were right, fixed!

PR #819, opened 2020-05-29
feat(input-prop-types): Add type defs for input related components

mjbyrnes4664:


packages/types/types/test.tsx
```diff
@@ -53,6 +61,14 @@ const mySpan = <Text.span regular /> // $ExpectType Element
...
 const myFlex = <Flex flexDirection={['row', null, 'column']} m={3} /> // $ExpectType Element
 const myHide = <Hide xs sm md lg xl xxl m={3} /> // $ExpectType Element
+const myCheckbox = <Checkbox id="test" size={1} onChange={() => {}}/> // $ExpectType Element
+const myFormField = <FormField /> // $ExpectType Element
+const myLabel = <Label /> // $ExpectType Element
+const myRadio = <Radio /> // $ExpectType Element
+const mySelect = <Select /> // $ExpectType Element
+const myTextArea = <TextArea id='test'/> // $ExpectType Element
+const myInput = <Input id='test'/> // $ExpectType Element
+const myInputGroup = <InputGroup /> // $ExpectType Element
```
craigpalermo, line 71: Can you add some test props to the new components? That seems to be a good way to validate prop types in addition to the expected return types

packages/types/types/index.d.ts
```diff
@@ -147,6 +154,38 @@ export interface ThemeProviderProps {
...
+}
+
+export interface SelectProps extends DsRefProps, SpaceProps, FontSizeProps, ColorValidation, BordersProps {}
+
+export interface TextAreaProps extends ColorValidation, DsRefProps, BordersProps, SpaceProps {
+  id: string
+}
+
+export interface InputProps extends DsRefProps, ColorValidation, BordersProps, SpaceProps, FontSizeProps {
+  id: string
```
craigpalermo, line 184: Maybe we should make an `IdProps` interface too?

packages/types/types/index.d.ts
```diff
@@ -147,6 +154,38 @@ export interface ThemeProviderProps {
...
+}
+
+export interface CheckboxProps extends DsRefProps, ColorValidation {
+  id: string
+  size?: number
+  onChange: () => void
+}
+
+export interface FormFieldProps {
+  children?: PropValidator
```
craigpalermo, line 168: It should be safe to omit `children` since that's already taken care of when extending `React.Component`

packages/types/types/index.d.ts
```diff
@@ -147,6 +154,38 @@ export interface ThemeProviderProps {
   customBreakpoints?: string[]
 }
 
+export interface ColorValidation {
+  color?: PropValidator
```
craigpalermo, line 158: Would `TextColorProps` from `styled-system` work instead?

mjbyrnes4664: Does `TextColorProps` handle what we're doing to validate if a color is semantic or not?

craigpalermo: No, I've just been using `TextColorProps` until we can replace it everywhere with whatever custom color types we come up with

Issue #787, opened 2020-05-27
Define prop types for static display components


craigpalermo:
Define prop types for the following core components:

- Avatar
- Banner
- Card
- Container
- Divider
- Flag
- Hug
- Stamp
- Tooltip

PR #823, opened 2020-05-29
Add prop types for static display components

pcln-barry:
Resolves https://github.com/priceline/design-system/issues/787

### DIFF
--- a/packages/types/types/index.d.ts
+++ b/packages/types/types/index.d.ts
@@ -5,6 +5,7 @@
 
 import * as React from 'react'
 import {
+  AlignItemsProps,
   BorderColorProps,
   BorderRadiusProps,
   BottomProps,
@@ -16,7 +17,9 @@ import {
   JustifyContentProps,
   LeftProps,
   LineHeightProps,
+  ResponsiveValue,
   RightProps,
+  SizeProps,
   SpaceProps,
   TextAlignProps,
   TextColorProps,
@@ -25,9 +28,6 @@ import {
   TopProps,
   WidthProps,
   ZIndexProps,
-  SizeProps,
-  ResponsiveValue,
-  AlignItemsProps,
 } from 'styled-system'
 import * as CSS from 'csstype'
 
@@ -92,7 +92,7 @@ export interface BreadcrumbLinkProps {
   /** This is the last link in the breadcrumbs list */
   isLastChild?: boolean
   href?: string
-  icon?: typeof Icon
+  icon?: React.ComponentElement<IconProps, any>
   label?: string
   dsRef?: RefPropType
   onClick?: (any) => void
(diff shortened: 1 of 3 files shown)


packages/types/types/test.tsx
```diff
@@ -193,3 +201,39 @@ const myPlaceholderImage = (
...
+  className={'className'}
+  title={'Title'}
+  subtitle={'Subtitle'}
+  src={'opossum.jpg'}
+  initials={'TS'}
+  size={100}
+/>
+// $ExpectType Element
+const myBanner = <Banner
+  header={'Header text'}
```
craigpalermo, line 215: n-b strings can just be `header="..."`

packages/types/types/index.d.ts
```diff
@@ -193,18 +194,73 @@ export interface ThemeProviderProps {
...
+}
+
+export interface FlagProps extends FlexProps,
+    ColorProps,
+    WidthProps {}
+
+export interface HugProps extends CardProps, ColorProps {
+  iconDisplay?: string[],
+  icon?: React.ComponentElement<IconProps, any>,
+  text?: React.ComponentElement<any, any> | React.ComponentElement<any, any>[] | string
```
craigpalermo, line 240: I think you probably want `ReactNode | string` here

packages/types/types/test.tsx
pcln-barry: Oh perfect, that's generally my preference but at a glance I saw the brackets being used for strings!

packages/types/types/index.d.ts
pcln-barry: Thank you, wasn't quite sure what to use.

packages/types/types/test.tsx
pcln-barry: Unless you mean literally `{'...'}`, which works too

packages/types/types/index.d.ts
pcln-barry: I included the array since the PropTypes for that component has
```
  text: PropTypes.oneOfType([
    PropTypes.node,
    PropTypes.arrayOf(PropTypes.node),
    PropTypes.string,
  ]),
```

craigpalermo: Good call 👍 I've been referring to the React types when I'm not sure if something exists or not. Here's what I found for `ReactNode`.

```
    interface ReactNodeArray extends Array<ReactNode> {}
    type ReactFragment = {} | ReactNodeArray;
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

packages/types/types/index.d.ts
```diff
@@ -193,18 +194,73 @@ export interface ThemeProviderProps {
...
+}
+
+export interface FlagProps extends FlexProps,
+    ColorProps,
+    WidthProps {}
+
+export interface HugProps extends CardProps, ColorProps {
+  iconDisplay?: string[],
+  icon?: React.ComponentElement<IconProps, any>,
+  text?: React.ReactNode | React.ReactNode[] | string
```
craigpalermo, line 240: I think this can be changed to `  text?: ReactNode | string` because `ReactNode` already handles an array of nodes:

```
    interface ReactNodeArray extends Array<ReactNode> {}
    type ReactFragment = {} | ReactNodeArray;
    type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

packages/types/types/index.d.ts
```diff
@@ -193,18 +194,73 @@ export interface ThemeProviderProps {
...
+  maxWidth?: number
+}
+
+export interface FlagProps extends FlexProps,
+    ColorProps,
+    WidthProps {}
+
+export interface HugProps extends CardProps, ColorProps {
+  iconDisplay?: string[],
+  icon?: React.ComponentElement<IconProps, any>,
```
craigpalermo, line 239: We can actually make this a bit more restrictive by making it:

```
  icon?: typeof Icon,
```

That'll use the type of the `Icon` class we create below.

packages/types/types/index.d.ts
```diff
@@ -193,18 +194,73 @@ export interface ThemeProviderProps {
...
+  src?: string,
+  /** Alternative initials to display when src is not provided */
+  initials?: string,
+  /** Size in px to use for height, width, and background-size */
+  size?: number
+}
+
+export interface BannerProps extends BoxProps {
+  header?: string | React.ComponentElement<IconProps, any>,
+  icon?: React.ComponentElement<IconProps, any>,
```
craigpalermo, line 217: See comment below about `typeof Icon`, but for line 216 we can change it to:

```
  header?: string | ReactNode,
```

That's because `header` accepts any node, but we want to restrict `icon` to just `Icon`s.

packages/types/types/index.d.ts
pcln-barry: Ahhhh that's good to know! I'll update!

packages/types/types/index.d.ts
pcln-barry: Good to know for future reference, thank you!

packages/types/types/index.d.ts
pcln-barry: 👍!

packages/types/types/index.d.ts
pcln-barry: Makes sense, updated 👍

packages/types/types/index.d.ts
```diff
@@ -193,18 +194,73 @@ export interface ThemeProviderProps {
...
+  icon?: React.ComponentElement<IconProps, any>
+  onClose?: (any) => void
+  showIcon?: boolean
+  text?: string | React.ReactNode
+}
+
+export interface StampProps
+  extends SpaceProps, FontSizeProps, ColorProps {
+    variation?: 'outline' | 'fill' | 'solid',
+    size?: 'small' | 'medium'
```
craigpalermo, line 226: Needs `bg` deprecated prop

packages/types/types/index.d.ts
```diff
@@ -193,18 +194,73 @@ export interface ThemeProviderProps {
...
+  /** Subheading displayed beneath heading */
+  subtitle?: string
+  src?: string
+  /** Alternative initials to display when src is not provided */
+  initials?: string
+  /** Size in px to use for height, width, and background-size */
+  size?: number
+}
+
+export interface BannerProps extends BoxProps {
```
craigpalermo, line 215: Needs `bg` deprecated prop

packages/types/types/index.d.ts
```diff
@@ -193,18 +194,73 @@ export interface ThemeProviderProps {
...
+  extends SpaceProps, FontSizeProps, ColorProps {
+    variation?: 'outline' | 'fill' | 'solid',
+    size?: 'small' | 'medium'
+  }
+
+export interface ContainerProps {
+  maxWidth?: number
+}
+
+export interface FlagProps extends FlexProps,
```
craigpalermo, line 233: Needs `bg` deprecated prop

packages/types/types/index.d.ts
```diff
@@ -193,18 +194,73 @@ export interface ThemeProviderProps {
...
+  text?: React.ReactNode | string
+}
+
+export interface TooltipProps extends BoxProps {
+  zIndex?: number | string
+  bottom?: boolean
+  top?: boolean
+  center?: boolean
+  left?: boolean
+  right?: boolean
```
craigpalermo, line 249: Needs `bg` deprecated prop

Issue #830, opened 2020-05-29
Add declarations for all deprecated propTypes


craigpalermo:
Bri made a good point that we still need to declare prop types for deprecated props, otherwise TSC will complain when trying to use deprecated props.

PR #831, opened 2020-05-29
Add deprecated props to components that already have declarations

craigpalermo:
Related to #830


pcln-barry:
LGTM, I'll rebase once this is in to use the new deprecated type

PR #832, opened 2020-06-01
Integrate ESLint into Design System

sdalonzo:
- Add ESLint
- Add ESLint pre-commit hook
- Fix low hanging fruit
- Add ignores for other issues to be addressed later


.eslintrc.js
```diff
@@ -0,0 +1,44 @@
...
+    node: true,
+  },
+  extends: [
+    'eslint:recommended',
+    'plugin:react/recommended',
+    'plugin:jest/style',
+    'pcln-accessibility',
+    'prettier',
+    // 'plugin:@typescript-eslint/eslint-recommended',
+    // 'plugin:@typescript-eslint/recommended',
```
craigpalermo, line 14: ESLint requires a bit of extra configuration to work with TypeScript, so it'll probably make sense to keep these commented out until we support `.tsx?` files outside of the `types` project.

docs/IconField.md
```diff
@@ -45,7 +45,7 @@ It will parse children based on the `isField` and `isIcon` static properties.
 By default the design system `Input`, `Select`, and `Icon` components will work with the `IconField` component, but to allow other children to render, add an appropriate static property.
 
 ```jsx
-import React from 'react'
+import * as React from 'react'
```
craigpalermo, line 48: Do we need the changed import syntax since we're not turning on TS plugins?

sdalonzo: maybe not? I thought I remembered reading that this was the new best practice..is that only for TS?

.eslintrc.js
sdalonzo: sure, makes sense

sdalonzo: these came as part of `eslint --init` when prompted for TS support. i figured it was easier to add them now and comment it out, then to have to piece it back together later.

PR #841, opened 2020-06-02
Typescript bugfixes

### LINKED ISSUE #835: Inaccurate comment about alternative to deprecated "bg" prop
**Describe the bug**
The comment says `bg` is deprecated and to use `color` instead, but `color` changes the text color, not the background.

Screen Shot 2020-06-01 at 2 46 14 PM

### LINKED ISSUE #836: backgroundColor incorrectly listed in autocomplete for some components
**Describe the bug**
Prop type declarations that extend `ColorProps` from `styled-system` show `backgroundColor` as an prop in the code completion options, but this prop doesn't seem to actually do anything. The correct prop to use is `bg`. We should figure out how to exclude `backgroundColor` so people aren't confused.

Screen Shot 2020-06-01 at 2 45 25 PM

### LINKED ISSUE #837: Component declarations appear to have multiple overloads
**Describe the bug**
It looks like components have multiple overloads because we're using explicit `any` for the type of `context` for all component declarations. Removing the explicit `any` should solve this problem. None of our components use context, so this should be safe.

Screen Shot 2020-06-01 at 2 55 02 PM

craigpalermo:
- Resolves #835 
- Resolves #836 (same fix for #835)
- Resolves #837

### DIFF
--- a/packages/types/types/index.d.ts
+++ b/packages/types/types/index.d.ts
@@ -4,7 +4,7 @@
 /// <reference types="@types/styled-system" />
 
 /* Note: Using this syntax instead of synthetic default export because we shouldn't assume the TS compiler
-* options used by DS consumers. */
+ * options used by DS consumers. */
 import * as React from 'react'
 
 import {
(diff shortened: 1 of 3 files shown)


packages/types/types/index.d.ts
```diff
@@ -65,6 +68,27 @@ export interface IdProps {
...
+
+  /**
+   * Sets the background color of the component to the given semantic color. Also sets the text color
+   * based on the theme and the given semantic color name to ensure an accessible contrast level.
+   *
+   * [See the docsite for the list of semantic color names](https://priceline.github.io/design-system/palette)
+   */
+  color?: string // TODO replace with type that enumerates semantic color names, as well as combinations of colors and shades
+}
+
```
bmansell, line 91: 🎉

PR #844, opened 2020-06-05
Add boxShadowSize prop to Box

### LINKED ISSUE #843: Add boxShadow as a prop to Box
On the flights team, we often implement designs with a box shadow, but the only DS component that offers an easy box-shadow implementation is `<Card />`. Would be great to have this standardized across other components, especially Box and those that inherit from it.

craigpalermo:
Also updates the relevant type declarations.

Resolves #843


craigpalermo:
Latest travis build succeeded, but status isn't reporting for some reason

https://travis-ci.org/github/priceline/design-system/builds/695098472

packages/core/src/Card/__snapshots__/Card.spec.js.snap
```diff
@@ -13,6 +13,7 @@ exports[`Card renders border 0 without warning 1`] = `
 
 exports[`Card renders large box shadow with default border 1`] = `
 .c0 {
+  box-shadow: 0 0 2px 0 rgba(0,0,0,.08),0 4px 16px 0 rgba(0,0,0,.16);
```
sdalonzo, line 16: why 2 shadows now??

Issue #710, opened 2020-01-17
Prepare for `styled-components` v5


sdalonzo:
This is preparatory work for SCv5

Increase `devDependency` to styled-components 5
Loosen `peerDependency` to `>=3` so consumers are free to use v5

PR #848, opened 2020-06-08
feat(sc-version-bump): uptick styled components to v5

muddu-sreeram:
* Trying to uptick styled-components dev dependency to `v5` resolving  #710


sdalonzo:
This probably requires a broader discussion as raising the peer floor on SC will need to be a major and most apps are still on v3. Can you come to the WG meeting this week?

PR #852, opened 2020-06-17
fix(821): Rename size to iconSize in CloseButton

### LINKED ISSUE #821: CloseButtonProps size type conflicts with inherited ButtonProps type
`CloseButton` uses its `size` prop to control the Close icon's size. This is a problem because `CloseButtonProps` extends `ButtonProps` which defines `size` as a specific string set so the resulting type is a mix of specific strings or a number. 

https://github.com/priceline/design-system/blob/aede4ea27538802e7283c0c7d7e210cf1ca3a0c6/packages/types/types/index.d.ts#L73

We should refactor `CloseButton`'s size prop to be more specific like `iconSize` so that these types don't have to be collapsed into one.

bmansell:
- Rename the `size` prop to `iconSize` to prevent type conflicts between `Button` and `CloseButton`
- Update snapshots

Closes #821

### DIFF
--- a/packages/core/src/CloseButton/CloseButton.js
+++ b/packages/core/src/CloseButton/CloseButton.js
@@ -6,17 +6,17 @@ import { refPropType } from '../utils'
 import { IconButton } from '../IconButton'
 
 const CloseButton = (props) => (
-  <IconButton {...props} icon={<Close size={props.size} />} />
+  <IconButton {...props} icon={<Close size={props.iconSize} />} />
 )
 
 CloseButton.defaultProps = {
-  size: 24,
+  iconSize: 24,
   title: 'close',
 }
 
 CloseButton.propTypes = {
   onClick: PropTypes.func,
-  size: PropTypes.number,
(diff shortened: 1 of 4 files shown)


packages/core/src/CloseButton/CloseButton.js
```diff
@@ -6,17 +6,17 @@ import { refPropType } from '../utils'
...
 CloseButton.defaultProps = {
-  size: 24,
+  iconSize: 24,
   title: 'close',
 }
 
 CloseButton.propTypes = {
   onClick: PropTypes.func,
-  size: PropTypes.number,
+  iconSize: PropTypes.number,
```
craigpalermo, line 19: For consistency, I think it'd make sense to keep `size`, but change it to use the same values as `size` on `Button`. Then, we could add `iconSize`, which would control the size of the icon. However, if we remove `size` or change it so it no longer accepts a number, I think that would be a breaking change.

bmansell:
Closing as we need a better, less break-y way to do this.

PR #863, opened 2020-06-29
Use GH Actions to build docsite and static storybook

craigpalermo:


sdalonzo:
I know Pages isn't the solution for this (and it's out of scope for this PR) - but we should look into something similar to Storybook Sandbox for DS - maybe just a quick Netlify thing - for reviewing storybook and/or docsite at PR time.

Issue #861, opened 2020-06-29
Increase React and React DOM peer floor to 16.10


craigpalermo:
In order to support React hooks, we need to raise the minimum supported versions of `react` and `react-dom`.

craigpalermo:
@sdalonzo Should this apply to just `pcln-design-system`, or the non-core packages as well?

sdalonzo:
I'm going to do it across the board. We want to keep majors in sync anyway (everything should be 4.x after)

PR #865, opened 2020-06-30
Design System v4

sdalonzo:
- Bump react and react-dom peerDependency floor to ^16.10.0
- Rein in peers to use caret instead of >= until we know what the breaking change in React 17 might look like
- resolves gh-861


packages/autocomplete/package.json
```diff
@@ -45,15 +45,16 @@
...
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "rimraf": "^3.0.2",
-    "styled-components": "^3.4.10",
+    "styled-components": "^4.4.1",
     "us": "^2.0.0"
   },
   "peerDependencies": {
-    "pcln-design-system": "^3.7.1"
+    "pcln-design-system": "^4.0.0",
```
pcln-james, line 57: Should this require `react` as a peer too?

pcln-james (approved):

packages/autocomplete/package.json
craigpalermo: I think we can leave it out here because they'll get the same warning if they're missing the same peer that's required by `pcln-design-system` core.

pcln-barry (approved):

PR #871, opened 2020-07-06
Establish TS pattern for core DS components

### LINKED ISSUE #862: Rewrite Box component with TypeScript to establish pattern for core components written in TS

craigpalermo:
This PR configures the tooling required to support TS core components. It also includes a TS refactor of `Box`, `ThemeProvider`, and a few utility functions. The declarations in `src/@types` will also be useful as we begin to refactor more components.

Resolves #862


packages/core/src/Box/Box.tsx
```diff
@@ -0,0 +1,52 @@
...
+  ${space} ${width} ${textAlign} ${boxShadow as never}
+  ${color}
+  ${applyVariations('Box') as never}
+` as React.FunctionComponent<BoxProps>
+
+/**
+ * We can keep the regular prop types around for deprecation warnings at runtime
+ */
+Box.propTypes = {
+  align: deprecatedPropType('textAlign') as never,
```
craigpalermo, line 47: TODO: find type for propTypes validators

packages/core/src/Box/Box.tsx
```diff
@@ -0,0 +1,52 @@
...
+  style?: React.CSSProperties
+}
+
+/**
+ * A low-level layout component for setting width, margin, padding, and color
+ */
+export const Box = styled.div<BoxProps>`
+  ${space} ${width} ${textAlign} ${boxShadow as never}
+  ${color}
+  ${applyVariations('Box') as never}
```
craigpalermo, line 40: TODO: find type for styled component style functions

packages/core/package.json
```diff
@@ -9,9 +9,9 @@
   "scripts": {
     "prepare": "run-s clean build",
     "build": "run-s build:js build:types",
-    "build:babel:cjs": "cross-env NODE_ENV=production BABEL_ENV=cjs babel ./src --ignore '**/*.stories.js,**/*.spec.js' --out-dir ./dist/cjs",
-    "build:babel:esm": "cross-env NODE_ENV=production BABEL_ENV=esm babel ./src --ignore '**/*.stories.js,**/*.spec.js' --out-dir ./dist/esm",
-    "build:babel:es": "cross-env NODE_ENV=production BABEL_ENV=es babel ./src --ignore '**/*.stories.js,**/*.spec.js' --out-dir ./dist/es",
+    "build:babel:cjs": "cross-env NODE_ENV=production BABEL_ENV=cjs babel ./src --out-dir ./dist/cjs --extensions \".js,.ts,.tsx\"",
```
sdalonzo, line 12: please don't remove the `--ignore` flags

packages/core/src/Flag/__snapshots__/Flag.spec.js.snap
```diff
@@ -58,7 +58,7 @@ exports[`Flag renders 1`] = `
   display: flex;
 }
 
-@media screen and (min-width:32em) {
+@media screen and (min-width:32) {
```
sdalonzo, line 61: this is a breaking change that needs correction

packages/core/src/FormField/__snapshots__/FormField.spec.js.snap
```diff
@@ -72,7 +72,7 @@ exports[`FormField renders 1`] = `
   font-weight: 700;
 }
 
-@media screen and (min-width:40em) {
+@media screen and (min-width:40) {
```
sdalonzo, line 75: same here

packages/core/src/Hide/__snapshots__/Hide.spec.js.snap
```diff
@@ -7,7 +7,7 @@ exports[`Hide renders 1`] = `
 `;
 
 exports[`Hide renders with lg prop 1`] = `
-@media screen and (min-width:48em) and (max-width:63.99em) {
+@media screen and (min-width:48) and (max-width:63.99em) {
```
sdalonzo, line 10: here as well

packages/core/src/IconField/__snapshots__/IconField.spec.js.snap
```diff
@@ -66,7 +66,7 @@ exports[`IconField renders 1`] = `
   box-shadow: 0 0 0 2px #007aff;
 }
 
-@media screen and (min-width:40em) {
+@media screen and (min-width:40) {
```
sdalonzo, line 69: same

packages/core/src/Input/__snapshots__/Input.spec.js.snap
```diff
@@ -48,7 +48,7 @@ exports[`Input it renders 1`] = `
   box-shadow: 0 0 0 2px #007aff;
 }
 
-@media screen and (min-width:40em) {
+@media screen and (min-width:40) {
```
sdalonzo, line 51: here as well

packages/core/src/Select/__snapshots__/Select.spec.js.snap
```diff
@@ -58,7 +58,7 @@ exports[`Select renders 1`] = `
   display: none;
 }
 
-@media screen and (min-width:40em) {
+@media screen and (min-width:40) {
```
sdalonzo, line 61: one more time

packages/core/src/theme/__snapshots__/theme.spec.ts.snap
```diff
@@ -10,11 +10,11 @@ Object {
     "0 0 2px 0 rgba(0,0,0,.08),0 8px 32px 0 rgba(0,0,0,.16)",
   ],
   "breakpoints": Array [
-    "32em",
-    "40em",
-    "48em",
-    "64em",
-    "80em",
+    "32",
```
sdalonzo, line 13: same here

packages/core/src/theme/theme.ts
```diff
@@ -1,18 +1,22 @@
...
       enumerable: false,
       get() {
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-return,@typescript-eslint/no-unsafe-member-access
         return this[i]
       },
     })
   )
 
-export const breakpoints = [32, 40, 48, 64, 80].map((n) => n + 'em')
+export const breakpoints = [32, 40, 48, 64, 80].map((n) => `${n}`)
```
sdalonzo, line 19: I think this is the root cause
```suggestion
export const breakpoints = [32, 40, 48, 64, 80].map((n) => `${n}em`)
```

packages/core/src/utils/utils.tsx
```diff
@@ -170,20 +187,29 @@ export const applyVariations = (componentName, variations = null) => (
...
+
+  // eslint-disable-next-line @typescript-eslint/prefer-regexp-exec
   const colorShade = shade.match(/^([a-z]+)\.([a-z]+)$/i)
 
   if (colorShade) {
     color = colorShade[0]
     shade = colorShade[1]
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-return
```
sdalonzo, line 206: why are these necessary?

packages/core/package.json
craigpalermo: I moved it to the config so we could keep the scripts a bit shorter https://github.com/priceline/design-system/pull/871/files#diff-fc10a6682269c1fefc2abe05d1500b8aR31

packages/core/src/theme/__snapshots__/theme.spec.ts.snap
craigpalermo: I totally missed this, I'll have to look into what's causing it

packages/core/src/utils/utils.tsx
craigpalermo: The `themeGet` function from `styled-components` has return type `any`, and TS doesn't like that we're immediately invoking the result of those calls.

craigpalermo:
@sdalonzo Regarding your question about how intellisense will treat DS components written in TS when they're used in JS files, I confirmed that in VSCode and WebStorm that all the prop types defined in the interface appear as expected without having to supplement that with regular `PropTypes`.

Screen Shot 2020-07-08 at 4 52 53 PM

packages/core/package.json
sdalonzo: that works, I missed that. Just verify that they don't appear in `dist`

sdalonzo:
@craigpalermo can you fix the missing coverage?

craigpalermo:
> @craigpalermo can you fix the missing coverage?

I'm not really sure what it's complaining about since all I did was add type info:
Screen Shot 2020-07-09 at 1 25 52 PM

Any ideas?

PR #882, opened 2020-07-22
Enhancement: added additional props to Box to require fewer styled co…

tfquirk:
Add the following props to `Box` and update docs page for Box, to reflect these updates:
```
 display
 height
 maxHeight
 maxWidth
 minHeight
 minWidth
 size
```
https://github.com/styled-system/styled-system/blob/v4.1.1/docs/api.md#layout
(It looks like styled-system v5 nests all layout props under a `layout` prop. Which will simply code in the future, however, now we need to import each individually.)

This was motivated by the desire to have the ability to add responsive arrays for heights, specifically. Styled-system v4 supports each of the layout options above. Any and all are up for debate, but all could be useful.

### DIFF
--- a/docs/pages/Box.md
+++ b/docs/pages/Box.md
@@ -5,7 +5,7 @@ Use the `Box` component to control width, margin, padding, and color.
 50% width
 
 ```.jsx
-<Box width={1/2} bg='blue'>
+<Box width={1/2} color='primary.base'>
   Box
 </Box>
 ```
@@ -13,7 +13,7 @@ Use the `Box` component to control width, margin, padding, and color.
 Padding of `theme.space[3]` (16px)
 
 ```.jsx
-<Box p={3} bg='blue'>
+<Box p={3} color='secondary.base'>
   Box
 </Box>
 ```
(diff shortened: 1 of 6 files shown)


docs/pages/Box.md
```diff
@@ -5,23 +5,23 @@ Use the `Box` component to control width, margin, padding, and color.
 50% width
 
 ```.jsx
-<Box width={1/2} bg='blue'>
+<Box width={1/2} color='primary.base'>
```
sdalonzo, line 8: 😍

pcln-james: If you're using `base` for a prop it's best to leave it as `primary` and this is true for all colors since if the component have a hover state the DS will take care of that for you and by default use the `dark` color. However, this is only true for working with props, getPaletteColor requires you to be specific.

pcln-james: Not important for box and it's fine to leave the documentation the way it is here but I thought I'd bring it up as I taught someone that the other day.

achoboter: @pcln-james some more documentation on `getPaletteColor` https://priceline.github.io/design-system/utilities/getPaletteColor#palette-color-driven-from-color-prop

PR #883, opened 2020-07-23
Combine actions for static storybook and docsite

craigpalermo:
This PR removes the Travis config and replaces it with a new job in the `continuous-integration` workflow. It also fixes the broken static storybook.

Last step will be to delete the Travis webhook on this repo after merging.


.github/workflows/build-docs.yml
```diff
@@ -43,12 +18,25 @@ jobs:
...
 
-      # Deploy docsite to branch
-      - name: Push docsite build directory to branch
+      - name: Build docsite
+        working-directory: docs
+        env:
+          NODE_ENV: production
+        run: |
+          npm run build
+          touch out/.nojekyll
```
craigpalermo, line 31: Turns out this does need to happen after the build step because Next clears out the target directory before building

craigpalermo: @hakimelek FYI this step was missing, that's what broke the static assets from your issue: `touch out/.nojekyll`

docs/package-lock.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {
     "@ampproject/toolbox-core": {
       "version": "2.5.4",
-      "resolved": "https://registry.npmjs.org/@ampproject/toolbox-core/-/toolbox-core-2.5.4.tgz",
+      "resolved": "https://npm.prod.pcln.com/repository/npm-all/@ampproject/toolbox-core/-/toolbox-core-2.5.4.tgz",
```
pcln-james, line 9: this shouldn't be here 😕

craigpalermo: Good catch, forgot to switch to public

PR #900, opened 2020-08-07
Issue(845): Remove deprecated prop type warnings from tests

### LINKED ISSUE #845: Deprecated prop warnings when running tests
**Describe the bug**
All tests pass, but console shows warnings about deprecated props. We should update tests to use the latest/recommended props.

All

**To Reproduce**

1. Run the tests via `npm test`

**Expected behavior**
Tests shouldn't emit warnings

**Screenshots**
Screen Shot 2020-06-05 at 10 52 40 AM

bmansell:
- Refactor tests to remove all deprecated proptype warnings from tests
  - Change mock pattern for tests explicitly asserting `console.error` calls and messages
- Add components or properties where missing in tests
  - `FormField` uses `Input` which requires an `id` prop set, also uses `Select` which requires a `Label`
  - `Checkbox` requires an `onChange` prop which was missing from one test
- Refactor `Box` and `Tooltip` to remove deprecated prop type use from normal tests and move into deprecated prop type test suite like other existing tests
  - Couldn't do this for components that mark `bg` as deprecated but have not yet removed the requirement of `bg` to render the component correctly (includes `Badge`, `Banner`, `Flag`, `RatingBadge`, `Stamp`, `ToggleBadge`) so I added the `console.error` mock hooks in these components' test suites

Fixes #845 

Screen Shot 2020-08-07 at 3 28 49 PM

### DIFF
--- a/packages/core/src/Absolute/Absolute.spec.js
+++ b/packages/core/src/Absolute/Absolute.spec.js
@@ -1,5 +1,6 @@
 import React from 'react'
-import { Absolute, Flag, Icon, Text } from '..'
+import { Absolute, Flag, Text } from '..'
+import { Coupon as CouponIcon } from 'pcln-icons'
 
 describe('Absolute', () => {
   test('renders with top, left and zIndex props', () => {
@@ -14,13 +15,17 @@ describe('Absolute', () => {
   })
 
   test('renders with flag and Child', () => {
+    const consoleError = console.error
+    console.error = jest.fn()
+
     const json = rendererCreateWithTheme(
       <Absolute top={10} left={0}>
         <Flag>
-          <Icon name='Coupon' /> <Text.span>EXCLUSIVE</Text.span>
+          <CouponIcon /> <Text.span>EXCLUSIVE</Text.span>
         </Flag>
       </Absolute>
     ).toJSON
     expect(json).toMatchSnapshot()
+    console.error = consoleError
   })
 })
--- a/packages/core/src/Badge/Badge.spec.js
+++ b/packages/core/src/Badge/Badge.spec.js
@@ -2,9 +2,12 @@ import React from 'react'
 import { Badge, theme } from '..'
 
 describe('Badge', () => {
-  const consoleError = console.error
-  console.error = jest.fn()
-  afterAll(() => (console.error = consoleError))
+  let consoleError
+  beforeEach(() => {
+    consoleError = console.error
+    console.error = jest.fn()
+  })
+  afterEach(() => (console.error = consoleError))
 
   test('renders', () => {
     const json = rendererCreateWithTheme(<Badge />).toJSON()
--- a/packages/core/src/Badge/Badge.spec.js
+++ b/packages/core/src/Badge/Badge.spec.js
@@ -2,6 +2,10 @@ import React from 'react'
 import { Badge, theme } from '..'
 
 describe('Badge', () => {
+  const consoleError = console.error
+  console.error = jest.fn()
+  afterAll(() => (console.error = consoleError))
+
   test('renders', () => {
     const json = rendererCreateWithTheme(<Badge />).toJSON()
     expect(json).toMatchSnapshot()
--- a/packages/core/src/Banner/Banner.spec.js
+++ b/packages/core/src/Banner/Banner.spec.js
@@ -2,9 +2,12 @@ import React from 'react'
 import { Banner, Text, Heading, theme } from '..'
 
 describe('Banner', () => {
-  const consoleError = console.error
-  console.error = jest.fn()
-  afterAll(() => (console.error = consoleError))
+  let consoleError
+  beforeEach(() => {
+    consoleError = console.error
+    console.error = jest.fn()
+  })
+  afterEach(() => (console.error = consoleError))
 
   test('renders with no props other than theme', () => {
     const json = rendererCreateWithTheme(<Banner />).toJSON()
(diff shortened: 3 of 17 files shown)


packages/core/src/Box/Box.spec.js
```diff
@@ -46,30 +46,26 @@ describe('Box', () => {
...
+  describe('deprecated prop types', () => {
+    const consoleError = console.error
+    beforeEach(() => (console.error = jest.fn()))
+    afterEach(() => (console.error = consoleError))
 
-    expect(
-      console.error.mock.calls
-        .toString()
-        .indexOf(
-          'The Box `align` prop will deprecated. Please use Text instead.'
```
bmansell, line 70: This test was removed as the deprecated prop `align` has been replaced with `textAlign`, so there is no longer a warning thrown when passing `align`.

packages/core/src/Banner/Banner.spec.js
```diff
@@ -2,6 +2,10 @@ import React from 'react'
 import { Banner, Text, Heading, theme } from '..'
 
 describe('Banner', () => {
+  const consoleError = console.error
+  console.error = jest.fn()
```
BeniCheni, line 6: Would you think the code here can be moved inside the `afterAll` function at L7?

packages/core/src/Box/Box.spec.js
```diff
@@ -46,30 +46,25 @@ describe('Box', () => {
...
   })
 
-  test('align prop triggers warning', () => {
-    console.error = jest.fn()
-
-    rendererCreateWithTheme(<Box align='center' />).toJSON()
+  describe('deprecated prop types', () => {
+    const consoleError = console.error
+    beforeEach(() => (console.error = jest.fn()))
+    afterEach(() => (console.error = consoleError))
```
BeniCheni, line 58: Would you think L56 can be omitted and keep everything and simplicity inside `afterEach` function?

packages/core/src/Banner/Banner.spec.js
bmansell: It has to be mocked before the tests run otherwise the console errors will be shown as output while the tests are running and `Banner` still needs `bg` to set its colours correctly.

packages/core/src/Box/Box.spec.js
bmansell: Sort of same answer as my previous comment, `console.error` has to be mocked before any of the tests run to catch the output and use it for the expectations. To properly restore the function after the tests, it needs L56 to get the original `console.error` implementation unless you're thinking of a different method to do that.

packages/core/src/Tooltip/Tooltip.spec.js
```diff
@@ -55,4 +55,23 @@ describe('Tooltip', () => {
     ).toJSON()
     expect(json).toMatchSnapshot()
   })
+
+  describe('deprecated prop types', () => {
+    const consoleError = console.error
+    beforeEach(() => (console.error = jest.fn()))
+    afterEach(() => (console.error = consoleError))
```
craigpalermo, line 62: Just for consistency, we might want to go with something like this where applicable:
```javascript
    let consoleError
    beforeEach(() => {
      consoleError = console.error
      console.error = jest.fn()
    })
    afterEach(() => (console.error = consoleError))
```

I'm not sure it makes a difference, but I generally like keeping all assignments inside of a before/afterEach just to be safe.

packages/core/src/Box/Box.spec.js
bmansell: Okay so I think I understand now, I'll be refactoring these snippets to move the `consoleError = console.error` assignments into the `beforeEach` to be safer if that's along the same lines of what you're talking about, see @craigpalermo's comment.

bmansell: [Updated!](https://github.com/priceline/design-system/pull/900/files/167e8cedaa0b0eb52fcce4eded81b63f420adc8b..8452cb48afe66bbd483f8a42d0fbbc4e428d5744)

packages/core/src/Tooltip/Tooltip.spec.js
bmansell: [Updated!](https://github.com/priceline/design-system/pull/900/files/167e8cedaa0b0eb52fcce4eded81b63f420adc8b..8452cb48afe66bbd483f8a42d0fbbc4e428d5744)

packages/core/src/Absolute/Absolute.spec.js
```diff
@@ -14,13 +15,17 @@ describe('Absolute', () => {
   })
 
   test('renders with flag and Child', () => {
+    const consoleError = console.error
```
sdalonzo, line 18: n-b but setting this up at the top of the spec, and using the `afterEach` hook to restore it is generally the best practice.

packages/core/src/Badge/Badge.spec.js
```diff
@@ -2,6 +2,13 @@ import React from 'react'
 import { Badge, theme } from '..'
 
 describe('Badge', () => {
+  let consoleError
+  beforeAll(() => {
+    consoleError = console.error
+    console.error = jest.fn()
+  })
+  afterAll(() => (console.error = consoleError))
```
sdalonzo, line 10: afterEach

packages/core/src/Absolute/Absolute.spec.js
bmansell: I did it this way for this test because I didn't want to unintentionally suppress warnings coming from the other `Absolute` tests in the suite. This one test's warning was coming from the use of the `Flag` component.

packages/core/src/Badge/Badge.spec.js
bmansell: Refactored the `afterAll` hooks to be `afterEach` hooks!

PR #906, opened 2020-08-19
add new kitchen icon

gbellini911:
Adding a Kitchen icon

second commit is to remove non unique ids that were repeated in several svg files
id="Layer_1"
id="path-1_1_"
id="shape"
id="Combined-Shape"


packages/icons/svg/kitchen.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>kitchen</title>
+    <g id="kitchen" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
```
dansternfeld1, line 4: (non-blocking) Is `id` right? I thought we had some low-level a11y violations due to dupe ids from these.

gbellini911: I'm seeing id used in most other svg files, so not sure if it's an issue 🤔

dgiraldo313: In the current state, I don't think this is an issue because all SVGs are wrapped around the Icon component which contains `aria-hidden="true"` by default. This automatically removed these elements form the navigation of the screen reader. However, if that attribute was removed we would probably start seeing some violations wherever these icons are used.

I brought this app in the accessibility guild channel. We will be discussing this and how we can mitigate this issue going forward.

mpgonzalez: aria-hidden aside, it's still a violation of the w3c standard too, so worth fixing.

PR #915, opened 2020-08-26
Update all tests and stories to use named icons

bmansell:
- Replace all tests and stories usage of `<Icon />` with named icons
- Add console.error mocks to popover tests to remove warnings about `StyledOverlay` (`Box`) `bg` prop use

PR #916, opened 2020-08-26
Update all tests and stories to use named icons

bmansell:
- Replace all tests and stories usage of `<Icon />` with named icons
- Add console.error mocks to popover tests to remove warnings about `StyledOverlay` (`Box`) `bg` prop use
- Update stories to use palette colours

Duplicate of #915 to try and kickoff CI

### DIFF
--- a/packages/core/src/stories/Layouts.stories.js
+++ b/packages/core/src/stories/Layouts.stories.js
@@ -2,19 +2,20 @@ import React from 'react'
 import styled from 'styled-components'
 import { storiesOf } from '@storybook/react'
 
-import { Flex, Box, Text, Icon, Image, Heading } from '..'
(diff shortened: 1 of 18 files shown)


packages/core/src/stories/Layouts.stories.js
```diff
@@ -68,7 +69,7 @@ storiesOf('Layout Examples', module)
   ))
 
 const Border = styled(Box)`
-  border: 1px solid ${(props) => props.theme.colors.lightGray};
+  border: 1px solid ${(props) => getPaletteColor('border.light')(props)};
```
pcln-james, line 72: this can just be `border: 1px solid ${getPaletteColor('border.light')}`

bmansell: Updated!

PR #924, opened 2020-09-02
Add v4 announcement and migration guide

craigpalermo:


MIGRATING_to_v4.md
```diff
@@ -0,0 +1,14 @@
+# Migrate to Design System version 4
```
pcln-james, line 1: Should we start moving the migration guides and similar to folders now since we are getting quite a few of them?

craigpalermo: Probably wouldn't hurt

Issue #913, opened 2020-08-26
Runtime deprecation warnings when using unnamed Icon


craigpalermo:
**Is your feature request related to a problem? Please describe.**
In order to keep moving away from the usage of `<Icon/>`, we should add a runtime warning that suggests using named icons instead.

**Describe the solution you'd like**
Runtime console warning when using Icon

craigpalermo:
Probably want to add an environment check to fire warning anywhere outside of `NODE_ENV='production'`

PR #925, opened 2020-09-02
issue(913): Add runtime deprecation for unnamed Icon

pcln-barry:
https://github.com/priceline/design-system/issues/913


pcln-barry:
One more time, folks!

Issue #476, opened 2019-04-15
Components are using hard coded font-weight values


dgiraldo313:
We are currently hard coding  the `font-weight` for  to `600` for the following components:
- `Badge`
- `Button`
- `Stamp`

Instead, we need to be using the theme regular value, so that we can map it properly when using other themes.

craigpalermo:
Bumping because this is still relevant, but only for `Button` and `Stamp`.

PR #941, opened 2020-10-15
Remove hard coded font-weight from button / stamp

pcln-barry:
https://github.com/priceline/design-system/issues/476 - Removing hard coded font-weights from button / stamp.

### DIFF
--- a/packages/core/src/Button/Button.spec.js
+++ b/packages/core/src/Button/Button.spec.js
@@ -212,7 +212,6 @@ describe('Button', () => {
 
         expect(button).toHaveStyleRule('color', theme.palette.primary.base)
         expect(button).toHaveStyleRule('vertical-align', 'inherit')
-        expect(button).toHaveStyleRule('font-weight', '500')
         expect(button).toHaveStyleRule('-webkit-font-smoothing', 'inherit')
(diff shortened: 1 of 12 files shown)


craigpalermo (approved):

packages/core/src/Button/Button.spec.js
```diff
@@ -212,7 +212,6 @@ describe('Button', () => {
 
         expect(button).toHaveStyleRule('color', theme.palette.primary.base)
         expect(button).toHaveStyleRule('vertical-align', 'inherit')
-        expect(button).toHaveStyleRule('font-weight', '500')
```
craigpalermo, line 215: Should we still include a check for the correct `font-weight`?

pcln-barry: Couldn't hurt - currently it's undefined though since it's not on the underlying components anymore. Would we want to add it as a prop?

craigpalermo: That's what I was wondering about, I'm not sure where the font-weight actually originates. @sdalonzo thoughts?

sdalonzo (approved):

PR #957, opened 2020-11-03
allow sizes prop to accept an array

gbellini911:
https://5a85e56213417800200978ab-zetreramsm.chromatic.com/?path=/story/stamp--pass-an-array-of-sizes

Update `Stamp` so its size property can also accept an array 

_(Shoutout: @pcln-james helped out immensely with this)_

### DIFF
--- a/packages/core/src/Stamp/Stamp.js
+++ b/packages/core/src/Stamp/Stamp.js
@@ -67,8 +67,8 @@ const Stamp = styled.div`
   color: ${getPaletteColor('base')};
 
   ${space} ${fontSize};
-  ${applySizes(sizes)}
-  ${applyVariations('Stamp', variations)}
+  ${applySizes(sizes)};
+  ${applyVariations('Stamp', variations)};
 `
 
 Stamp.displayName = 'Stamp'
@@ -76,7 +76,10 @@ Stamp.displayName = 'Stamp'
 Stamp.propTypes = {
   ...space.propTypes,
   ...fontSize.propTypes,
-  size: PropTypes.oneOf(Object.keys(sizes)),
+  size: PropTypes.oneOfType([
+    Object.keys(sizes),
+    PropTypes.arrayOf(Object.keys(sizes)),
+  ]),
   variation: PropTypes.oneOf(Object.keys(variations)),
   color: deprecatedColorValue(),
   bg: deprecatedPropType('color'),
--- a/packages/core/src/Stamp/Stamp.stories.js
+++ b/packages/core/src/Stamp/Stamp.stories.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import styled from 'styled-components'
 import { storiesOf } from '@storybook/react'
 import { withInfo } from '@storybook/addon-info'
 import { Cartesian } from '@compositor/kit'
@@ -25,6 +26,13 @@ const colors = {
   border: 'border',
   background: 'background',
 }
+
+const BlueStamp = styled(Stamp).attrs({
+  borderColor: 'primary',
+  bg: 'primary',
+  color: 'background.lightest',
+  mr: 2,
+})``
 
 storiesOf('Stamp', module)
   .add(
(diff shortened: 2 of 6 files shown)


packages/core/src/Stamp/Stamp.stories.js
```diff
@@ -87,3 +87,67 @@ storiesOf('Stamp', module)
...
+    <div>
+      <Stamp
+        borderColor='blue'
+        bg='blue'
+        color='white'
+        mr={2}
+        size={['medium', null, null, null, null, null]}
+      >
+        <PinIcon />
+        <Text>Medium at xs</Text>
```
craigpalermo, line 100: I checked the updated storybook in Chromatic, which works great, but can we change these to `Large at xyz`? That'd make it a bit easier to see the differences.

packages/core/src/utils/utils.js
```diff
@@ -108,23 +109,49 @@ export const getContrastRatio = (colorA, colorB) => {
   return (Math.max(luminA, luminB) + 0.05) / (Math.min(luminA, luminB) + 0.05)
 }
 
+export const getBreakpointSize = (array, length) => {
+  const filteredArray = array.slice(0, length)
+  return filteredArray.reduceRight((carry, item) => {
+    return !carry ? item : carry
+  }, null)
```
craigpalermo, line 116: Can we add some comments describing the purpose of each of the steps in this function? It'll make it easier to maintain.

craigpalermo (changes requested):
Looking great so far! Can you also run `rush change`?

packages/core/src/utils/utils.js
gbellini911: Updated 👍 (open to adding even more detail if you think necessary)

craigpalermo (approved):
LGTM, nice work!

packages/core/src/Stamp/Stamp.js
```diff
@@ -66,8 +66,8 @@ const Stamp = styled.div`
   color: ${getPaletteColor('base')};
 
   ${space} ${fontSize};
-  ${applySizes(sizes)}
-  ${applyVariations('Stamp', variations)}
+  ${applySizes(sizes)};
+  ${applyVariations('Stamp', variations)};
 `
 
 Stamp.displayName = 'Stamp'
```
craigpalermo, line 73: JW, do we need to make any changes to the `sizes` prop type, or add any additional values?

craigpalermo:
I think we need to verify that the default behavior still works when no size prop is provided. This storybook diff appears to have changed for the default.

https://www.chromatic.com/snapshot?appId=5a85e56213417800200978ab&id=5fa41c93e8b10200221cb26f

gbellini911:
> I think we need to verify that the default behavior still works when no size prop is provided. This storybook diff appears to have changed for the default.
> 
> https://www.chromatic.com/snapshot?appId=5a85e56213417800200978ab&id=5fa41c93e8b10200221cb26f

@craigpalermo split the 'breaking' line height change into a separate PR

packages/core/src/utils/utils.spec.js
```diff
@@ -1,18 +1,20 @@
...
+  getLuminance,
   getPaletteColor,
-  hasPaletteColor,
   getTextColorOn,
-  getByPalette,
-  color,
-  borders,
-  createTheme,
+  hasPaletteColor,
+  hexToRgb,
```
bmansell, line 17: Import alphabetizing 🙌

packages/core/src/Stamp/Stamp.stories.js
```diff
@@ -87,3 +87,67 @@ storiesOf('Stamp', module)
       </Stamp>
     </div>
   ))
+  .add('Pass an array of sizes', () => (
+    <div>
+      <Stamp
+        borderColor='blue'
+        bg='blue'
+        color='white'
```
bmansell, line 95: Could we switch these to use palette colours instead?
```suggestion
        color='primary'
```

bmansell (changes requested):

packages/core/src/Stamp/Stamp.stories.js
craigpalermo: I'm not sure exactly how it's supposed to work, but this is what `color='primary'` looks like on Stamp (far left):
Screen Shot 2020-11-06 at 3 55 09 PM

craigpalermo: @pcln-brianna , just saw your update. That looks as expected 👍

bmansell: Yeah I realized it was the reverse of what I was thinking after checking our docs haha. Tested the updated suggestion:
Screen Shot 2020-11-06 at 2 57 35 PM

bmansell (approved):

PR #964, opened 2020-11-13
Fix font-weight regression

### LINKED ISSUE #965: Button text is no longer bold by default
**Describe the bug**
Text inside of `Button` no longer has `font-weight: 600` as a result of #941. The solution here is to restore the `font-weight` rule for `Button`, but using the correct `fontWeight` from `theme` instead of hardcoding it.

List of Design System projects relevant to this issue

**Expected behavior**
`Button` has `font-weight: 600`

**Screenshots**
Untitled

pcln-james:
Closes #965

### DIFF
--- a/packages/core/src/Button/Button.js
+++ b/packages/core/src/Button/Button.js
@@ -75,6 +75,7 @@ const variations = {
   `,
   link: css`
     -webkit-font-smoothing: inherit;
+    font-weight: ${(props) => props.theme.fontWeights.medium};
     color: ${getPaletteColor('base')};
     line-height: ${(props) => props.theme.lineHeights.standard};
     vertical-align: inherit;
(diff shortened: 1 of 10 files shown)


packages/core/src/Button/Button.js
```diff
@@ -94,6 +95,7 @@ export const buttonStyles = css`
   text-align: center;
   text-decoration: none;
   font-family: inherit;
+  font-weight: 600;
```
craigpalermo, line 98: Is this where we wanted to get `fontWeight` from `theme`?

pcln-james: bold is 700 though, what do you think?

craigpalermo (approved):

packages/core/src/Button/Button.js
pcln-james: We don't even bring in the 600 font weight, at least on the landing page so I don't think changing it to 700 is an issue because it is already using that when set to 600
image

sdalonzo: yea i would change it to the theme.fontWeights that maps to 700

sdalonzo (approved):

PR #967, opened 2020-11-13
New pcln-menu package

### LINKED ISSUE #704: Create ul, ol, and li components
Create ul, ol, and li components

pcln-james:
Closes #704 

Took some inspiration from the material-ui Menu component https://material-ui.com/components/menus/

### DIFF
--- a/packages/menu/package.json
+++ b/packages/menu/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pcln-menu",
-  "version": "2.0.0",
+  "version": "3.0.0",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "sideEffects": false,
--- a/packages/menu/package.json
+++ b/packages/menu/package.json
@@ -1,0 +1,58 @@
+{
+  "name": "pcln-menu",
+  "version": "2.0.0",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "sideEffects": false,
+  "scripts": {
+    "prepare": "npm run clean && npm run build",
+    "build": "npm-run-all -p build:*",
+    "build:cjs": "cross-env NODE_ENV=production BABEL_ENV=cjs babel ./src --ignore '**/*.stories.js,**/*.spec.js' --out-dir ./dist/cjs",
+    "build:esm": "cross-env NODE_ENV=production BABEL_ENV=esm babel ./src --ignore '**/*.stories.js,**/*.spec.js' --out-dir ./dist/esm",
+    "build:es": "cross-env NODE_ENV=production BABEL_ENV=es babel ./src --ignore '**/*.stories.js,**/*.spec.js' --out-dir ./dist/es",
+    "clean": "rimraf dist",
+    "lint": "eslint .",
+    "lint:fix": "eslint . --fix",
+    "start": "start-storybook",
+    "test": "jest --verbose --coverage",
+    "test:ci": "jest --silent --coverage --runInBand",
+    "test:update": "jest --updateSnapshot",
+    "test:ci:update": "jest --runInBand --updateSnapshot"
+  },
+  "files": [
+    "dist/"
+  ],
+  "keywords": [],
+  "author": "Priceline",
+  "license": "MIT",
+  "devDependencies": {
+    "@babel/cli": "^7.11.6",
+    "@priceline/babel-preset": "^0.0.1",
+    "@priceline/eslint-config": "^0.0.1",
+    "@testing-library/jest-dom": "^5.11.4",
+    "@testing-library/react": "^11.0.4",
+    "cross-env": "^7.0.2",
+    "eslint": "^7.9.0",
+    "jest": "^26.4.2",
+    "jest-standard-reporter": "^1.1.1",
+    "jest-styled-components": "^6.3.4",
+    "npm-run-all": "^4.1.5",
+    "pcln-design-system": "^4.0.0",
+    "pcln-icons": "^4.0.0",
+    "pcln-popover": "^4.0.0",
+    "prop-types": "^15.7.2",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
+    "rimraf": "^3.0.2",
+    "styled-components": "^4.4.1",
+    "styled-system": "^4.2.4"
+  },
+  "peerDependencies": {
+    "pcln-design-system": "^4.0.0",
+    "pcln-icons": "^4.0.0",
+    "pcln-popover": "^4.0.0",
+    "styled-components": ">=4.4.1",
+    "react": "^16.10.0",
+    "react-dom": "^16.10.0"
+  }
+}
--- a/packages/menu/package.json
+++ b/packages/menu/package.json
@@ -22,7 +22,6 @@
   "files": [
     "dist/"
   ],
-  "keywords": [],
   "author": "Priceline",
   "license": "MIT",
   "devDependencies": {
--- a/packages/menu/src/Menu/Menu.js
+++ b/packages/menu/src/Menu/Menu.js
@@ -1,6 +1,7 @@
-import React from 'react'
+import React, { useCallback } from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
+import { themeGet } from 'styled-system'
 import { Button } from 'pcln-design-system'
 import { ChevronDown } from 'pcln-icons'
 import Popover from 'pcln-popover'
@@ -9,13 +10,13 @@ import MenuList from '../MenuList'
 const LinkButton = styled(Button)`
   display: flex;
   align-items: center;
-  padding: 8px;
+  padding: ${themeGet('space.2')}px;
 `
 
-function Menu({ buttonText, color, height, size, children, ...props }) {
+function Menu({ id, buttonText, color, height, size, children, ...props }) {
   const MenuContent = ({ handleClose }) => (
     <MenuList
-      id='menu-container'
+      id={id}
       color={color}
       size={size}
       height={height}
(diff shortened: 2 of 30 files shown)


packages/menu/src/Menu/Menu.js
```diff
@@ -0,0 +1,72 @@
...
+    >
+      {children}
+    </MenuList>
+  )
+
+  MenuContent.propTypes = {
+    handleClose: PropTypes.func,
+  }
+
+  const onKeyDown = (evt) => {
```
robnorman-pcln, line 32: `useCallback`?

packages/menu/src/MenuItem/MenuItem.js
```diff
@@ -0,0 +1,71 @@
...
+    background-color: ${getPaletteColor('base')};
+  }
+
+  ${(props) =>
+    props.selected
+      ? `background-color: ${getPaletteColor('background.base')(props)};`
+      : `background-color: ${getPaletteColor('background.lightest')(props)};`}
+`
+
+const MenuItem = React.forwardRef(function MenuItem(
```
robnorman-pcln, line 31: This line is so weird feels like it should complain about `const MenuItem` and `function MenuItem` sharing the same name.

packages/menu/src/MenuItem/MenuItem.js
```diff
@@ -0,0 +1,71 @@
...
+  { id, selected, children, handleClose, onClick, ...props },
+  ref
+) {
+  const handleClick = () => {
```
robnorman-pcln, line 35: useCallback?

packages/menu/src/MenuItem/MenuItem.js
pcln-james: You think it would be it doesn't 🤷‍♂️

packages/menu/src/MenuList/MenuList.js
```diff
@@ -0,0 +1,106 @@
...
+  const nextFocus = traversalFunction(currentFocus)
+  nextFocus && nextFocus.focus()
+}
+
+function MenuList({ children, color, size, height, handleClose }) {
+  const listRef = useRef(null)
+  const currentItemRef = useRef(null)
+
+  useEffect(() => {
+    currentItemRef?.current?.focus()
```
robnorman-pcln, line 50: I believe `current` is always defined on refs so no need for the first `?`

packages/menu/src/MenuList/MenuList.js
```diff
@@ -0,0 +1,106 @@
...
+  }, [])
+
+  function handleKeyDown(evt) {
```
robnorman-pcln: useCallback

robnorman-pcln (approved):

packages/menu/src/Menu/Menu.js
```diff
@@ -0,0 +1,72 @@
...
+const LinkButton = styled(Button)`
+  display: flex;
+  align-items: center;
+  padding: 8px;
+`
+
+function Menu({ buttonText, color, height, size, children, ...props }) {
+  const MenuContent = ({ handleClose }) => (
+    <MenuList
+      id='menu-container'
```
dgiraldo313, line 18: Since there's the possibility that there might be more than menu on the page, we probably need to add a prefix to this id. We can't have repeated ids on the page.

packages/menu/src/MenuList/MenuList.js
```diff
@@ -0,0 +1,106 @@
...
+    <MenuContainer
+      ref={listRef}
+      role='list'
+      flexWrap='wrap'
+      height={height}
+      size={size}
+      p={2}
+      pr={1}
+      pb={1}
+      onKeyDown={handleKeyDown}
```
dgiraldo313, line 78: This needs an `aria-activedescendant` and that needs to be mapped to the id of the current active child.

packages/menu/src/MenuList/MenuList.js
```diff
@@ -0,0 +1,106 @@
...
+    const currentFocus = list?.ownerDocument?.activeElement
+
+    if (key === 'ArrowDown') {
+      evt.preventDefault()
+      moveFocus(currentFocus, focusNext)
+    } else if (key === 'ArrowUp') {
+      evt.preventDefault()
+      moveFocus(currentFocus, focusPrevious)
+    }
+  }, [])
```
dgiraldo313, line 66: So for this, focus actually needs to stay at the list level (The container with the `role="list"`). Each menu item needs to have an `aria-selected` and the state will be updated to false with the logic above. 
Keeping the focus at the list container level allows the screen reader to read which element is currently selected and the position in the list (ie. "USD selected, 1 of 4")

packages/menu/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "pcln-menu",
+  "version": "3.0.0",
```
craigpalermo, line 3: Should we set it to 4 to match the current major version?

packages/menu/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "start": "start-storybook",
+    "test": "jest --verbose --coverage",
+    "test:ci": "jest --silent --coverage --runInBand",
+    "test:update": "jest --updateSnapshot",
+    "test:ci:update": "jest --runInBand --updateSnapshot"
+  },
+  "files": [
+    "dist/"
+  ],
+  "keywords": [],
```
craigpalermo, line 25: Can we either add keywords or remove this field?

packages/menu/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "version": "3.0.0",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "sideEffects": false,
+  "scripts": {
+    "prepare": "npm run clean && npm run build",
+    "build": "npm-run-all -p build:*",
+    "build:cjs": "cross-env NODE_ENV=production BABEL_ENV=cjs babel ./src --ignore '**/*.stories.js,**/*.spec.js' --out-dir ./dist/cjs",
+    "build:esm": "cross-env NODE_ENV=production BABEL_ENV=esm babel ./src --ignore '**/*.stories.js,**/*.spec.js' --out-dir ./dist/esm",
+    "build:es": "cross-env NODE_ENV=production BABEL_ENV=es babel ./src --ignore '**/*.stories.js,**/*.spec.js' --out-dir ./dist/es",
```
craigpalermo, line 12: @sdalonzo side note, we really need to consolidate these under a single script

craigpalermo (changes requested):
Just a couple minor comments before we merge, but great work @pcln-james.

packages/menu/package.json
pcln-james: I thought it would bump to 4 with the major bump, is that the standard with new packages?

craigpalermo: Ohhh my bad, you are correct!

packages/menu/package.json
sdalonzo: oh, you mean port `build-react-component`? yeam we could ticket that

packages/menu/src/Menu/Menu.js
```diff
@@ -0,0 +1,75 @@
...
+
+  MenuContent.propTypes = {
+    handleClose: PropTypes.func,
+  }
+
+  const onKeyDown = useCallback((evt) => {
+    const key = evt.key
+    if (key === 'ArrowDown' || key === 'ArrowUp') {
+      evt.preventDefault()
+      evt.target.click()
```
sdalonzo, line 37: can't you just call the handler here? firing a click on keypress feels wrong.

packages/menu/src/MenuList/MenuList.js
```diff
@@ -0,0 +1,112 @@
...
+  `,
+  twoColumns: css`
+    & > * {
+      width: calc(50% - ${themeGet('space.1')}px);
+    }
+  `,
+}
+
+const MenuContainer = styled(Flex)`
+  font-family: 'Montserrat', 'Helvetica Neue', Helvetica, Arial, sans-serif;
```
sdalonzo, line 21: is this necessary?

packages/menu/src/MenuList/MenuList.js
```diff
@@ -0,0 +1,112 @@
...
+  overflow-y: scroll;
+  height: ${(props) => props.height || 300}px;
```
sdalonzo, line 23: maybe rename this `heightPx`? this collides with the `styled-system` height that accepts percents and em/rem/viewport units

packages/menu/src/MenuList/MenuList.js
pcln-james: Yeah the popover content gets confused, we should create a ticket to look into this

craigpalermo: Not sure if it's strictly necessary, but fwiw, `popover` renders its content in a React portal, so that might have something to do with it.

sdalonzo (approved):
overall LGTM

craigpalermo (approved):

packages/menu/src/Menu/Menu.js
pcln-james: I tried to use useState and isOpen but it breaking some of the default functionality like the escape key closing or clicking on the background. This might be a bug with the Popover, I'm going to leave it how it is for now and add an issue to look into this

packages/menu/src/MenuList/MenuList.js
pcln-james: I just removed this line, since it's a Flex component it accepts `height` as a prop and handles everything correctly

packages/menu/src/MenuList/MenuList.js
dgiraldo313: I thought that had been fixed at some point 🤔

PR #969, opened 2020-11-19
Update createTheme to overwrite MQ based on breakpoints + update applyStyles to read theme from context

dgiraldo313:


packages/core/src/Badge/Badge.js
```diff
@@ -71,8 +71,8 @@ const Badge = styled.div`
   text-transform: uppercase;
   letter-spacing: ${themeGet('letterSpacings.caps')};
   ${space} ${type} ${color};
-  ${applySizes(sizes)}
-  ${applyVariations('Badge')}
+  ${({ theme }) => applySizes(sizes, undefined, theme.mediaQueries)};
```
dgiraldo313, line 74: This ensure that the sizes are true to the current theme in context and not the static sizes from the DS theme.

packages/core/src/utils/createTheme.js
```diff
@@ -151,11 +151,14 @@ const createPalette = ({ palette = {}, ...theme }) => {
...
   const mergedTheme = deepmerge(defaultTheme, theme)
 
+  const mediaQueries = customBreakpoints
+    ? createMediaQueries(customBreakpoints)
+    : mergedTheme.mediaQueries
   return {
     ...mergedTheme,
     contrastRatio: mergedTheme.contrastRatio || 2.6,
     breakpoints: addAliases(customBreakpoints || mergedTheme.breakpoints),
-    mediaQueries: addAliases(mergedTheme.mediaQueries),
```
dgiraldo313, line 158: With this previous logic, only the breakpoints were updated. This means that any DS component would adhere to those new breakpoints. However, the media queries we still the same as the default theme, so they new breakpoints do not work when accessing the diff media query inside a styled component.

sdalonzo (commented):
overall LGTM. made a few minor suggestions in Slack. withholding approval for now, since we are going to publish a beta for testing first.

packages/core/src/utils/createMediaQueries.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Create media queries array based on the passed in breakpoints
+ *
+ * @param {Array} breakpoints The passed in breakpoints array
+ *
+ * @returns {Array} The generated media queries array
+ */
+const createMediaQueries = (breakpoints) =>
+  breakpoints.map((n) => `@media screen and (min-width:${n})`)
```
craigpalermo, line 9: Could you add a test or two for this new function?

craigpalermo (commented):
LGTM pending the addition of tests for the updated `createMediaQueries` function.

gbellini911 (approved):

sdalonzo (approved):

craigpalermo (approved):

PR #970, opened 2020-11-25
Add support for Core components written in TypeScript

craigpalermo:
**Changes included:**

- Rewrite `Box` component and related files in TypeScript
- Update Storybook project to SB 6 for the built-in TS support. This also fixes the broken docs pages in the existing storybook.
- Delete unused `types` project

Tracking [this issue](https://github.com/storybookjs/storybook/issues/13304) related to prop table on docsite for TS components and [this issue](https://github.com/storybookjs/storybook/issues/11933) related to missing docs page props for styled components.

**Todo**

- [x] Fix error in `rush lint:fix`
- [x] Run `rush change`
- [ ] Confirm that prop types appear on Storybook Docs page for `Box`, as well as JS components

PR #971, opened 2020-11-29
Convert all stories to CSF

craigpalermo:
This PR is prep for #970 to make it easier to review. Long story short, the easiest way to add support for TypeScript in the `core` package requires upgrading to Storybook 6 because it provides zero-config TS support. In order to do that, we need to remove usages of the deprecated `@storybook/addon-info`, and we do that by replacing those usages with `parameters.docs.description.component` in the default export for CSF story files. Obviously, this requires refactoring all stories to CSF, hence this PR.

This PR **does not** upgrade to SB6, it's just prep and will not change any functionality. The docs pages of all stories are still broken.

- [x] run `rush change`

### DIFF
--- a/apps/storybook/package.json
+++ b/apps/storybook/package.json
@@ -24,7 +24,6 @@
     "@storybook/addon-a11y": "^5.3.21",
     "@storybook/addon-actions": "^5.3.21",
     "@storybook/addon-docs": "^5.3.21",
-    "@storybook/addon-info": "^5.3.21",
     "@storybook/addon-knobs": "^5.3.21",
     "@storybook/addon-viewport": "^5.3.21",
     "@storybook/react": "^5.3.21",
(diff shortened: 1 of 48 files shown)


pcln-james (approved):
LGTM

pcln-barry (approved):

packages/core/test/jest.setup.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
sdalonzo, line 1: why convert this at all, if all we have is is `@tsignore`s and no value added?

packages/core/test/testing-library.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { ThemeProvider } from '../src'
+
+const WithThemeProvider = ({ children }) => (
+  <ThemeProvider>{children}</ThemeProvider>
+)
+WithThemeProvider.propTypes = {
+  children: PropTypes.node,
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
sdalonzo, line 13: same thing here. seems silly to update this file from JS to TS if all we are doing is adding noise.

sdalonzo (approved):

PR #978, opened 2020-12-04
Fix bug when passing aria-label to Button and related components

craigpalermo:
This was an unintended regression in DS4. Both the `title` and `aria-label` props will now set the `aria-label` on the rendered `<button>` (`aria-label` prop takes precedence).

### DIFF
--- a/packages/core/src/IconButton/IconButton.stories.js
+++ b/packages/core/src/IconButton/IconButton.stories.js
@@ -13,6 +13,7 @@ export default {
 export const Default = () => (
   <IconButton
     onClick={action('Clicked IconButton')}
+    aria-label={'Click to choose date'}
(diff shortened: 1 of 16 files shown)


pcln-james (approved):

packages/core/src/IconButton/IconButton.stories.js
```diff
@@ -13,6 +13,7 @@ export default {
 export const Default = () => (
   <IconButton
     onClick={action('Clicked IconButton')}
+    aria-label={'Click to choose date'}
```
sdalonzo, line 16:
```suggestion
    aria-label='Click to choose date'
```

sdalonzo (approved):

PR #979, opened 2020-12-05
Pass parent theme to popover content

### LINKED ISSUE #976: pcln-popover content drops theme (at least the font family)
**Describe the bug**
Popover content doesn't inherit the theme properly, this is visible due to the font family not appearing on Text components inside of the popover content unless you specifically set it using styled-components.

pcln-popover

**To Reproduce**
Create a popover component with some text in the content.

**Expected behavior**
Text should be using Montserrat font family

**Screenshots**

Current Behaviour:
image

Expected Behaviour:
image

**Desktop (please complete the following information):**

- OS: MacOS 10.15.7
- Browser: Chrome
- Version: pcln-popover@4.0.0

I believe this issue is a result of ThemeProvider not wrapping the ReactPortal that is created by the Popover but not exactly sure of the best solution.

craigpalermo:
Resolves #976

### DIFF
--- a/packages/popover/src/Popover/Popover.spec.js
+++ b/packages/popover/src/Popover/Popover.spec.js
@@ -67,7 +67,7 @@ describe('Popover', () => {
       expect(asFragment()).toMatchSnapshot()
     })
 
-    it('renders content with correct theme', () => {
+    it('renders content with nested ThemeProviders', () => {
       const expectedColor = 'red'
       const { getByText } = render(
         <ThemeProvider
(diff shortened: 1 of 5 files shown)


pcln-james (approved):

packages/popover/src/Popover/Popover.spec.js
```diff
@@ -67,6 +67,30 @@ describe('Popover', () => {
       expect(asFragment()).toMatchSnapshot()
     })
 
+    it('renders content with correct theme', () => {
```
craigpalermo, line 70:
```suggestion
    it('renders content with nested ThemeProviders', () => {
```

sdalonzo (approved):

Issue #606, opened 2019-10-10
Add Label and Helper Text to Input and other components


James300:
Instead of wrapping the component with a Label we should be able to pass a label prop that does it for us. I would expect similar behaviour for helper text too. This should help reduce the verboseness of our current implementation.

![image](https://user-images.githubusercontent.com/9325039/66676508-a9efd600-ec35-11e9-89d4-b018bcc4a4c5.png)

unixchad:
First phase we could just add the `helperText` prop to get the ball rolling.

@James300 Are you thinking something like this?

```
function Input(props) {
  const { helperText, ...rest } = props;

  return (
    <Fragment>
      <input {...rest} />
      {helperText && <div>{helperText}</div>}
    </Fragment>
  );
}
```

PR #982, opened 2020-12-06
Add helperText prop to Input

craigpalermo:
Begins to address the ask in #606 by adding a `helperText` prop for displaying error messages below inputs.

image


packages/core/src/Input/Input.js
```diff
@@ -38,6 +40,21 @@ const Input = styled.input`
   ${applyVariations('Input')}
 `
 
+export const Input = (props) => {
+  const { helperText, color, ...restProps } = props
+  const borderColor = (helperText && 'error.base') || color
+  return (
+    <>
+      <StyledInput {...restProps} color={borderColor} />
+      {helperText && (
```
craigpalermo, line 49: @dgiraldo313 any accessibility concerns we should be aware of here?

craigpalermo:
As discussed in the weekly meeting, I'm going to update this so `helperText` supports a React Node as a prop instead of a string, then add exports for the following three default flavors of helper text: `Error`, `Warning`, and `Info`.

packages/core/src/Input/Input.stories.js
```diff
@@ -41,6 +41,45 @@ export const WithExternalLabel = () => (
...
+        helperText={<Input.AlertText>Broccoli? Vile weed!</Input.AlertText>}
+      />
+    </Box>
+    <Divider />
+    <Box>
+      <Label fontSize={4} htmlFor='sample-input-3'>
+        Info text!
+      </Label>
+      <Input
+        id='sample-input'
```
craigpalermo, line 73: Remove the ID

packages/core/src/Input/Input.spec.js
```diff
@@ -23,4 +24,42 @@ describe('Input', () => {
...
+      <Input
+        id={id}
+        helperText={<Input.ErrorText>{helperText}</Input.ErrorText>}
+      />
+    )
+    let helperTextNode = getByText(helperText)
+    expect(helperTextNode).not.toBeNull()
+    expect(helperTextNode).toHaveStyleRule(
+      'color',
+      getPaletteColor('error.base')({ theme: paletteTheme })
```
pcln-james, line 40: This is a much better way to test that colours match, I wonder if we can make a testing specific helper for this.

packages/core/src/InputGroup/InputGroup.js
```diff
@@ -18,7 +21,7 @@ const InputGroup = styled.div`
     flex: 1 1 auto;
   }
 
-  & ${Input} {
+  & ${StyledInput} {
```
pcln-james, line 24: could we change this to `& input` since it'll still select the input element?

packages/core/src/Input/Input.js
```diff
@@ -38,6 +40,63 @@ const Input = styled.input`
...
+    }
+    switch (helperText.type.displayName) {
+      case INPUT_ERROR_TEXT:
+        return 'error.base'
+      case INPUT_ALERT_TEXT:
+        return 'alert.base'
+      case INPUT_INFO_TEXT:
+        return 'text.light'
+    }
+  }
```
pcln-james, line 62: Sorry for playing you about so much with this but I feel like the HelperText should inherit the color from the Input which an option to override if the dev really needs to.

Something like:
`<Input color='error' helperText={<Input.HelperText color='error.dark' />}  />`

And the HelperText could just preconfigure fontSize, weight, margins and other necessary things but all overridable

packages/core/src/InputGroup/InputGroup.js
craigpalermo: Huh good question, I can try that and see what happens.

pcln-james (approved):

pcln-barry (approved):

packages/core/src/Input/Input.spec.js
```diff
@@ -23,4 +24,55 @@ describe('Input', () => {
     ).toJSON()
     expect(json).toMatchSnapshot()
   })
+
+  describe('helper text', function () {
+    test('renders with same color as Input', () => {
+      const helperText = 'hello world'
+      const color = 'error.base'
+      const paletteTheme = createTheme(theme)
```
sdalonzo, line 32: you don't need theme. `createTheme()` will do the trick

PR #990, opened 2020-12-19
Change withComponent to as prop

### LINKED ISSUE #989: Remove use of deprecated with withComponent utility
Styled Components will remove the `withComponents` utility in a future release, so we should replace all usages of it with the `as` prop.

One known usage is in `Text`: https://github.com/priceline/design-system/blob/master/packages/core/src/Text/Text.js#L108.

Read more about the planned deprecation here: https://styled-components.com/docs/api#withcomponent

pcln-james:
Resolves #989

### DIFF
--- a/packages/core/src/Text/Text.js
+++ b/packages/core/src/Text/Text.js
@@ -111,15 +111,20 @@ Text.span = ({ children, ...props }) => (
     {children}
   </Text>
 )
+Text.span.displayName = 'Span'
+
 Text.p = ({ children, ...props }) => (
   <Text as='p' {...props}>
     {children}
   </Text>
 )
+Text.p.displayName = 'Paragraph'
+
 Text.s = ({ children, ...props }) => (
   <Text as='s' {...props}>
     {children}
   </Text>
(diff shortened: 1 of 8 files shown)


packages/core/src/Text/Text.js
```diff
@@ -105,8 +106,20 @@ Text.propTypes = {
...
 
-Text.span = Text.withComponent('span')
-Text.p = Text.withComponent('p')
-Text.s = Text.withComponent('s')
+Text.span = ({ children, ...props }) => (
+  <Text as='span' {...props}>
+    {children}
+  </Text>
+)
+Text.p = ({ children, ...props }) => (
```
craigpalermo, line 114: I think these are valid warnings, can you add a `displayName` to each of these?

craigpalermo (changes requested):
LGTM so far, just one minor comment and need to create a change.

- [ ] Run `rush change`

craigpalermo (approved):

gbellini911 (approved):

PR #991, opened 2020-12-21
add generic banner

gbellini911:
Add GenericBanner to DS


packages/core/src/Chip/Chip.spec.js
```diff
@@ -13,20 +13,24 @@ function useToggle(initialValue = false) {
 
 describe('Chip', () => {
   it('renders filter with close icon if selected', () => {
-    render(<Chip selected>Free Wifi</Chip>)
+    render(
```
gbellini911, line 16: just added some ids here to clean up warnings when running the test suite

packages/core/src/ShadowEffect/ShadowEffect.spec.js
```diff
@@ -9,7 +9,7 @@ describe('ShadowEffect', () => {
     const mockOnClick = jest.fn()
     render(
       <ShadowEffect onOpen={mockOnOpen} data-testid='overlay'>
-        <Input onClick={mockOnClick} data-testid='input' />
+        <Input onClick={mockOnClick} id='input' data-testid='input' />
```
gbellini911, line 12: also just added some ids here to clean up warnings when running the test suite

packages/core/src/GenericBanner/GenericBanner.js
```diff
@@ -0,0 +1,124 @@
...
+        <Flex alignItems='center' justifyContent={alignment}>
+          {!!iconLeft && iconLeft}
+          {!!imageLeft && imageLeft}
+          <Text px={2} fontSize={textFontSize}>
+            {!!heading && heading}
+            {!!text && text}
+
+            {URLProps && ctaText && (
+              <CustomLink fontSize={textFontSize} {...URLProps}>
+                {!!ctaText && ctaText}
```
gbellini911, line 72: passed ctaText as a node here, but ideally for accessibility, the node <Text> passed in should have an underline and hover style to let the user know it's an action item. So open to suggestions here. I could change this so it's always an UnderlinedLinkText like below, but also wanted to give more flexibility in what's passed in

```
const UnderlinedLinkText = styled(Text)`
  z-index: 2;
  text-decoration: underline;
  &:hover {
    color: ${(props) => getPaletteColor(props.hoverColor)};
  }
`
```

packages/core/src/GenericBanner/GenericBanner.js
```diff
@@ -0,0 +1,122 @@
...
+        </Flex>
+      </BannerWithRadius>
+    </Relative>
+  )
+}
+
+GenericBanner.displayName = 'GenericBanner'
+
+GenericBanner.propTypes = {
+  alignment: PropTypes.string,
```
craigpalermo, line 96: I think it'd be helpful to include the list of possible values for `alignment` in the prop type.

packages/core/src/GenericBanner/GenericBanner.js
```diff
@@ -0,0 +1,122 @@
...
+  text: PropTypes.node,
+  textFontSize: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+    PropTypes.arrayOf(
+      PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+    ),
+  ]),
+  buttonClick: PropTypes.func,
+  URLProps: PropTypes.shape({
```
craigpalermo, line 111: Can you add doc comments so we'll know what each prop is for on the Storybook docs page?

packages/core/src/GenericBanner/GenericBanner.js
```diff
@@ -0,0 +1,122 @@
...
+
+GenericBanner.propTypes = {
+  alignment: PropTypes.string,
+  ctaText: PropTypes.node,
+  heading: PropTypes.node,
+  iconLeft: PropTypes.node,
+  iconRight: PropTypes.node,
+  imageLeft: PropTypes.node,
+  text: PropTypes.node,
+  textFontSize: PropTypes.oneOfType([
```
craigpalermo, line 103: It's probably safe to just call this prop `fontSize` since the same value applies to text, link, and button in this component.

packages/core/src/GenericBanner/GenericBanner.spec.js
```diff
@@ -0,0 +1,78 @@
...
+      Read More
+    </Text.span>
+  ),
+  imageLeft: <Image alt='ImageLeft' />,
+  iconLeft: <Warning />,
+  iconRight: <Emoticon />,
+  color: 'caution.light',
+}
+
+describe('GenericBanner', () => {
```
craigpalermo, line 36: Can you add a test to assert that `fontSize` is being passed along everywhere that we expect it to be?

packages/core/src/GenericBanner/GenericBanner.js
gbellini911: Sure, I'll update :)

packages/core/src/GenericBanner/GenericBanner.stories.js
```diff
@@ -0,0 +1,154 @@
...
+    iconLeft={<Warning />}
+    color='caution.light'
+  />
+)
+export const basicBannerAllInlineRightIcon = () => (
+  <GenericBanner
+    p={2}
+    heading={
+      <Text.span fontWeight='bold' color='secondary'>
+        This is a Heading&nbsp;&nbsp;
```
craigpalermo, line 45: n-b it looks like your editor might have converted some unneeded whitespace

craigpalermo (changes requested):
Nice work so far! Just a few minor comments to help developer ergonomics and to add some tests.

packages/core/src/GenericBanner/GenericBanner.stories.js
craigpalermo: Ah I see what those were for; replaced with `mr` to make it a bit cleaner.

craigpalermo (approved):

packages/core/src/GenericBanner/GenericBanner.stories.js
gbellini911: Yes exactly! Sorry left this comment open so I could respond saying that and then forgot to actually write that comment 😆

packages/core/src/GenericBanner/GenericBanner.spec.js
```diff
@@ -0,0 +1,80 @@
...
+import { Warning, Emoticon } from 'pcln-icons'
+import { Text } from '../Text'
+import { Image } from '../Image'
+import { GenericBanner } from '.'
+
+const props = {
+  p: 2,
+  heading: (
+    <Text.span fontWeight='bold' textColor='primary.base'>
+      COVID-19&nbsp;&nbsp;
```
pcln-james, line 12: Do we need these backspaces? Can we change them to margin right?

packages/core/src/GenericBanner/GenericBanner.spec.js
```diff
@@ -0,0 +1,80 @@
...
+    </Text.span>
+  ),
+  text: (
+    <Text.span textColor='primary.base'>
+      Update: Your travel may be impacted. Please review this hotels important
+      info.&nbsp;&nbsp;
```
pcln-james, line 18: Same here

packages/core/src/ShadowEffect/ShadowEffect.spec.js
pcln-james: Thanks for improving my tests

pcln-james (approved):
LGTM couple of non blocking comments if you feel like it

PR #994, opened 2020-12-24
Add Accessibility SkipMenu component to DS

dgiraldo313:
Hidden menu that appears on the screen when users focus into any of the links inside of it. This is a common practice to allow keyboard users to be able to find a menu that will let them navigate to different sections on the page, without having to tab their way to that particular section.

Screen Shot 2020-12-24 at 12 30 48 PM


packages/core/src/SkipMenu/SkipMenu.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
+import styled from 'styled-components'
+import PropTypes from 'prop-types'
+import { Flex } from '../Flex'
+import { Link } from '../Link'
+import { SrOnly } from '../SrOnly'
+
+const OffScreenPanel = styled(SrOnly)`
+  &:focus-within {
```
dgiraldo313, line 9: This css attribute is not supported by IE11, so if we are planning on supporting that browser we will need to figure out an alternative.

craigpalermo: Maybe just add a comment denoting this for future reference

apps/docs/pages/SkipMenu.md
```diff
@@ -0,0 +1,23 @@
+# SkipMenu
+
+Use the `SkipMenu` at the top of the page to allow keyboard users to easily skip to different sections on the page.
+
+Additionally, it supports `Box` props for styling menu wrapper and each link also support `Link` props.
```
craigpalermo, line 5:
```suggestion
Additionally, it supports `Box` props for styling menu wrapper and each link also supports `Link` props.
```

packages/core/src/SkipMenu/SkipMenu.js
```diff
@@ -0,0 +1,54 @@
...
+    <OffScreenPanel
+      data-testid='skip-menu'
+      as={Flex}
+      className={className}
+      {...props}
+    >
+      {skipLinks?.map(({ label, targetId, ...skipLink }) => (
+        <Link
+          key={`skipLink_${targetId}`}
+          href={`#${targetId}`}
```
craigpalermo, line 31: n-b maybe as a future enhancement we log an error if the `targetId` provided is already prefixed with `#`, or we could handle it gracefully in the component.

apps/docs/pages/SkipMenu.md
```diff
@@ -0,0 +1,23 @@
...
+	}
+]
+<SkipMenu skipLinks={skipLinks} />
+```
+
+## Props
+
+| Prop        | Type   | Description                         |
+| ----------- | ------ | ----------------------------------- |
+| `skipLinks` | Array  | Object container link info targetId)|
```
craigpalermo, line 23:
```suggestion
| `skipLinks` | Array  | Array of objects containing `label` and `targetId` |
```

packages/core/src/SkipMenu/SkipMenu.spec.js
```diff
@@ -0,0 +1,21 @@
...
+  })
+
+  it('should render all links in skipLinks list', () => {
+    const skipLinks = [
+      { label: 'link 1', targetId: 'link1' },
+      { label: 'link 2', targetId: 'link2' },
+    ]
+    const { getByText } = render(<SkipMenu skipLinks={skipLinks} />)
+    expect(getByText('link 1')).toBeTruthy()
+    expect(getByText('link 2')).toBeTruthy()
```
craigpalermo, line 19: Can you add assertions to check the `href` attribute for each link as well?

packages/core/src/SkipMenu/SkipMenu.stories.js
```diff
@@ -0,0 +1,43 @@
...
+    label: 'Skip to filters',
+    targetId: 'filter-section',
+  },
+  {
+    label: 'Skip to footer',
+    targetId: 'footer',
+  },
+]
+
+storiesOf('SkipMenu', module).add('default', () => (
```
craigpalermo, line 25: Mind changing this to CSF?

craigpalermo (changes requested):
Looking good! Just suggested a couple minor tweaks and additions.

packages/core/src/SkipMenu/SkipMenu.stories.js
dgiraldo313: Yeap. Just FYI this is the template that you get when you run `rush create`, so maybe that needs to be updated.

craigpalermo (approved):

pcln-james (approved):

PR #995, opened 2020-12-28
Fix cyclic dependency that will cause pcln-icons tests to fail in future

craigpalermo:
While working on #970, I noticed that `pcln-icons` tests failed after upgrading to `pcln-design-system@^4.2.1`. It was previously on `pcln-design-system@^3`, and upgrading seemed to cause an inexplicable error when running `rush update` related to an already existent symlink to `packages/icons/node_modules/pcln-design-system/dist/cjs/index.js`.

**This PR includes the following changes:**

- Upgrade to latest Rush and PNPM
- ~Use PNPM workspaces to resolve cyclic dependency issues~
- Upgrade to latest `pcln-design-system` in devDeps of `pcln-icons`
- Update allowed peerDep of `react` and `react-dom` in `babel-preset` to match that of the rest of the packages in the repo. This mismatch was causing a React invalid hooks error when running tests for `pcln-icons` because there would be multiple versions of React in the dependency tree.
- Move require for `@rushstack/eslint-patch` to individual project `.eslintrc.js` files as recommended in their README


pcln-james (approved):

sdalonzo (changes requested):
As discussed, waiting to see the `rush-lib` bump as well as seeing if the snapshots calm down after a rebase.

packages/core/package.json
```diff
@@ -73,7 +73,7 @@
     "deepmerge": "^4.2.2",
     "hoist-non-react-statics": "^3.3.2",
     "moize": "^5.4.7",
-    "pcln-icons": "4.0.0",
+    "pcln-icons": "4.0.1-holiday-2020.1",
```
craigpalermo, line 76: This was necessary to apply the latest changes from `pcln-icons` due to the cyclic dependency between `pcln-design-system` and `pcln-icons`. It should be inconsequential as it will be replaced when we apply changes prior to the next release.

sdalonzo (approved):

PR #1001, opened 2021-01-04
Apply changes for core 4.3.0

craigpalermo:
## 4.3.0 (tentative)
Mon, 04 Jan 2021 16:21:42 GMT

### Minor changes

- Add Shimmer component
- Add Accessibility SkipMenu component
- Add helperText prop for showing error messages in forms
- add a new component GenericBanner to core

### Patches

- Fix bug when using RatingBadge with color but not bg prop
- Resolve warnings in Storybook and remove use of deprecated styled().attrs({}) syntax
- Update Text Extension Display Names
- Change withComponent to as


packages/autocomplete/package.json
```diff
@@ -55,7 +55,7 @@
     "@rushstack/eslint-config": "^2.3.1"
   },
   "peerDependencies": {
-    "pcln-design-system": "^4.2.1",
+    "pcln-design-system": "4.3.0",
```
sdalonzo, line 58: peers should never be this specific

packages/menu/package.json
```diff
@@ -48,9 +48,9 @@
     "@rushstack/eslint-config": "^2.3.1"
   },
   "peerDependencies": {
-    "pcln-design-system": "^4.2.1",
-    "pcln-icons": "^4.0.0",
-    "pcln-popover": "^4.0.1",
+    "pcln-design-system": "4.3.0",
```
sdalonzo, line 51: same here.

packages/modal/package.json
```diff
@@ -51,7 +51,7 @@
     "@rushstack/eslint-config": "^2.3.1"
   },
   "peerDependencies": {
-    "pcln-design-system": "^4.2.1",
+    "pcln-design-system": "4.3.0",
```
sdalonzo, line 54: and here

packages/popover/package.json
```diff
@@ -57,7 +57,7 @@
     "@rushstack/eslint-config": "^2.3.1"
   },
   "peerDependencies": {
-    "pcln-design-system": "^4.2.1",
+    "pcln-design-system": "4.3.0",
```
sdalonzo, line 60: same

packages/slider/package.json
```diff
@@ -63,7 +63,7 @@
     "@rushstack/eslint-config": "^2.3.1"
   },
   "peerDependencies": {
-    "pcln-design-system": "^4.2.1",
+    "pcln-design-system": "4.3.0",
```
sdalonzo, line 66: you get the idea ;)

sdalonzo (changes requested):
Peers need to be careted. is 4.3 actually required?

craigpalermo:
> Peers need to be careted. is 4.3 actually required?

I don't think `4.3.0` is strictly required by any sub packages, but I think it makes sense to keep the peers up-to-date to encourage consumers to keep all of their DS-related packages at the latest instead of mixing versions within the same major because it seems like less work.

sdalonzo (approved):

packages/core/CHANGELOG.json
```diff
@@ -1,6 +1,57 @@
...
+        ],
+        "patch": [
+          {
+            "comment": "Fix bug when using RatingBadge with color but not bg prop",
+            "author": "Craig Palermo <3260642+craigpalermo@users.noreply.github.com>",
+            "commit": "366e6521107e6ccf61b2c59cfed536067a6eed55"
+          },
+          {
+            "comment": "Resolve warnings in Storybook and remove use of deprecated styled().attrs({}) syntax",
+            "author": "Craig Palermo <3260642+craigpalermo@users.noreply.github.com>",
```
pcln-james, line 39: weird most of our emails didn't work

pcln-james (approved):
LGTM but we should look into the email issue

craigpalermo:
> LGTM but we should look into the email issue

Oh yeah hmm, agreed. We'll have to look into that.

PR #1005, opened 2021-01-21
Fix ref forwarding for Input component

craigpalermo:
Ref forwarding broke for the `Input` component as a result of this change: https://github.com/priceline/design-system/pull/982/files#diff-7fcb0b1ea1caf8fd0ac3152da8bef7153ba71b396215388de930d7907f5c21e3R45

This restores ref forwarding.


packages/core/src/Input/Input.spec.js
```diff
@@ -1,7 +1,8 @@
 import React from 'react'
-import { render } from 'testing-library'
+import { render, fireEvent } from 'testing-library'
+import ForwardRefDemo from '../../storybook/utils/ForwardRefsDemo'
 
-import { Input, theme, createTheme, getPaletteColor } from '..'
+import { Input, theme, createTheme, getPaletteColor, Button } from '..'
```
pcln-james, line 5: Craig you're killing my OCD

pcln-james (approved):
LGTM besides the linter breakages

abautist (approved):

packages/core/src/Input/Input.spec.js
craigpalermo: Your OCD can now rest easy

sdalonzo (approved):

PR #1006, opened 2021-02-01
Update BackgroundImage to accept height as a number

pcln-james:


pcln-barry (approved):

pcln-barry (approved):

packages/core/src/BackgroundImage/BackgroundImage.js
```diff
@@ -31,8 +29,12 @@ BackgroundImage.defaultProps = {
 }
 
 BackgroundImage.propTypes = {
+  height: PropTypes.oneOfType([
+    PropTypes.number,
+    PropTypes.string,
+    PropTypes.array
+  ]),
```
craigpalermo, line 36: Could we replace this with `...height.propTypes`? Probably wouldn't hurt to do the same for `width`.

https://github.com/priceline/design-system/blob/7096fa92411261a21adf0c11c911d9e4a178c51e/packages/core/src/Text/Text.js#L91

packages/core/src/BackgroundImage/BackgroundImage.stories.js
```diff
@@ -64,6 +64,18 @@ export const FixedHeight = () => (
...
+      <Box p={4}>
+        <Text fontSize={6} bold textAlign='center' color='white'>
+          Hello
+        </Text>
+      </Box>
+    </BackgroundImage>
+  </Box>
+)
+
 export const Responsive = () => (
```
craigpalermo, line 79: Can you add a responsive array for the `height` prop so we have an example of that?

craigpalermo (changes requested):

craigpalermo (approved):

PR #1009, opened 2021-02-05
add check for ctaText

gbellini911:
In GenericBanner:
- allow the heading size to be different than the rest of the text fontSize
-Move the check for the ctaText prop into a better spot so an empty Button or Link doesn't render

### DIFF
--- a/packages/core/src/GenericBanner/GenericBanner.js
+++ b/packages/core/src/GenericBanner/GenericBanner.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { alignItems, justifyContent } from 'styled-system'
+import { alignItems, fontSize, justifyContent } from 'styled-system'
 import PropTypes from 'prop-types'
 import { Banner } from '../Banner'
 import { Box } from '../Box'
@@ -65,14 +65,14 @@ const GenericBanner = ({
           <Box px={2} fontSize={fontSize}>
             {!!heading &&
               React.cloneElement(heading, {
-                fontSize,
+                fontSize: heading.props.fontSize || fontSize,
               })}
             {!!text &&
               React.cloneElement(text, {
                 fontSize,
               })}
 
-            {URLProps && (
+            {URLProps && ctaText && (
               <CustomLink
                 p={
                   linkVariation === 'fill' || linkVariation === 'outline'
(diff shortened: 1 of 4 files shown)


packages/core/src/GenericBanner/GenericBanner.stories.js
```diff
@@ -217,6 +217,30 @@ export const bannerWithURLProps = () => (
   />
 )
 
+export const bannerWithURLPropsAndCustomHeadingSize = () => (
+  <GenericBanner
+    p={2}
+    heading={
+      <Text.span fontWeight='bold' textColor='primary.base' mr={1}>
```
pcln-james, line 224: does adding a fontSize onto this Text component here achieve the same thing or does the GenericBanner's fontSize prop overwrite it?

gbellini911: no, it gets overwritten because the fontSize prop in GenericBanner is being passed in to heading like this : 
`{
        !!heading &&
          React.cloneElement(heading, {
            fontSize: headingFontSize || fontSize,
          })
      }	 `

gbellini911: so by giving it another option with headingFontSize is the best non breaking change I came up with haha

pcln-james: Ah okay that makes sense. Something I did for the Menu component might work here and it saves us adding another prop `!!heading && React.cloneElement(heading, { fontSize: heading.props.fontSize || fontSize }) }`

MenuList example: https://github.com/priceline/design-system/blob/master/packages/menu/src/MenuList/MenuList.js#L87

pcln-james: I don't really like doing `heading.props.fontSize` as it just feels a bit weird to me but that might just be because you don't see this very often 🤷

packages/core/src/GenericBanner/GenericBanner.js
```diff
@@ -130,6 +129,13 @@ GenericBanner.propTypes = {
     ),
   ]),
   heading: PropTypes.node,
+  headingFontSize: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+    PropTypes.arrayOf(
+      PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+    ),
+  ]),
```
craigpalermo, line 138: Can we use the `fontSize` prop types like we do in [Text](https://github.com/priceline/design-system/blob/master/packages/core/src/Text/Text.js#L89)?

common/changes/pcln-design-system/generic-banner-fix_2021-02-05-13-58.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "pcln-design-system",
+      "comment": "move where ctaText check is passed and add headerFontSize prop to GenericBanner",
+      "type": "minor"
```
pcln-james, line 6: Sorry for being difficult but I think this can be a patch change now that no new props are added 😄

pcln-james (approved):
LGTM! Can probably be a patch instead of a minor but not blocking

common/changes/pcln-design-system/generic-banner-fix_2021-02-05-13-58.json
gbellini911: SO DIFFICULT. but jk you're right. I will update

craigpalermo (approved):

PR #1010, opened 2021-02-05
add new button variations

gbellini911:
[Invision](https://priceline.invisionapp.com/d/main/#/console/17267042/444026802/preview) for different button variations

Added new variations: plain, lightFill and subtle to Button
Added css styling for focus state in Fill variation
(Will likely be renaming the lightFill variation after discussing with Wes)

### DIFF
--- a/apps/docs/pages/Button.md
+++ b/apps/docs/pages/Button.md
@@ -29,7 +29,7 @@ Buttons make common actions immediately visible and easy to perform with one cli
 | `radius`    | string                   | Custom button radius in pixel                  |
 | `size`      | string                   | Sets button size to small, medium or large     |
 | `width`     | number, string, or array | Sets button width                              |
-| `variation` | `fill`, `outline`        | The button variation                           |
+| `variation` | `fill`, `link`, `outline`, `plain`, `primary`, `secondary`, `subtle`, `warning`        | The button variation                           |
 | `color`     | string                   | The color of the button, defaults to `primary` |
 | `disabled`  | bool                     | Disabled the button and apply light color      |
 
--- a/apps/docs/pages/Button.md
+++ b/apps/docs/pages/Button.md
@@ -24,14 +24,14 @@ Buttons make common actions immediately visible and easy to perform with one cli
 
 ## Props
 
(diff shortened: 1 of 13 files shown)


packages/core/src/Button/Button.js
```diff
@@ -65,22 +78,71 @@ const variations = {
...
-  link: css`
-    -webkit-font-smoothing: inherit;
-    font-weight: ${(props) => props.theme.fontWeights.medium};
-    color: ${getPaletteColor('base')};
-    line-height: ${(props) => props.theme.lineHeights.standard};
-    vertical-align: inherit;
-    padding: 0;
+  plain: css`
     background-color: transparent;
+    color: ${(props) => getPaletteColor('primary.base')(props)};
```
achoboter, line 88: Do we want to limit these new variations to always use the same color, instead of the passed in color? What's the purpose of adding `primary` and `secondary` variations when it's the same as (minus the focus state, but that seems like it could be added to `fill`):
```
<Button /> = <Button variation="primary" />
<Button color="secondary" /> = <Button variation="secondary" />
```
Also, don't these new variations make the `color` prop redundant? `<Button variation="secondary" color="primary" />` etc?

pcln-james (approved):
LGTM

apps/docs/pages/Button.md
```diff
@@ -29,7 +29,7 @@ Buttons make common actions immediately visible and easy to perform with one cli
 | `radius`    | string                   | Custom button radius in pixel                  |
 | `size`      | string                   | Sets button size to small, medium or large     |
 | `width`     | number, string, or array | Sets button width                              |
-| `variation` | `fill`, `outline`        | The button variation                           |
+| `variation` | `fill`, `link`, `outline`, `plain`, `primary`, `secondary`, `subtle`, `warning`        | The button variation                           |
```
jes708, line 32: 🙌

jes708 (approved):
Nice work! Can't wait to start using this.

achoboter (approved):
Good work!

pcln-james:
Thanks Jon and Adam I really gave it my all on this one 😝

PR #1013, opened 2021-02-08
Updates to the Chip component to allow more variations

ryan-hunt-priceline:


packages/core/src/index.js
```diff
@@ -8,7 +8,7 @@ export { Box } from './Box'
 export { Breadcrumbs } from './Breadcrumbs'
 export { Button } from './Button'
 export { Card } from './Card'
-export { Chip } from './Chip'
```
jes708, line 11: If we remove Chip, this will need to be released as DS v5. It might be best to also export `FilterChip` as `Chip` to avoid breaking changes

packages/core/src/Chip/ChipContent/ChipContent.js
```diff
@@ -0,0 +1,101 @@
...
+    <ChipContentWrapper
+      height={!children && getSize(size, evaluateChipHeight)}
+      px={getSize(size, evaluatePadding)}
+      disabled={disabled}
+      selected={selected}
+      size={size}
+      {...props}
+    >
+      {children}
+      {!!image && <ImageWrapper disabled={disabled}>{image}</ImageWrapper>}
```
jes708, line 50: I think it would be better to replace all instances of `!!` with `Boolean(` so that you only have to perform one operation instead of 2.

packages/core/src/Chip/ChipContent/ChipContent.js
```diff
@@ -0,0 +1,101 @@
...
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { space, fontSize } from 'styled-system'
+import { ArrowRight } from 'pcln-icons'
+import { ChipContentWrapper } from '../ChipContentWrapper'
+import { Flex } from '../../Flex'
+import { Text } from '../../Text'
+
+const ImageWrapper = styled(Flex)`
+  background-color: white;
```
jes708, line 11: This should use `${getPaletteColor('background.lightest')}`

packages/core/src/Chip/ChipContent/ChipContent.js
```diff
@@ -0,0 +1,101 @@
...
+  border-radius: 6px;
```
jes708, line 12:
```suggestion
  border-radius: ${({ theme }) => theme.radii[2]}px;
```

packages/core/src/Chip/ButtonChip/ButtonChip.js
```diff
@@ -0,0 +1,74 @@
...
+    onChange={onClick}
+    aria-expandaded={expanded}
+    m={m}
+  >
+    <ChipContent
+      label={label}
+      disabled={disabled}
+      selected={expanded}
+      action={showActionIcon && { Icon: ChevronDown, title: getTitle(props) }}
+      {...props}
```
jes708, line 47: Since you‘re passing `m` as a prop to ChipButton, it will also get applied to ChipContent here. Is that intentional?

packages/core/src/Chip/FilterChip/FilterChip.js
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { space, fontSize } from 'styled-system'
+import { Close } from 'pcln-icons'
+import { ChipContent } from '../ChipContent'
+import { ChipLabel } from '../ChipLabel'
+import { ChipInput } from '../ChipInput'
+
+const getTitle = ({ disabled }) => (disabled ? 'Disabled' : 'Close')
```
craigpalermo, line 9: Should this text be driven by a prop? I could see a need to change this to something like "Remove" in the future, so it'd be nice if "Close" came from a default prop.

packages/core/src/index.js
craigpalermo: @ryan-hunt-priceline is the new `FilterChip` identical in functionality to the old `Chip`? It's probably worth a discussion, but we might want to keep the old `Chip` around, make this change purely additive, and mark the old `Chip` for deprecation in v5. Otherwise, we end up with inconsistent naming between the Chip variants.

packages/core/src/Chip/ChipContent/ChipContent.js
```diff
@@ -0,0 +1,101 @@
...
+        <Text fontSize={fontSize} regular ml={!!Icon || !!image ? 2 : 0}>
+          {label}
+        </Text>
+      )}
+      {!!facet && (
+        <Text fontSize={fontSize} regular ml={1}>
+          {facet}
+        </Text>
+      )}
+      {!!bridgeLabel && <ArrowRight title='to' size='16px' ml={2} />}
```
craigpalermo, line 62: It probably wouldn't hurt to make the bridge accept a `node` prop, using `ArrowRight` as the default prop, and cloning whatever node is passed in to add the `title`, `size`, etc.

packages/core/src/Chip/ChipContentWrapper/index.js
```diff
@@ -0,0 +1,45 @@
...
+    background-color: ${getPaletteColor(getBackgroundColor(props))(props)};
+    border: 1px solid ${getPaletteColor(getBorderColor(props))(props)};
+    ${
+      !props.disabled &&
+      `
+      &:hover {
+        border: 1px solid ${getPaletteColor('base')(props)};
+        background-color: ${getPaletteColor('background.lightest')(props)};
+      }
+    `
```
craigpalermo, line 33: Can you change this to a ternary where the false case is `""` so we don't risk weird behavior from Styled Components when it's disabled?

packages/core/src/Chip/ChipLabel/index.js
```diff
@@ -0,0 +1,19 @@
...
+import { Label } from '../../Label'
+
+const ChipLabel = styled(Label)`
+  display: inline-flex;
+  width: auto;
+  user-select: none;
+  padding: 0;
+  margin: 0;
+  > input:focus ~ ${ChipContentWrapper} {
+    box-shadow: 0 0 0 1px ${getPaletteColor('base')};
```
craigpalermo, line 13: Are we able to use a `boxShadow` from theme, or is this custom? https://github.com/priceline/design-system/blob/master/packages/core/src/theme/theme.js#L123

cc @wesleymartin85

packages/core/src/Chip/ChoiceChip/ChoiceChip.stories.js
```diff
@@ -0,0 +1,172 @@
...
+        </Text>
+        <Text textAlign='center' regular fontSize={0}>
+          <Text.span color='green'>$130</Text.span> avg. per night
+        </Text>
+      </Box>
+    </ChoiceChip>
+  </Box>
+)
+
+//Default
```
craigpalermo, line 168:
```suggestion
```

packages/core/src/index.js
ryan-hunt-priceline: Yeah, for now I can export the FilterChip as Chip so it doesn't break current functionality

packages/core/src/Chip/ChipLabel/index.js
ryan-hunt-priceline: @craigpalermo This is custom - Because outline can only have squared edges, this is a replacement, so it just looks like a focus outline with rounded corners

ryan-hunt-priceline: We should consider adding to the theme though

packages/core/src/Chip/ChipContentWrapper/index.js
ryan-hunt-priceline: Updated

packages/core/src/index.js
ryan-hunt-priceline: Updated to export `FilterChip` as default of `Chip`

packages/core/src/Chip/ChipContent/ChipContent.js
ryan-hunt-priceline: Updated

packages/core/src/Chip/ChipContent/ChipContent.js
ryan-hunt-priceline: Updated

packages/core/src/Chip/ChipContent/ChipContent.js
ryan-hunt-priceline: Updated

packages/core/src/Chip/FilterChip/FilterChip.js
ryan-hunt-priceline: Updated - added in prop `actionTitle`

packages/core/src/Chip/ChipContent/ChipContent.js
ryan-hunt-priceline: Updated - added in prop `BridgeIcon`

packages/core/src/Chip/FilterChip/FilterChip.js
```diff
@@ -6,7 +6,8 @@ import { ChipContent } from '../ChipContent'
 import { ChipLabel } from '../ChipLabel'
 import { ChipInput } from '../ChipInput'
 
-const getTitle = ({ disabled }) => (disabled ? 'Disabled' : 'Close')
+const getTitle = ({ disabled, actionTitle }) =>
+  disabled ? 'Disabled' : actionTitle
```
craigpalermo, line 10: Would it make sense to append the word `disabled` after the action title, so it would read something like `<action title> disabled`?

packages/core/src/Chip/ChipContentWrapper/index.js
```diff
@@ -38,6 +54,8 @@ const ChipContentWrapper = styled(Box)`
   align-items: center;
   position: relative;
   border-radius: 2px;
+  ${({ theme, hasChildren }) =>
+    applySizes(getSizes({ hasChildren }), undefined, theme.mediaQueries)};
```
craigpalermo, line 58: @James300 similar to #1025, does the ordering of this `applySizes` rule matter?

jes708 (approved):
Thanks for addressing my concerns. This has my approval 👍

packages/core/src/Chip/FilterChip/FilterChip.js
ryan-hunt-priceline: Updated

sdalonzo (changes requested):
LGTM. Just needs `rush change`

sdalonzo (approved):

PR #998, opened 2020-12-29
Fix bug when using RatingBadge without bg prop

craigpalermo:
Resolves #840 

Screen Shot 2020-12-29 at 10 52 38 AM

PR #1019, opened 2021-02-17
Fix incorrect default color of RatingBadge

craigpalermo:
This broke inadvertently as a result of https://github.com/priceline/design-system/pull/998.

### DIFF
--- a/packages/core/src/RatingBadge/RatingBadge.js
+++ b/packages/core/src/RatingBadge/RatingBadge.js
@@ -15,7 +15,7 @@ const RatingBadge = styled(Box).attrs(({color, bg}) => ({
 RatingBadge.defaultProps = {
   fontWeight: 'bold',
   px: 2,
-  color: 'white',
+  color: 'alert',
   bg: 'orange',
   borderRadius: 1,
 }
--- a/common/changes/pcln-design-system/ratingbadge-default-color_2021-02-17-15-18.json
+++ b/common/changes/pcln-design-system/ratingbadge-default-color_2021-02-17-15-18.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "pcln-design-system",
+      "comment": "Fix incorrect default color of RatingBadge",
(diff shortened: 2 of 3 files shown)


packages/core/src/RatingBadge/RatingBadge.js
```diff
@@ -15,7 +15,7 @@ const RatingBadge = styled(Box).attrs(({color, bg}) => ({
 RatingBadge.defaultProps = {
   fontWeight: 'bold',
   px: 2,
-  color: 'white',
+  color: 'alert',
   bg: 'orange',
```
pcln-james, line 19: should we get rid of `bg` too?

craigpalermo: I think we'd want to leave it until we actually remove the `bg` prop just to be safe

gbellini911: Just confirming to be safe, the text within the RatingBadge will always be white unless styled differently, correct?

gbellini911 (approved):

packages/core/src/RatingBadge/RatingBadge.js
craigpalermo: I don't think it'll alway necessarily be white, e.g. `<RatingBadge color="background.lightest">7.6</RatingBadge>`, but the color prop will behave the same as it does on `Box`.

achoboter (approved):

packages/core/src/RatingBadge/RatingBadge.js
sdalonzo: this is actually a breaking change. should we tee it up for v5?

pcln-james (approved):

sdalonzo (changes requested):
this is a breaking change and we should consider deferring it to v5

sdalonzo (approved):
approved, now that I understand the regression. @craigpalermo will add a note to the release notes of the affected version(s)

PR #1022, opened 2021-02-19
UXPT-1163: progress bar

JonChoi16:
Created progress bar component.
https://priceline.atlassian.net/browse/UXPT-1163


packages/core/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,47 @@
...
+
+function ProgressBar({ steps, stepIndex, stepHeight }) {
+  return (
+    <Flex>
+      {steps.map((step, index) => {
+        const stepColor =
+          index < stepIndex ? steps[stepIndex - 1].color : defaultStepColor
+        return (
+          <CustomBox
+            key={step.color}
```
robnorman-pcln, line 22: Is the `step.color` definitely going to be unique? Would it make sense to use `index` since you're using that for `data-testid` anyway?

pcln-james: Yeah that makes more sense probably, using index in this case when things can't move around is fine.

packages/core/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,47 @@
...
+`
+
+const defaultStepColor = 'background.light'
+
+function ProgressBar({ steps, stepIndex, stepHeight }) {
+  return (
+    <Flex>
+      {steps.map((step, index) => {
+        const stepColor =
+          index < stepIndex ? steps[stepIndex - 1].color : defaultStepColor
```
robnorman-pcln, line 19: Does stepIndex start at 1 and that's why we're doing -1 here?

pcln-james: yeah does that make sense to you? should we remove index from the name since that implies starting from 0, maybe `currentStep`?

packages/core/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { Flex, Box } from 'pcln-design-system'
+
+const CustomBox = styled(Box)`
+  width: 100%;
+  border-radius: 2px;
+  margin-right: 4px;
```
pcln-james, line 9: for width you can use it as a prop on the component `width={1}`
for margin-right you can use the prop `mr={1}` to achieve the same effect

pcln-james:
Don't forget to run `rush change`

packages/core/src/ProgressBar/ProgressBar.js
robnorman-pcln: Yeah either that or just making it start from 0 would make sense to me.

packages/core/src/ProgressBar/ProgressBar.js
```diff
@@ -0,0 +1,47 @@
...
+
+const CustomBox = styled(Box)`
+  border-radius: 2px;
+`
+
+const defaultStepColor = 'background.light'
+
+function ProgressBar({ steps, currentStep, stepHeight }) {
+  return (
+    <Flex>
```
pcln-james, line 14: can you add a className prop to this flex component please this'll allow us to use styled components on the progress bar should we need to

pcln-james (approved):

common/changes/pcln-design-system/UXPT-1163-progress-bar_2021-02-22-19-42.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "pcln-design-system",
+      "comment": "Create new progress bar component",
+      "type": "patch"
```
pcln-james, line 6: this should be a minor change

craigpalermo (approved):

PR #1028, opened 2021-02-23
feat(RC-17925): hide popover overlay

michael-alterio:
Added new optional prop hideOverlay to Popover component - so that when opened, popover has the option to not be closed by clicking on overlay (e.g. only by clicking on X (close) button inside popover content.

### DIFF
--- a/packages/popover/src/Popover/Popover.js
+++ b/packages/popover/src/Popover/Popover.js
@@ -113,6 +113,8 @@ Popover.propTypes = {
   children: PropTypes.node,
   onOpen: PropTypes.func,
   onClose: PropTypes.func,
+  hideArrow: PropTypes.bool,
+  hideOverlay: PropTypes.bool,
 }
 
 Popover.defaultProps = {
--- a/packages/popover/CHANGELOG.json
+++ b/packages/popover/CHANGELOG.json
@@ -2,6 +2,18 @@
   "name": "pcln-popover",
   "entries": [
     {
+      "version": "4.2.0",
+      "tag": "pcln-popover_v4.2.0",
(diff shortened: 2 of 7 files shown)


packages/popover/src/Popover/Popover.js
```diff
@@ -113,6 +113,8 @@ Popover.propTypes = {
   children: PropTypes.node,
   onOpen: PropTypes.func,
   onClose: PropTypes.func,
+  hideArrow: PropTypes.bool,
```
pcln-james, line 116: Thanks for adding this propType I must've forgotten to 😄

packages/popover/src/Popover/__snapshots__/Popover.spec.js.snap
```diff
@@ -100,6 +100,38 @@ exports[`Popover Trigger Element renders with FocusLock 1`] = `
 </DocumentFragment>
 `;
 
+exports[`Popover Trigger Element renders with overlay 1`] = `
```
pcln-james, line 103: Weird that this snapshot got added since you didn't add a snapshot test 🤔 

The CI also points out that this snapshot is obsolete
image

Could you run a test update please

sdalonzo (changes requested):
run `rush change` and commit the result please.

packages/popover/src/Popover/__snapshots__/Popover.spec.js.snap
michael-alterio: snapshot was removed after running test:update

sdalonzo (approved):
Approved pending resolution of obsolete snapshot test failure.

pcln-james (approved):

PR #1047, opened 2021-03-22
Fix link button variations

pcln-james:


pcln-barry (approved):

sdalonzo (approved):
n-b would be nice to have a test in Link that proves the size is applied to prevent regression

craigpalermo (approved):

PR #1054, opened 2021-03-25
Upgrade to Styled Components 5

craigpalermo:


pcln-james (approved):

gbellini911 (approved):

achoboter (approved):

packages/icons/src/index.spec.js
```diff
@@ -116,11 +116,11 @@ describe('pcln-icons', () => {
     test.each(iconList)(
       '%s Icon should render with no outline ',
       (key, Component) => {
```
sdalonzo, line 118: if you are using query, the test needs to be async

sdalonzo (changes requested):

packages/icons/src/index.spec.js
craigpalermo: Replaced w/ `getAllByTitle`

PR #1067, opened 2021-04-21
Created DotLoader component

pcln-james:
New DotLoader component for displaying something is loading, typically used for displaying a button is loading.

* Export Spinner component
* Export DotLoader component

https://user-images.githubusercontent.com/62619213/115631025-a7f6e080-a2ca-11eb-9d7d-191a98c1ad1f.mov

- [x] Add Tests


packages/core/src/DotLoader/DotLoader.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    const container = screen.getByTestId('dot-loader')
+    const children = container.children
+
+    expect(children).toHaveLength(3)
+    expect(children[0]).toHaveAttribute('color', 'primary')
+    expect(children[0]).toHaveStyleRule('animation', 'bqqVik 2s ease infinite')
+    expect(children[0]).toHaveStyleRule('animation-delay', '0s')
+
+    expect(children[1]).toHaveAttribute('color', 'primary')
+    expect(children[1]).toHaveStyleRule('animation', 'bqqVik 2s ease infinite')
```
craigpalermo, line 18: I think we should consider making this a bit more permissive. I'm concerned that the hash generated for the animation might change over time and cause false negatives.

craigpalermo (commented):
LGTM overall. I just have the comment about the tests.

packages/core/src/DotLoader/DotLoader.spec.js
pcln-james: I was unable to fix this since the animation is on the same css line... leaving it as it is unless somebody has a solution

sdalonzo (approved):

packages/core/src/DotLoader/DotLoader.spec.js
sdalonzo, line 18:
```suggestion
    expect(children[1]).toHaveStyleRule('animation', expect.stringContaining('2s ease infinite'))
```

this _might_ work

packages/core/src/DotLoader/DotLoader.js
```diff
@@ -0,0 +1,108 @@
...
+  large: css`
+    height: 24px;
+
+    & > * {
+      width: 16px;
+      height: 16px;
+    }
+
+    & > :not(:last-child) {
+      margin-right: 32px;
```
bmansell, line 42: A good chunk of these hardcoded pixel values could be replaced with `themeGet()`s 👀

pcln-james: not anymore they're not! good catch 👍

bmansell (approved):

PR #1077, opened 2021-06-07
List component

pcln-james:
A simple list component that supports some of our common styling props 😄


LiamPimlott (approved):

packages/core/src/List/List.spec.js
```diff
@@ -0,0 +1,40 @@
...
+      <List.ol color='text.light' data-testid='list'>
+        <li>Example Text 1</li>
+        <li>Example Text 2</li>
+        <li>Example Text 3</li>
+      </List.ol>
+    )
+
+    const list = screen.getByTestId('list')
+
+    expect(list).toHaveStyleRule('color', '#4f6f8f')
```
craigpalermo, line 17: Can you get the expected value from theme?

craigpalermo (approved):
LGTM, just one comment about the tests.

pcln-barry (approved):

packages/core/src/List/List.spec.js
pcln-james: Do you mean having `text.light` here instead of this hex value? I tried that and it didn't work unfortunately 😞

PR #1087, opened 2021-11-04
Popover - onOpen and onClose pass event object

craigpalermo:


sdalonzo (approved):

edisonalulema (approved):

packages/popover/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pcln-popover",
-  "version": "4.2.13",
+  "version": "4.2.14-beta.0",
```
pcln-james, line 3: do we need to revert this version?

craigpalermo: Nope, rush handles it when we publish

pcln-james (approved):

PR #1107, opened 2022-01-13
Add support for borderRadius and boxShadow to Image

craigpalermo:
image (2)

### DIFF
--- a/packages/core/src/theme/theme.js
+++ b/packages/core/src/theme/theme.js
@@ -176,6 +176,21 @@ export const typography = {
 export const radii = [0, 2, 6]
 export const radius = '2px'
 
+// export const borderRadii = {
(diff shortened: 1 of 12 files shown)


packages/core/src/theme/theme.js
```diff
@@ -176,6 +176,21 @@ export const typography = {
...
+//   lg: '12px',
+//   xl: '16px',
+//   '2xl': '24px',
+//   '3xl': '32px',
+//   full: '9999px',
+//   'action-sm': '9999px',
+//   'action-md': '9999px',
+//   'action-lg': '12px',
+// }
+
```
pcln-james, line 193: should we just delete these?

sdalonzo (approved):

PR #1120, opened 2022-01-28
Add coverage

sdalonzo:


packages/core/src/Text/Text.stories.tsx
```diff
@@ -36,6 +37,21 @@ ResponsiveTypography.args = {
   textStyle: ['heading6', 'heading5', 'heading4', 'heading3', 'heading2', 'heading1'],
 }
 
+export const TypographyScale = () => (
```
craigpalermo, line 40: 💯

rigs/component/profiles/react/config/jest.config.json
```diff
@@ -24,5 +24,8 @@
       "lines": 100,
       "statements": 100
     }
-  }
+  },
+
+  "silent": true,
+  "verbose": true
```
craigpalermo, line 30: Having both of these set to true seems funny 🤷

craigpalermo (approved):

wesleymartin85 (approved):

rigs/component/profiles/react/config/jest.config.json
sdalonzo: i know it does, they aren't opposites though

PR #1126, opened 2022-02-15
add Share icon

nmalayev:
Adding a new Share icon:
image


craigpalermo (approved):

packages/icons/svg/share.svg
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+  <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24' width='24px' height='24px'>
+    <title>Share Link</title>
+    <path
+      fill='#0068ef'
```
pcln-james, line 5: this should be `currentcolor` not the primary blue color right?

nmalayev: I believe the component wrapper passes that in and it overrides the svg's fill. It's confirmed working with the DS color props - I tested with creating some sample stories.

pcln-james (approved):

PR #1139, opened 2022-03-15
Update borderRadii.xl to 6px for backward compatibility

craigpalermo:
This will enable consumers of components like `Card` to use `borderRadius='xl'` to achieve the same 6px border-radius they've already been using.


sdalonzo (changes requested):
this would be a breaking change and could compromise Site-X's tests that are ongoing. we should talk about this - they are all 2px for a reason

PR #1142, opened 2022-03-21
feat(cu): rename buttonColorProps to buttonStyles and some additional props

abautist:
* renaming a prop so opted for a major version change
* buttonHeight, buttonWidth, sideButtonMargin props to allow for custom rules from consumer
* isTwoColumnLayout prop avoids responsive issue where carousel cards will be scrunched up if adjacent to a sibling element

### DIFF
--- a/packages/carousel/src/ArrowButton/ArrowButton.js
+++ b/packages/carousel/src/ArrowButton/ArrowButton.js
@@ -27,6 +27,8 @@ ArrowButton.defaultProps = {
   buttonColor: 'primary.base',
   buttonHoverBackground: 'background.base',
   buttonHoverColor: 'primary.dark',
+  buttonHeight: '60px',
+  buttonWidth: '60px',
 }
 
 ArrowButton.propTypes = {
@@ -41,6 +43,8 @@ ArrowButton.propTypes = {
   buttonHoverBackground: PropTypes.string,
   /** color.shade pattern to be passed to getPaletteColor */
   buttonHoverColor: PropTypes.string,
+  buttonHeight: PropTypes.string,
+  buttonWidth: PropTypes.string,
 }
 
 export { ArrowButton }
--- a/packages/carousel/src/Carousel.js
+++ b/packages/carousel/src/Carousel.js
@@ -43,8 +43,12 @@ export const Carousel = ({
   visibleSlides = 1,
   arrowsPosition = 'side',
   slideSpacing = 2,
+<<<<<<< HEAD
   buttonColorProps,
   onSlideChange,
+=======
+  buttonStyles,
+>>>>>>> 79352593 (feat(cu): add button width, height props, change buttonColorProps name to buttonStyles)
   isTwoColumnLayout = false,
 }) => {
   const widths = layoutToFlexWidths(layout, children.length)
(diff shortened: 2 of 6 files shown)


packages/carousel/src/Carousel.js
```diff
@@ -119,15 +115,15 @@ export const Carousel = ({
...
             />
             {showDots && <Dots />}
             <ArrowButton
               ml={3}
               type='next'
               position='bottom'
               setPosition={arrowsPosition}
-              {...buttonColorProps}
+              {...buttonStyles}
             />
```
craigpalermo, line 127: We'd probably want to clone the node that's passed in as a prop to define props like next, position, etc. in case they're not set by the consumer.

```
            {props.arrowButton ? React.cloneElement(props.arrowButton, {
              type: 'next',
              position: 'bottom',
              setPosition: arrowsPosition
            }) || <ArrowButton
              ml={3}
              type='next'
              position='bottom'
              setPosition={arrowsPosition}
            />
```

packages/carousel/src/Carousel.js
```diff
@@ -43,13 +43,15 @@ export const Carousel = ({
   visibleSlides = 1,
   arrowsPosition = 'side',
   slideSpacing = 2,
-  buttonColorProps,
   onSlideChange,
+  sideButtonMargin = '-30px',
```
ryan-hunt-priceline, line 47: If we made `buttonSize` a `number` type, we could set sideButtonMargin = `'${buttonSize / 2}px'`

packages/carousel/src/Carousel.stories.js
```diff
@@ -179,10 +184,38 @@ AlternateButtonColors.args = {
...
+  sideButtonMargin: '-23px',
+  buttonStyles: {
     buttonBackground: 'primary.base',
     buttonColor: 'text.lightest',
     buttonHoverBackground: 'primary.dark',
     buttonHoverColor: 'text.lightest',
+    buttonWidth: '48px',
+    buttonHeight: '48px',
   },
+  isTwoColumnLayout: true,
```
ryan-hunt-priceline, line 220: I don't think that this prop is needed, you can override the `visibleSlides` prop, which should allow you to do this

packages/carousel/src/Carousel.js
abautist: I made this suggested change for `position: side` buttons only since those are the ones i'm focused on and incorporating the change for `position: top/bottom` would add more dynamic padding/margin code changes.

packages/carousel/src/ArrowButton/ArrowButton.js
```diff
@@ -27,6 +27,8 @@ ArrowButton.defaultProps = {
   buttonColor: 'primary.base',
   buttonHoverBackground: 'background.base',
   buttonHoverColor: 'primary.dark',
+  buttonHeight: '60px',
```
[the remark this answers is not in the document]
abautist: I removed `buttonHeight` and `buttonWidth` and replaced it with `buttonSize` but kept the type as string and `sideButtonMargin` flexible because in the end making the change would mean i would have to pass buttonSize to the Carousel component and the passed in ArrowButton component `<Carousel buttonSize={48} sidePositionArrowButton={<ArrowButton buttonSize={48} />}` and wanted to avoid the duplication.

packages/carousel/src/Carousel.stories.js
abautist: I removed `isTwoColumnLayout` prop. Your suggested change worked for my use case and I ended up using `visibleSlides={[2.1, 1, 1]}`

packages/carousel/src/ArrowButton/ArrowButton.js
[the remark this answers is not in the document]
ryan-hunt-priceline: Sounds good

ryan-hunt-priceline (approved):

craigpalermo (approved):

PR #1144, opened 2022-03-28
Pass zIndex prop to Layout children wrappers

craigpalermo:
This is useful if you need children to overlap.

image


sdalonzo (approved):

packages/core/src/Layout/Layout.tsx
```diff
@@ -139,6 +142,10 @@ const propTypes = {
   ]),
 }
 
+const ZIndexBox = styled(Box)`
+  ${zIndex}
+`
```
pcln-james, line 147: Would it make sense to add this to `Box` instead?

pcln-james (approved):
lgtm once the console.log is gone

PR #1149, opened 2022-04-08
Fixes #1148 | Added mandatory name prop to FilterChip component.

### LINKED ISSUE #1148: FilterChip with "name" prop support
**Is your feature request related to a problem? Please describe.**
`FilterChip` does not support the `name="FilterChip"` attribute in its `<input>` tag for form control.

**Describe the solution you'd like**
Add support of the `name` prop to `FilterChip`, similar to the same feature in `ChoiceChip`.

**Describe alternatives you've considered**
Host apps using the `FilterChip` would have to override their code to provide the same support.

**Additional context**
I have discussed this issue and the potential PR with @sachin-gadagi  / @craigpalermo / @sdalonzo.

sachin-gadagi:
Fixes #[1148](https://github.com/priceline/design-system/issues/1148) | Added mandatory name prop to FilterChip component.

### DIFF
--- a/packages/core/src/Chip/FilterChip/FilterChip.tsx
+++ b/packages/core/src/Chip/FilterChip/FilterChip.tsx
@@ -12,6 +12,7 @@ const propTypes = {
   ...space.propTypes,
   ...fontSize.propTypes,
   id: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
   disabled: PropTypes.bool,
   selected: PropTypes.bool,
   facet: PropTypes.string,
@@ -26,6 +27,7 @@ const propTypes = {
 
 const FilterChip: React.FC<InferProps<typeof propTypes>> = ({
   id,
+  name,
   disabled,
   selected,
   children,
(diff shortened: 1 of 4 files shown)


packages/core/src/Chip/FilterChip/FilterChip.tsx
```diff
@@ -12,6 +12,7 @@ const propTypes = {
   ...space.propTypes,
   ...fontSize.propTypes,
   id: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
```
craigpalermo, line 15: I'm not sure if we should require the new prop. Even though it doesn't break functionality, it's technically a breaking change because the change to the API isn't purely additive if we add a required prop. What do you think?

craigpalermo (commented):
Overall, LGTM! Just one question about the new prop and you'll need to run

- [ ] rush change

packages/core/src/Chip/FilterChip/FilterChip.tsx
sachin-gadagi: Per the discussion with @craigpalermo @sdalonzo and @BeniCheni, the final agreement is to mark the property as resolved.

sdalonzo (approved):

craigpalermo (approved):

BeniCheni (approved):

PR #1151, opened 2022-04-14
feat(UXPT-2737): add overrides for borderRadius and boxShadowSize

bradras:
Update Button component:
- Adds overrides for borderRadius
- Adds overrides for boxShadowSize


bradras:
Updated the coverage but not sure how to show changes.

image

packages/core/src/Button/Button.stories.args.ts
```diff
@@ -1,7 +1,8 @@
-import { colors } from '../storybook/args'
+import { colors, shadows } from '../storybook/args'
 
 export const variations = ['fill', 'link', 'outline', 'plain', 'subtle', 'lightFill']
 export const sizes = ['small', 'medium', 'large']
+export const borderRadiusButtonValues = ['', 'none', 'sm', 'md', 'lg', 'NOT VALID']
```
craigpalermo, line 5: Can you export these values from Button.tsx instead of duplicating them here?

packages/core/src/Button/Button.tsx
```diff
@@ -144,17 +159,23 @@ const propTypes = {
   variation: PropTypes.oneOf(Object.keys(variations)),
   color: deprecatedColorValue(),
   disabled: PropTypes.bool,
+  ...boxShadow.propTypes,
+  borderRadius: PropTypes.oneOf(['', 'none', 'sm', 'md', 'lg']),
```
craigpalermo, line 163: Let's make sure these values are only defined once in this file. Also, You should be able to omit the `''` value since `borderRadius` isn't a required prop.

packages/popover/src/Popover/__snapshots__/Popover.spec.js.snap
```diff
@@ -12,10 +12,21 @@ exports[`Popover Trigger Element Renders trigger element and appends action prop
...
+  sm: 4px;
+  md: 8px;
+  lg: 12px;
+  xl: 16px;
+  2xl: 24px;
+  3xl: 32px;
+  full: 9999px;
+  action-sm: 4px;
+  action-md: 8px;
+  action-lg: 12px;
```
craigpalermo, line 26: We need to figure out where these CSS properties are coming from

sdalonzo (approved):

craigpalermo (approved):

PR #1154, opened 2022-04-22
Carousel: add stretchSlideHeight prop option

ryan-hunt-priceline:


packages/carousel/src/StretchSlide/StretchSlide.js
```diff
@@ -0,0 +1,23 @@
+import React, { useEffect, useRef } from 'react'
+import PropTypes from 'prop-types'
+
+const StretchSlide = ({ slideContents, height, setHeight }) => {
+  const ref = useRef(null)
+
+  useEffect(() => {
+    ref.current.clientHeight > height && setHeight(ref.current.clientHeight)
+  })
```
craigpalermo, line 9: I think you should add `height` to the dependency array for this `useEffect`

ryan-hunt-priceline: Added `height` and `setHeight` to the dependency array

packages/carousel/src/Carousel.js
```diff
@@ -8,6 +8,7 @@ import { Dots } from './Dots'
 import { ArrowButton } from './ArrowButton'
 import { Slider } from './Slider'
 import { getSlideKey, getVisibleSlidesArray, useResponsiveVisibleSlides } from './helpers'
+import { StretchSlide } from './StretchSlide/StretchSlide'
```
craigpalermo, line 11: n-b mind adding an `index` file so you can omit `/StretchSlide`?

craigpalermo (approved):

packages/carousel/src/Carousel.js
ryan-hunt-priceline: Updated

sdalonzo (approved):

PR #1155, opened 2022-05-02
Feat/uxpt 3027 input sizes

harpreetmand:
[UXPT-3027](https://priceline.atlassian.net/browse/UXPT-3027)
[Invision](https://priceline.invisionapp.com/spec/cl2f58fpw05eo01z4bf85hv5f/select)

added size to input and select
updated storybook controls for
FormField
Input
Select
InputGroup


craigpalermo (approved):

packages/core/src/Select/Select.tsx
```diff
@@ -16,8 +30,7 @@ const propTypes = {
   ...fontSize.propTypes,
   color: deprecatedColorValue(),
 }
-
-const SelectBase: React.FC<InferProps<typeof propTypes>> = styled.select`
+const SelectBase = styled.select.attrs(borderRadiusAttrs)`
```
craigpalermo, line 33: I think you accidentally reverted the typing for `SelectBase`

craigpalermo (changes requested):

packages/core/src/Input/Input.stories.tsx
```diff
@@ -21,27 +21,27 @@ const Template = (args) => <Input my={3} {...args} />
...
-    <Input mb={3} id='input-colors-3' color='secondary' placeholder='Secondary' />
-    <Input mb={3} id='input-colors-4' color='warning' placeholder='Warning' />
-    <Input mb={3} id='input-colors-5' color='alert' placeholder='Alert' />
-    <Input mb={3} id='input-colors-6' color='caution' placeholder='Caution' />
+    <Input mb={3} id='input-colors-1' placeholder='No color' {...args} />
+    <Input mb={3} id='input-colors-2' color='primary' placeholder='Primary' {...args} />
+    <Input mb={3} id='input-colors-3' color='secondary' placeholder='Secondary' {...args} />
+    <Input mb={3} id='input-colors-4' color='warning' placeholder='Warning' {...args} />
+    <Input mb={3} id='input-colors-5' color='alert' placeholder='Alert' {...args} />
+    <Input mb={3} id='input-colors-6' color='caution' placeholder='Caution' {...args} />
```
craigpalermo, line 31: I know it's not entirely in the scope of your change, but I think we should stick to using args only with stories that are using a template. Otherwise, this might set an example for others to use args in regular stories, which isn't the [recommended pattern](https://storybook.js.org/docs/react/writing-stories/args).

harpreetmand: good to know. can you check if its good now @craigpalermo

sdalonzo (approved):

craigpalermo (approved):

PR #1162, opened 2022-05-31
Add white variation to Button

craigpalermo:
This adds variation `white` to Button.


packages/core/src/Button/Button.tsx
```diff
@@ -110,6 +110,18 @@ const variations = {
       box-shadow: 0 0 0 2px ${getPaletteColor('dark')};
     }
   `,
+  white: css`
```
pcln-james, line 113: should we call this `lightestFill` instead so it's naming is more dark mode compatible?

PR #1164, opened 2022-06-03
add onClick and onKeyDown handlers to Carousel children

nmalayev:
Add 2 event handler props for `Carousel`: `onSlideClick` and `onSlideKeyDown`. These are passed to the `Carousel` component and will be called for each carousel item.

Add a `stopPropagation` prop to `Popover` allowing an override when Popover causes issues with swallowing events that need to bubble up higher, like in a Carousel for example.

### DIFF
--- a/packages/carousel/src/Carousel.stories.js
+++ b/packages/carousel/src/Carousel.stories.js
@@ -50,6 +50,12 @@ export default {
     onSlideChange: {
       action: 'Current slide changed',
     },
+    onSlideClick: {
+      action: 'Current slide click',
+    },
+    onSlideKeyDown: {
+      action: 'Current slide key',
+    },
   },
 }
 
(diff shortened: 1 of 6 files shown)


packages/carousel/src/Carousel.stories.js
```diff
@@ -81,7 +81,15 @@ function renderCards() {
   ))
 }
 
-const BasicTemplate = (args) => <Carousel {...args}>{renderCards()}</Carousel>
+const BasicTemplate = (args) => (
+  <Carousel
+    {...args}
+    onSlideClick={(props) => console.log('Slide click!', props)}
+    onSlideKeyDown={(props) => console.log('Keyboard click', props)}
```
craigpalermo, line 88: Mind turning these into actions for the stories? https://storybook.js.org/docs/react/essentials/actions#action-argtype-annotation

craigpalermo (changes requested):

packages/carousel/src/Carousel.stories.js
nmalayev: Oops! Totally meant to do that, thanks for pointing it out!

craigpalermo (approved):

PR #1167, opened 2022-07-05
feat(indeterminate-checkbox): add support for indeterminate checkboxes

byrekt:
This adds support for the very legitimate indeterminate checkbox state.

`indeterminate` can only be set via javascript (you can't do something like `<input type='checkbox' indeterminate />`. It just won't work. 

Here is an example below (from the storybook story):
indeterminateCheckboxes

I made it so that it works by passing `indeterminate` as a prop, but it also has support for you to set `indeterminate` via the ref and the styling should apply correctly as well.

### DIFF
--- a/packages/core/src/Checkbox/Checkbox.tsx
+++ b/packages/core/src/Checkbox/Checkbox.tsx
@@ -1,11 +1,12 @@
-import React from 'react'
+import React, { useRef, useEffect } from 'react'
 import styled, { withTheme } from 'styled-components'
-import { BoxChecked, BoxEmpty } from 'pcln-icons'
+import { BoxChecked, BoxEmpty, BoxMinus } from 'pcln-icons'
 import PropTypes, { InferProps } from 'prop-types'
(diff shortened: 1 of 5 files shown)


pcln-james (approved):
Fantastic feature! Don't forget to include a `rush change`

packages/core/src/Checkbox/Checkbox.tsx
```diff
@@ -76,21 +98,40 @@ const StyledInput = styled.input`
...
       theme={props.theme}
       color={props.color}
       disabled={disabled}
     >
-      <StyledInput type='checkbox' {...props} role='checkbox' ref={ref} />
+      <StyledInput
+        type='checkbox'
+        {...props}
+        role='checkbox'
+        aria-checked={props.indeterminate ? 'mixed' : props.checked}
```
trenaryja, line 122: great accessibility addition here!

byrekt: You know I love helping blind people rent cars

PR #1171, opened 2022-07-18
feat(unselected-checkbox-color): add prop to select checkbox color

anjorinio:
### Type of change

- [x] New Feature (non-breaking change which adds functionality)

### Description/Summary of changes

Add a property to allow for color change for unselected checkboxes.

#### PR Creator Checklist

- [x] I wrote a description of requested changes
- [x] I performed a self-review of my own code (and followed code review checklist)

#### Code review checklist

- [x] Code is self-documenting and easy to understand
- [x] There is no commented out code
- [x] There is no unnecessary logging or debugging code
- [x] Naming of methods, variables, and classes are proper

### DIFF
--- a/packages/core/src/Checkbox/Checkbox.tsx
+++ b/packages/core/src/Checkbox/Checkbox.tsx
@@ -62,7 +62,7 @@ const CheckBoxWrapper = styled.div`
   > input {
     & ~ svg[data-name='empty'] {
       color: ${(props) =>
-        props.disabled ? getPaletteColor('border.base') : getPaletteColor(props.unselectedColor)};
+        props.disabled ? getPaletteColor('border.base')(props) : getPaletteColor(props.unselectedColor)};
     }
     &:focus ~ svg {
       border: 1px solid ${getPaletteColor('border.base')};
--- a/packages/core/src/Checkbox/Checkbox.tsx
+++ b/packages/core/src/Checkbox/Checkbox.tsx
@@ -62,7 +62,7 @@ const CheckBoxWrapper = styled.div`
   > input {
     & ~ svg[data-name='empty'] {
       color: ${(props) =>
-        props.disabled ? getPaletteColor('border.base')(props) : getPaletteColor(props.unselectedColor)};
+        props.disabled ? getPaletteColor('border.base') : getPaletteColor(props.unselectedColor)};
     }
     &:focus ~ svg {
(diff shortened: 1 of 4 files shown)


packages/core/src/Checkbox/Checkbox.tsx
```diff
@@ -59,6 +60,10 @@ const CheckBoxWrapper = styled.div`
   }
 
   > input {
+    & ~ svg[data-name='empty'] {
+      color: ${(props) =>
+        props.disabled ? getPaletteColor('border.base')(props) : getPaletteColor(props.unselectedColor)};
```
pcln-james, line 65: below we set the svg with that attribute to display: none so I don't think this'll work.
```
& ~ svg[data-name='empty'] {
  display: none;
}
```

Additionally, I'd prefer if we just did `color: ${getPaletteColor('base')}` outside of any selectors

anjorinio:
> @craigpalermo should the default color be for `unselectedColor='text.base'` set to `'text.light'` so that nothing in production is effected by this change? cc @anjorinio

@wesleymartin85, good point. I've updated the default to 'text.light'

packages/core/src/Checkbox/Checkbox.tsx
anjorinio: > 

@pcln-james , the two places with `display: none` are for `> input:indeterminate` and `> input:checked`, I think it will work(it works fine on local). 

When I try out `${getPaletteColor('base')}`, the default color for disabled unchecked checkbox is box, and on hover, it's grey. I am not remotely familiar with the codebase but my guess would be that it could be a breaking change? Not sure (plus I see everywhere in the code is using `getPaletteColor('border.base')(props)` when handling props.disabled

wesleymartin85:
@anjorinio not sure if it's just the example in this story but the colors for the unchecked by default should be text.light and the disabled state should be border.base. Can you update those and make sure that these changes wont effect anything in production.

anjorinio:
> Details

anjorinio:
> @anjorinio not sure if it's just the example in this story but the colors for the unchecked by default should be text.light and the disabled state should be border.base. Can you update those and make sure that these changes wont effect anything in production. https://5a85e56213417800200978ab-kuzccgqqxi.chromatic.com/?path=/story/checkbox--checkbox-states

@wesleymartin85, fixed. Thanks

PR #1173, opened 2022-07-26
Add font size and font weight to link

pcln-james:
This is required for some work I am doing on creating header links. I could use styled components, however this is likely something others want too and therefore makes sense to me for it to be a base feature.


pcln-james:
Closing as we have a way of handling this already using the `Text` component. I will have a follow up PR next week adding examples of this to the storybook

PR #1199, opened 2022-09-01
feat() Add prop loadMoreCallback which will add load more slide for m…

ryan-hunt-priceline:
…obile


craigpalermo:
No longer needed because the existing callback to load more items was fixed.

PR #1201, opened 2022-09-15
Adding a new size for the buttons 'xl' or 'extraLarge'

Bsoong:
Next-Landing has recently adapted to having inputs of height 56px, which is larger than any of the variants that we have here. After talking with @wesleymartin85, he thought that adding a xL button would be a good way to get onboarded into the design-system but also benefit my team! 

Screen shots: 
Screen Shot 2022-09-15 at 3 07 39 PM

### DIFF
--- a/common/changes/pcln-design-system/xlButtons_2022-09-15-18-37.json
+++ b/common/changes/pcln-design-system/xlButtons_2022-09-15-18-37.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "pcln-design-system",
+      "comment": "Added an extra large button with a height of 56px",
+      "type": "none"
+    }
+  ],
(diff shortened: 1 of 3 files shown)


common/changes/pcln-design-system/xlButtons_2022-09-15-18-37.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "pcln-design-system",
+      "comment": "Added an extra large button with a height of 56px",
+      "type": "none"
```
craigpalermo, line 6: This should probably be a minor since you're adding functionality

craigpalermo (commented):
LGTM once you fix the change type

craigpalermo (approved):

sdalonzo (approved):

PR #1202, opened 2022-09-21
feature(RC-21569): adds new sort icon and updated MenuItem styles

michael-alterio:
- Updated background-color and color for MenuItem default, selected and hover states

**MenuItem**

**Default Color**
image

**Custom Color**
image

**With Icon**
image

### DIFF
--- a/packages/menu/src/MenuItem/MenuItem.js
+++ b/packages/menu/src/MenuItem/MenuItem.js
@@ -6,28 +6,37 @@ import { borderRadius } from 'styled-system'
 import { Check as CheckIcon } from 'pcln-icons'
 
 const MenuButton = styled(Button).attrs(borderRadiusAttrs)`
+  align-items: center;
+  background-color: ${(props) =>
+    props.selected ? getPaletteColor('light')(props) : getPaletteColor('background.lightest')(props)};
+  border: 2px solid transparent;
+  color: ${getPaletteColor('text.base')};
   display: flex;
   justify-content: space-between;
-  align-items: center;
-  border: 2px solid transparent;
-  color: ${getPaletteColor('text.dark')};
   outline: none;
   padding: 12px;
 
   &:hover {
-    background-color: ${getPaletteColor('base')};
+    ${(props) =>
+      props.selected
+        ? `
+        background-color: ${getPaletteColor('light')(props)};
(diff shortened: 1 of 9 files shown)


pcln-james (changes requested):
removing the ability to customize colors for MenuItem would be a breaking change. Is there a way that we can have your new changes will keeping what's there today?

wesleymartin85:
@michael-alterio before we merge this lets meet to update the styles

packages/menu/src/MenuItem/MenuItem.js
```diff
@@ -6,28 +6,37 @@ import { borderRadius } from 'styled-system'
...
   &:hover {
-    background-color: ${getPaletteColor('base')};
+    ${(props) =>
+      props.selected
+        ? `
+        background-color: ${getPaletteColor('light')(props)};
+        color: ${getPaletteColor('dark')(props)};`
+        : `
+        background-color: ${getPaletteColor('background.base')(props)};
+        color: ${getPaletteColor('text.base')(props)};
```
pcln-brianna, line 27: I think we still need to allow for custom colours to be set on the button's hover state to be backwards compatible. Also since this is changing from the default `primary.base` on hover it might still be worth making it a breaking/major change.

Screen Shot 2022-09-26 at 4 46 03 PM
Screen Shot 2022-09-26 at 4 46 22 PM

michael-alterio: Updated to major change! Let me discuss with Wes and other members of design-system team on how we want to handle the custom color hovering.

michael-alterio: @wesleymartin85 @sdalonzo @pcln-james How should we best support custom hover color while also supporting these new changes? (FYI: I made this a major change)

pcln-brianna: Discussed offline and we're good with only selected colour customization going forward (based on existing implementations getting the in-path update) + also setting it back down to a minor, thanks!

It might be good to split this up into two PRs (move icon changes into a new PR, keep this for the menu update) so that your icon update isn't blocked by testing this.

pcln-brianna (approved):

pcln-james (approved):

PR #1204, opened 2022-09-27
Fix serialization bug in Layout

craigpalermo:
Upgrading to the latest `moize` to get this fix: https://github.com/planttheidea/moize/pull/177

This PR also refactors `Layout` to no longer use `moize`.


scottned (approved):
You updated the moize package, but it also looks like you completely removed it from Layout.  Do you still need it as a dependency?

PR #1207, opened 2022-10-06
Feat/ppn 58101 create swatch component

devin-efendy-pcln:
https://priceline.atlassian.net/browse/PPN-58101

Create a new component called `Swatch` that renders a group of colors.

Accepts an array of colors and and optional onClick callback that return selected color.

### Select color
https://user-images.githubusercontent.com/90356592/194409604-4cb9834f-d6f6-41e5-a317-54b4b1bcb033.mov

### Wrap
Screen Shot 2022-10-06 at 3 12 15 PM

### Docsite
Screen Shot 2022-10-06 at 3 12 37 PM


packages/core/src/Swatch/Swatch.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Flex } from '../Flex'
+
+const SwatchColor = styled.div`
+  height: 24px;
+  width: 24px;
+  background-color: ${(props) => props.color};
+  border-radius: 50%;
+  display: flex;
+  margin: 4px;
+`
```
pcln-james, line 13: could you add `cursor: pointer;` to indicate that it's clickable

packages/core/src/Swatch/Swatch.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const propTypes = {
+  colors: PropTypes.arrayOf(PropTypes.string),
+  onClick: PropTypes.func,
+}
+
+const Swatch: React.FC<InferProps<typeof propTypes>> = ({ colors, onClick }) => (
+  <Flex flexWrap='wrap'>
```
pcln-james, line 21: could spread props on this please so that we can adjust margins, padding, etc. `{...props}`

packages/core/src/Swatch/Swatch.tsx
devin-efendy-pcln: Thanks, that's a good catch! I changed it if onClick is provided then we show a pointer cursor otherwise we show the regular cursor.

packages/core/src/Swatch/Swatch.tsx
devin-efendy-pcln: Added!

packages/core/src/Swatch/Swatch.tsx
pcln-james: `mouse` isn't a valid value, you want `default`

devin-efendy-pcln: Right.. 😅 changed to default

pcln-james (approved):

craigpalermo (approved):

PR #1208, opened 2022-10-10
Create SlideBox component and use it for mobile Carousel 

ryan-hunt-priceline:
Created a `SlideBox` component. This new component will replace the mobile experience for the `Carousel`. We are able to use `layout`, `visibleSlides`, `stretchHeight`, etc components.
This should not contain any breaking changes to the existing Carousel users.
Notes:
- If user provides `layout` prop to carousel, will default to **1.2**, and **2.2** visible slides for xs, and sm browser widths
- Same as above if user specifies an `int` for `visibleSlides` prop
- If user provides an array for `visibleSlides`, no defaults will be used for mobile browser sizes


packages/carousel/src/Slider/Slider.js
```diff
@@ -1,11 +1,6 @@
 import styled from 'styled-components'
 import { Slider } from 'pure-react-carousel'
 
-const StyledSlider = styled(Slider)`
-  overflow: scroll !important;
-  ${(props) => props.theme.mediaQueries.md} {
-    overflow: hidden !important;
-  }
-`
+const StyledSlider = styled(Slider)``
```
pcln-james, line 4: does this still need to be a styled component?

packages/carousel/src/constants.js
```diff
@@ -1,17 +1,17 @@
-const XS_VISIBLE_SLIDES = 1.1
+const VISIBLE_SLIDES_BREAKPOINT_1 = 1.1
 
-const SM_VISIBLE_SLIDES = 2.1
+const VISIBLE_SLIDES_BREAKPOINT_2 = 2.1
```
pcln-james, line 3: what's the rational behind these names vs the previous ones? matching the named breakpoints makes more sense than numbers imo

packages/carousel/src/helpers.js
```diff
@@ -50,7 +52,17 @@ const useResponsiveVisibleSlides = (visibleSlides) => {
...
+  }
 }
 
-export { getSlideKey, getVisibleSlidesArray, useResponsiveVisibleSlides }
+//This is to keep consistant with previous version.
+//We should make a major version release that will allow more breakpoints
+const getMobileVisibleSlidesArray = (visibleSlides) => [visibleSlides[0], null, visibleSlides[1]]
+
+const getMobileVisibleSlides = (visibleSlides) =>
+  Array.isArray(visibleSlides) ? getMobileVisibleSlidesArray(visibleSlides) : visibleSlides
```
pcln-james, line 66: should there be tests for this code?

packages/carousel/src/constants.js
ryan-hunt-priceline: Renamed because they don't actually map to the regular DS breakpoints.

packages/core/src/SlideBox/styles.ts
```diff
@@ -0,0 +1,16 @@
...
+
+import styled from 'styled-components'
+import { Flex, Box } from '..'
+
+const SlideWrapper = styled(Box)``
+
+const ScrollFlex = styled(Flex)`
+  position: relative;
+  overflow: auto;
+  > ${SlideWrapper} {
```
pcln-james, line 11: should this have an `&`?

packages/carousel/src/constants.js
pcln-james: that's fair lol is one the smallest and the other the biggest breakpoint that they render for?

ryan-hunt-priceline: Yeah, the carousel only has 2 breakpoints for customizing currently.

pcln-james: would `VISIBLE_SLIDES_BREAKPOINT_SMALLEST` and `VISIBLE_SLIDES_BREAKPOINT_LARGEST` fit the situation better? I'm not 100% on what these variables are used for

packages/core/src/SlideBox/styles.ts
ryan-hunt-priceline: Just tested - seems to work both ways. I'll change to `&>`

packages/carousel/src/constants.js
ryan-hunt-priceline: Possibly, if we end up adding another one though, i think it would be hard to name `VISIBLE_SLOIDES_BREAKPOINT_MEDIUMIST`

PR #1209, opened 2022-10-13
Create ScaledText component

trenaryja:
This PR adds the `ScaledText` component, which renders SVG Text elements that auto-scale to fit the width of their parent container. Here's an example:

image

Each line is independent, allowing you to style it according to your needs. Here are some of the properties you can set per line:
- fontWeight (bold on "Hello")
- fill (deepskyblue on "World")
- fontStyle (italic on "Today Is")
- fontFamily (courier on "Thursday")
- rotate (10 degrees on each emoji)

There is also a color prop to set the fill-color for all lines, unless an individual line overrides it (in this case it is set to midnightblue, except on the "World" line, which is overridden to be deepskyblue)


packages/core/src/ScaledText/ScaledText.tsx
```diff
@@ -0,0 +1,41 @@
...
+  lines: {
+    text: string
+    key?: React.Key
+    props?: React.SVGProps<SVGTextElement>
+  }[]
+  color?: string
+}
+
+const ScaledText = ({ id, lines, color }: IScaledTextProps) => {
+  const _id = id ?? uuidv4()
```
craigpalermo, line 15: The ID should probably be set on state, otherwise, it seems like this could cause unnecessary rerenders.

trenaryja: Thanks for the 👀! Added id to State to prevent any unnecessary rerenders!

jligi (approved):

PR #1214, opened 2022-10-20
Add props to allow more flexible control for pcln-menu

devin-efendy-pcln:
Added the following props to `Menu` component:
- `placement`: control placement of the popover

**custom `placement='bottom'`**
Screen Shot 2022-10-20 at 5 18 02 PM

Use `max-height` for `MenuList` component so `MenuItem` is not stretched out when the length of the content does not exceed the given height.

**Before**
Screen Shot 2022-10-20 at 4 41 25 PM
**After**
Screen Shot 2022-10-20 at 4 41 35 PM


packages/menu/src/Menu/Menu.js
```diff
@@ -16,18 +16,22 @@ function Menu({
...
   trapFocus,
+  placement,
   children,
   ...props
 }) {
   const MenuContent = ({ handleClose }) => (
-    <MenuList id={id} color={color} size={size} height={height} handleClose={handleClose}>
-      {children}
-    </MenuList>
+    <Box {...contentProps}>
```
pcln-james, line 30: can we not pass `contentProps` directly to the MenuList?

devin-efendy-pcln: I don't pass it to `MenuList` since it has it's own styling https://github.com/priceline/design-system/blob/f53e03e2a7247666a1b9d1ca303c62cc3df779da/packages/menu/src/MenuList/MenuList.js#L71-L83

And overriding those styling would mess with the placement of the children. Especially when adjusting margin and padding.

devin-efendy-pcln: Discussed this offline. We don't need adjustment at all so removing `contentProps`

packages/menu/src/Menu/Menu.stories.args.js
```diff
@@ -86,4 +86,33 @@ export const argTypes = {
...
+        'top-end',
+        'bottom',
+        'bottom-start',
+        'bottom-end',
+        'right',
+        'right-end',
+        'right-end',
+        'left',
+        'left-end',
+        'left-end',
```
pcln-brianna, line 114: Looks like some repeated values are in here:
```suggestion
        'right-end',
        'left',
        'left-end',
```

devin-efendy-pcln: right, thanks for catching that!

packages/menu/src/Menu/Menu.js
```diff
@@ -60,7 +61,7 @@ function Menu({
       trapFocus={trapFocus}
       width={width}
       zIndex={1600}
-      placement='bottom-start'
+      placement={placement ?? 'bottom-start'}
```
pcln-james, line 64: this didn't change the code coverage requirements? either way, could you add a test for this please.

devin-efendy-pcln: Added a snapshot tests since the placement is important for the popover content.

pcln-james: I think this is still testing that the prop passed to the Menu controls the placement behaviour, doesn't have to be a big test but something there to make sure that we have a control and a fallback.

devin-efendy-pcln: ah right the fallback 😅 added the test using data-placement!

pcln-james (approved):
LGTM

pcln-brianna (approved):

PR #1215, opened 2022-11-04
Update GHA workflows to run on Node 16

craigpalermo:
Upgrading our workflows to run on Node 16 since 14 is deprecated.

### DIFF
--- a/.github/workflows/npm-publish-beta.yaml
+++ b/.github/workflows/npm-publish-beta.yaml
@@ -25,6 +25,11 @@ jobs:
           fetch-depth: 0
           token: ${{ secrets.GH_PAT }}
 
+      - name: Use Node.js 16
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16
+
       - name: PNPM cache via actions/cache@v2
         id: pnpm-cache
         uses: actions/cache@v2
--- a/.github/workflows/on-main.yml
+++ b/.github/workflows/on-main.yml
@@ -13,10 +13,10 @@ jobs:
         with:
           fetch-depth: 0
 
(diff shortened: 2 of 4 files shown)


.github/workflows/on-main.yml
```diff
@@ -13,10 +13,10 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js 16
         uses: actions/setup-node@v1
```
huzefakagdi, line 17: Is anything blocking to update to `actions/setup-nodev3` instead? It uses `Node 16` by default as per the [docs](https://github.com/actions/setup-node/releases/tag/v3.0.0).

craigpalermo: Oh nice, I wasn't aware of that! I'll give it a try.

craigpalermo: Updating to v3 of that action, but I'm going to keep the `node-version` input in place based on the docs:

> However, it is recommended to always specify Node.js version and don't rely on the system one.

.github/workflows/npm-publish-beta.yaml
```diff
@@ -25,6 +25,11 @@ jobs:
           fetch-depth: 0
           token: ${{ secrets.GH_PAT }}
 
+      - name: Use Node.js 16
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
```
sdalonzo, line 31: n-b `16` to match the matrix in ci workflow?

sdalonzo (approved):

PR #1216, opened 2022-11-07
Update link and textColorOn disabled states

pcln-james:
* `textColorOn` updated to include overrides for the light and dark colors, falling back to `text.lightest` and `text.base` as it was before. This allows us to set the fallback dark color to `primary` for example.
  * Created `getValidPaletteColor` and `hasPaletteColorShade` as support functions
* `Link` disabled state added to support cards that can be disabled.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -29,6 +29,7 @@
     "debug": "heft --debug test --clean",
     "start": "heft start",
     "storybook": "heft start --storybook",
+    "test": "heft test",
     "test:snapshots": "heft test --update-snapshots",
     "test:watch": "heft test --watch"
   },
--- a/packages/core/src/Link/Link.tsx
+++ b/packages/core/src/Link/Link.tsx
@@ -35,6 +35,7 @@ const variations = {
 
 const propTypes = {
   color: deprecatedColorValue(),
+  disabled: PropTypes.bool,
   variation: PropTypes.oneOf(Object.keys(variations)),
 }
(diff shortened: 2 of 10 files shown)


amyamccaskill (approved):

packages/core/package.json
```diff
@@ -29,6 +29,7 @@
     "debug": "heft --debug test --clean",
     "start": "heft start",
     "storybook": "heft start --storybook",
+    "test": "heft test",
```
pcln-brianna, line 32: 👍

packages/core/src/Link/Link.tsx
```diff
@@ -43,23 +44,38 @@ export interface ILinkProps
     SpaceProps,
     React.HTMLAttributes<HTMLAnchorElement>,
     React.RefAttributes<unknown> {
+  children?: React.ReactNode | string
   color?: string
+  disabled?: boolean
```
pcln-brianna, line 49: I'm wondering if this should be closer to a conventional boolean name like `isDisabled`? If this is already consistent with how other components name it in `core` though that's fine.

pcln-james: I was considering that too but my reasoning was 1. like you said all our components already use `disabled` and 2. that's the attribute name in plain html for things that are disableable.

pcln-brianna (approved):

craigpalermo (approved):

PR #1221, opened 2022-11-10
Add currentSlideOverride to SlideBox component

ryan-hunt-priceline:


common/changes/pcln-design-system/slidebox-setvisibleSlide_2022-11-10-23-10.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "pcln-design-system",
+      "comment": "Add currentSlideOverride option to SlideBo",
```
craigpalermo, line 5:
```suggestion
      "comment": "Add currentSlideOverride option to SlideBox",
```

packages/core/src/SlideBox/Slide.tsx
```diff
@@ -0,0 +1,54 @@
...
+const propTypes = {
+  onSlideChange: PropTypes.func,
+  slideSpacing: PropTypes.number,
+  stretchHeight: PropTypes.bool,
+  index: PropTypes.number,
+  content: PropTypes.node,
+  isCurrentSlide: PropTypes.bool,
+  width: PropTypes.any,
+  numSlides: PropTypes.number,
+}
```
craigpalermo, line 16: Would you mind updating this to use both the propTypes object and an interface? Here's an example: https://github.com/priceline/design-system/blob/main/packages/core/src/Flex/Flex.tsx#L22

ryan-hunt-priceline: No problem. just made the update

packages/core/src/SlideBox/Slide.tsx
```diff
@@ -0,0 +1,65 @@
...
+  numSlides,
+}) => {
+  const ref = useRef()
+
+  useEffect(() => {
+    if (isCurrentSlide === true) {
+      /* istanbul ignore next */
+      ref?.current?.scrollIntoView({ block: 'end' })
+    }
+  }, [isCurrentSlide])
```
sdalonzo, line 35: shouldn't `ref` be a dependency to the effect?

sdalonzo (approved):

packages/core/src/SlideBox/Slide.tsx
ryan-hunt-priceline: Updated

PR #1226, opened 2022-11-22
Button input variation

pcln-james:
Created a new variation for the `Button` component called `input` which makes the Button render in a similar fashion to an input. This is useful for situation like my screenshot when we want a dropdown, the clickable element to open the dropdown is a Button but it looks like an input.

image

### DIFF
--- a/packages/core/src/Button/Button.tsx
+++ b/packages/core/src/Button/Button.tsx
@@ -1,12 +1,22 @@
 import React from 'react'
 import styled, { css } from 'styled-components'
 import PropTypes from 'prop-types'
-import { width, space, boxShadow, WidthProps, SpaceProps, BoxShadowProps } from 'styled-system'
+import {
+  borderRadius,
+  fontSize,
+  width,
+  space,
+  boxShadow,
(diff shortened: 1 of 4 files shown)


packages/core/src/Button/Button.tsx
```diff
@@ -143,6 +153,24 @@ const variations = {
...
+    background-color: transparent;
+    border-style: solid;
+    border-width: 1px;
+    color: ${getPaletteColor('text.base')};
+    display: block;
+    font-family: inherit;
+    font-weight: normal;
+    line-height: normal;
+    margin: 0;
+    padding: 14px 12px;
```
pcln-brianna, line 167: Should we allow for this padding to be configurable?

pcln-james: `${space}` a few lines below means it still allows it, this is just a default value for it 😄

pcln-brianna: Missed that 👍 thanks

pcln-brianna (approved):

PR #1227, opened 2022-11-22
Add animate component

trenaryja:
Add Animate component to wrap content in an animating div. 

Thinly wraps the Framer-Motion library, and exposes some pre-determined animations for use within the PCLN design system.


craigpalermo (changes requested):
The code change LGTM, but be sure to confirm that the Interactions tab is configured in the Storybook app as well.

jligi (approved):

ryan-hunt-priceline (approved):

packages/core/.storybook/main.js
```diff
@@ -2,8 +2,11 @@ const { heftStoryGlobs, storykitAddons, features, framework, core } = require('@
 
 module.exports = {
   stories: heftStoryGlobs,
-  addons: storykitAddons,
-  features,
+  addons: [...storykitAddons, '@storybook/addon-interactions'],
+  features: {
+    ...features,
+    interactionsDebugger: true,
+  },
```
pcln-james, line 9: do we want to commit this change?

pcln-james (approved):
Great work! The storybook examples are great.

packages/core/.storybook/main.js
trenaryja: I believe so! This configurations change allows us to perform interaction testing through storybook, which is particularly useful for animation testing!

craigpalermo (approved):

PR #1230, opened 2022-11-29
Add Popout component

trenaryja:
This PR introduces the `Popout` component. This is very similar to the concept of a `Popover` component, but uses absolute positioning to "pop-out" a "modal" around/on-top of the trigger element when it is clicked or focused. The triggering element can be any JSX element, not just an input! 

This gif shows an example of the behavior:
ezgif-2-812d6536aa


jligi (approved):

packages/core/src/Popout/Popout.tsx
```diff
@@ -0,0 +1,128 @@
...
+  }, [onClose])
+
+  useLayoutEffect(() => {
+    setHeight(baseInputRef.current.clientHeight)
+  }, [trigger])
+
+  useEffect(() => {
+    const escapeListener = (e: KeyboardEvent) => {
+      if (isOpen && e.key === 'Escape') handleClose()
+    }
```
craigpalermo, line 82: I noticed that tabbing out of the focused input has some weird behavior. Dismissing via the Escape key works as expected.

https://user-images.githubusercontent.com/3260642/205312036-63d854a2-b75b-4b50-95a6-bff3114035eb.mp4

packages/core/src/Popout/Popout.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const content = value && <h1>{value}</h1>
+
+  return (
+    <Wrapper>
+      <Popout trigger={trigger} triggerRef={ref} content={content} />
+    </Wrapper>
+  )
+}
+
+export const IconPopout = () => {
```
craigpalermo, line 88: Would it make sense for clicking the trigger again when the popout is open to close the popout?

https://user-images.githubusercontent.com/3260642/205312872-1b4e73f6-606e-4a81-a212-131c8a0a843f.mp4

craigpalermo (changes requested):
Code change LGTM overall, but I left a couple of comments about functionality we should probably discuss.

packages/core/src/Popout/Popout.stories.tsx
trenaryja: Great idea, added a prop called `closeOnTriggerRefClick` for this behavior! Makes it optional, so it can be applied for cases like the Icon Example, but left off for input fields which you'd want to be able to click without closing the Popout. Thanks for the feedback!

packages/core/src/Popout/Popout.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const focusListener = (e: FocusEvent) => {
+      if (
+        isOpen &&
+        e.target instanceof HTMLElement &&
+        e.target !== baseInputRef.current &&
+        !modalRef.current.contains(e.target)
+      )
+        handleClose()
```
pcln-james, line 96: could you add braces to all the if statements please to be consistent with how DS currently does it. it's more lines of code but should make it a bit more readable

packages/core/src/Popout/Popout.tsx
```diff
@@ -0,0 +1,134 @@
+import React, { RefObject, useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react'
+import styled from 'styled-components'
+import { Box } from '..'
+
+export interface IPopoutProps {
+  trigger: JSX.Element
+  content?: JSX.Element | undefined
```
pcln-james, line 7: I normally use React.ReactElement as my type but I just randomly found that, is JSX.Element better?

packages/core/src/Popout/Popout.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const handleOpen = useCallback(() => {
+    setIsOpen(true)
+    setTimeout(() => {
+      setPadding(16)
+      setOpacity(0.5)
+      /* istanbul ignore next */
+      triggerRef?.current?.focus()
+      /* istanbul ignore next */
+      onOpen?.()
+    })
```
pcln-james, line 63: why is there a `setTimeout` here? this implies there is an issue with state

ryan-hunt-priceline: You might want to try putting it on a `useEffect` with `triggerRef` and `isOpen` in the dependency array and try removing the `setTimeout` and see if that still works. You can then add an if statement for `isOpen`.

packages/core/src/Popout/Popout.tsx
```diff
@@ -0,0 +1,134 @@
...
+  trigger: JSX.Element
+  content?: JSX.Element | undefined
+  triggerRef?: RefObject<HTMLElement>
+  onOpen?: () => void
+  onClose?: () => void
+  closeOnTriggerRefClick?: boolean
+}
+
+const TriggerWrapper = styled(Box)`
+  position: relative;
```
ryan-hunt-priceline, line 15: NB: we have a `Relative` component that should work here. You might even be able to set `width='fit-content'` as a prop. although I haven't tried that before.

packages/core/src/Popout/Popout.tsx
```diff
@@ -0,0 +1,134 @@
...
+  width: fit-content;
+`
+
+const OnClickHandler = styled(Box)`
```
ryan-hunt-priceline, line 19: NB: can also use `<Absolute top={0} width='100%' height='100%>`

packages/core/src/Popout/Popout.tsx
```diff
@@ -0,0 +1,134 @@
...
+  width: 100vw;
+  height: 100vh;
+  z-index: 1;
+  transition: opacity 250ms ease-in-out;
+`
+
+const PopoutModal = styled(Box)`
+  position: absolute;
+  z-index: 2;
+  background-color: white;
```
ryan-hunt-priceline, line 41: Can you update to use `bg='background.lightest'` ?

packages/core/src/Popout/Popout.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  return (
+    <Box>
+      <TriggerWrapper tabIndex={isOpen ? -1 : 0} onFocus={handleOpen} ref={baseInputRef}>
+        {isOpen ? placeHolderDiv : trigger}
+        <OnClickHandler onClick={handleOpen} />
+      </TriggerWrapper>
+      {isOpen && (
+        <>
+          <Shadow bg='background.darkest' style={{ opacity }} onClick={handleClose} />
```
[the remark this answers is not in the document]
trenaryja: I love adding consistency throughout the Design-System! I used the `ShadowOverlay` that is exported alongside the `ShadowEffect` component, since its use case is exactly the same as what I needed here. Works like a charm, thanks for pointing it out!

packages/core/src/Popout/Popout.tsx
trenaryja: I use `JSX.Element` for just about the same reason you use `React.ReactElement`, I just randomly found it lol. From some skimming on StackOverflow, it looks like they are nearly equivalent, but using `JSX.Element` allows other JSX-ish libraries (like Preact, Solid, etc.) to still use the prop and understand what to do with it, even if they handle JSX their own way. 

Here's the post that I read through: [https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement](link)

packages/core/src/Popout/Popout.tsx
trenaryja: I am a bit embarassed that I didn't realize this should have been a useEffect the whole time 🤦‍♂️ it seems so obvious now. Implemented the suggestions @ryan-hunt-priceline recommended and it works perfectly, and is much more readable I feel. Thank you both for being so thorough and helpful with your feedback!

packages/core/src/Popout/Popout.tsx
trenaryja: Added!

packages/core/src/Popout/Popout.tsx
trenaryja: Fixed this, great catch!

ryan-hunt-priceline (approved):

packages/core/src/Popout/Popout.tsx
```diff
@@ -0,0 +1,135 @@
...
+      <TriggerWrapper tabIndex={isOpen ? -1 : 0} onFocus={handleOpen} ref={baseInputRef}>
+        {isOpen ? placeHolderDiv : trigger}
+        <OnClickHandler onClick={handleOpen} />
+      </TriggerWrapper>
+      {isOpen && (
+        <>
+          <ShadowOverlay
+            zIndex={1}
+            onClick={handleClose}
+            style={{ opacity, transition: 'opacity 250ms ease-in-out' }}
```
pcln-james, line 115: nb: this would be cool to add to ShadowOverlay

pcln-james (approved):

craigpalermo (approved):

PR #1233, opened 2022-12-14
Add control and story to demo Stamp variations

trenaryja:
This PR adds a control for the `variation` prop on the `Stamp` component, as well as a story to demo all of the available variations

### DIFF
--- a/packages/core/src/Stamp/Stamp.stories.args.ts
+++ b/packages/core/src/Stamp/Stamp.stories.args.ts
@@ -10,4 +10,9 @@ export const argTypes = {
       type: 'select',
     },
   },
+  variation: {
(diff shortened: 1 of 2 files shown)


packages/core/src/Stamp/Stamp.stories.args.ts
```diff
@@ -10,4 +10,9 @@ export const argTypes = {
       type: 'select',
     },
   },
+  variation: {
+    type: { name: 'string' },
+    options: ['fill', 'outline', 'solid'],
```
craigpalermo, line 15: Good thing I noticed this in the migration guide re `solid`! 
CleanShot 2022-12-14 at 13 41 29@2x

PR #1246, opened 2023-01-16
fix(v4): Make bold optional in getLinkStylesOn

pcln-brianna:
- Add fourth param to allow for making bold optional if a different colour than the default is chosen for the link styles
- Update docs page and add toggle on the Link storybook
- Util is not yet in use so patching this in
- Add new test and update existing

#### Screenshots
Screen Shot 2023-01-16 at 9 55 08 AM
Screen Shot 2023-01-16 at 9 55 13 AM


pcln-james (approved):
~needs a change file but~ nvm lgtm

PR #1248, opened 2023-01-19
Fix bug in Checkbox when using indeterminate prop

craigpalermo:
The existing implementation of the `indeterminate` prop on `Checkbox` broke ref forwarding, which broke support for libraries like `react-hook-form`. This implementation is based on Material UI's implementation (see [here](https://github.com/mui/material-ui/blob/master/packages/mui-material/src/Checkbox/Checkbox.js#L182)) and uses React state rather than HTML attributes.


anjorinio (approved):
Thanks @craigpalermo, this fix works on NL. Just need to confirm it works for @byrekt as well and we are good.

PR #1249, opened 2023-01-23
Fix bug with indeterminate Checkbox

craigpalermo:
This is the v4 implementation of https://github.com/priceline/design-system/pull/1248


anjorinio (approved):
LGTM!

PR #1252, opened 2023-01-26
Storybook redesign

trenaryja:
ATTN: This PR is on hold until [this PR](https://github.com/priceline/design-system/pull/1273) is merged in, due to a cyclic dependency order

The PR is the first redesign the documentation for components in the Design System. While this PR focuses on documenting the `Button` component, it sets up patterns that can be followed to quickly and easily document other components as well. 

Features of the new documentation/story format:
- Reusable Documentation Components coming from new `@priceline/docs-utils` package:
- Full Typescript Safety using types directly from storybook
- Concept of a `Playground` story
  - this is the simplest implementation of the component, with storybook controls enabled
  - when all components have this `Playground` story, it can be used as what gets rendered in the Related Components section
- Live Editable Stories
  - Every story can be rendered in a playground environment, allowing modifications, using the `LiveDemo` component 

image


packages/core/src/Button/Button.tsx
```diff
@@ -224,7 +224,7 @@ export interface IButtonProps
     React.RefAttributes<unknown> {
   color?: string
   variation?: 'fill' | 'link' | 'outline' | 'plain' | 'subtle' | 'white' | 'lightFill' | 'input'
-  size?: Sizes | Sizes[]
+  size?: Sizes
```
pcln-james, line 227: are we dropping support for `Sizes[]` this is for when we want it to be responsive e.g., `size={['small', null, 'medium']}`

packages/core/src/storybook/components/RelatedComponents.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const RelatedComponentContainer = ({ children }: RelatedComponentContainerProps) => (
+  <Box
+    style={{
+      display: 'grid',
+      gridAutoFlow: 'column',
+      gridTemplateColumns: `repeat(${Children.count(children)}, 1fr)`,
+      gap: theme.space[3],
+    }}
+  >
```
pcln-james, line 45: Would the `Container` component work for this?

packages/core/src/Button/Button.tsx
trenaryja: Ah great catch, this is a change I made when messing with the controls for the default story, and missed adding it back/undoing it! Thanks for finding this, I just made a commit to revert it 😄

packages/core/src/storybook/components/RelatedComponents.tsx
trenaryja: I looked at `Container`, and it seems to be a Box with a `size` prop to control width to some preset values. I don't think it accomplishes the same desired "grid-like" outcome,  

I have a [ticket](https://priceline.atlassian.net/jira/software/c/projects/UXPT/boards/4263?modal=detail&selectedIssue=UXPT-4286) on the JIRA board to introduce a `<Grid></Grid>` component to the design system, which I think would be perfect, and I intended to replace this implementations with a `Grid` as soon as it's available. 

Here's some [examples](https://chakra-ui.com/docs/components/grid/usage) of what we could do with the grid component! The CSS spec for grid is amazing, and since we don't have to worry as much about IE anymore we could lean in fully and make all sorts of [neat layouts](https://gridbyexample.com/examples/)

PR #1258, opened 2023-02-10
Layout spread props to allow overrides

pcln-james:
I need the ability to add margin to the top of a single row Layout, I could wrap with a Box component but this doesn't seem like a spec breaking thing and seems like it'd be valuable for others.


craigpalermo (approved):

trenaryja (approved):
Would love to see this pattern adopted in a lot of other components as well! If something can be "Box-Based" then it's props could be unioned with the Box props, and they could be spread on the Box/Container! Thanks for making the api for this component a little more useful 👍

robnorman-pcln (approved):

PR #1265, opened 2023-02-23
feat(UXPT-3871): Add variation prop to badge to support semantic colors

DerrickTheodore:
This change introduces a new prop `BadgeStyle` to the core `<Badge/>` component.   Consumers can now provide a semantic value to the `BadgeStyle` prop to style the `<Badge/>`.

### DIFF
--- a/packages/core/src/Badge/Badge.spec.tsx
+++ b/packages/core/src/Badge/Badge.spec.tsx
@@ -81,4 +81,68 @@ describe('Badge', () => {
     expect(json).toHaveStyleRule('background-color', theme.colors.lightBlue)
     expect(json).toHaveStyleRule('color', theme.colors.text)
   })
+
+  test.each([
+    { badgeStyle: 'primary', backgroundColor: theme.colors.blue, color: theme.colors.lightestText },
+    { badgeStyle: 'primaryLight', backgroundColor: theme.colors.lightBlue, color: theme.colors.blue },
+    { badgeStyle: 'primaryDark', backgroundColor: theme.colors.darkBlue, color: theme.colors.lightestText },
+    { badgeStyle: 'primaryDarkLight', backgroundColor: theme.colors.lightBlue, color: theme.colors.darkBlue },
+    { badgeStyle: 'secondary', backgroundColor: theme.colors.green, color: theme.colors.lightestText },
+    { badgeStyle: 'secondaryLight', backgroundColor: theme.colors.lightGreen, color: theme.colors.green },
+    {
+      badgeStyle: 'backgroundDarkest',
+      backgroundColor: theme.colors.darkestBackground,
+      color: theme.colors.lightestText,
+    },
+    {
+      badgeStyle: 'backgroundLightest',
+      backgroundColor: theme.colors.lightestBackground,
+      color: theme.colors.text,
+    },
+    { badgeStyle: 'neutral', backgroundColor: theme.colors.darkBackground, color: theme.colors.lightestText },
+    {
+      badgeStyle: 'neutralLight',
+      backgroundColor: theme.colors.background,
+      color: theme.colors.lightText,
+    },
+    {
+      badgeStyle: 'warning',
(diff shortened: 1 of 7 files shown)


packages/core/src/Badge/Badge.tsx
```diff
@@ -57,11 +57,73 @@ const sizes = {
...
+    highlight: {
+      backgroundColor: props.theme.palette.highlight.light,
+      color: props.theme.palette.highlight.shade,
+    },
+    notify: {
+      backgroundColor: props.theme.palette.notify.light,
+      color: props.theme.palette.text.base,
+    },
+  }
+  return !props.bg && !props.color && (badgeStyles[props.badgeStyle] || {})
```
DerrickTheodore, line 119: I wanted to start a discussion based on this opinion of mine.  Here I'd like to handle the case of when a consumer assigns a value to `bg`, `color` and `badgeStyle`, by making `badgeStyle` a no-op.  Ideally, I think it'll be thoughtful to raise a warning in their browser like we do with [deprecatedColorValue](https://github.com/priceline/design-system/pull/1265/files#diff-deaba2548643a017885d124c11f77e7d967dfa8cfdac98c6a8c6710e27b9d1e7L83)

packages/core/src/Badge/Badge.spec.tsx
```diff
@@ -81,4 +81,102 @@ describe('Badge', () => {
     expect(json).toHaveStyleRule('background-color', theme.colors.lightBlue)
     expect(json).toHaveStyleRule('color', theme.colors.text)
   })
+
+  test('badgeStyle primary sets background-color and color', () => {
+    const json = rendererCreateWithTheme(<Badge badgeStyle='primary' />).toJSON()
+    expect(json).toMatchSnapshot()
+    expect(json).toHaveStyleRule('background-color', theme.colors.blue)
+    expect(json).toHaveStyleRule('color', theme.colors.lightestText)
+  })
```
craigpalermo, line 90: You could combine these test cases since they all have the same setup with different variables

https://dev.to/bgord/simplify-repetitive-jest-test-cases-with-test-each-310m

packages/core/src/Badge/Badge.tsx
```diff
@@ -57,11 +57,73 @@ const sizes = {
   `,
 }
 
+const badgeStyle = (props) => {
+  const badgeStyles = {
+    primary: {
+      backgroundColor: props.theme.palette.primary.base,
```
craigpalermo, line 63: It wouldn't hurt to destructure out `palette` so you don't have to repeat `props.theme.palette`

packages/core/src/Badge/Badge.tsx
craigpalermo: I think that's a good idea, although most people tend to tune out DS warnings. I wonder if throwing an error would be too obtrusive.

packages/core/src/Badge/Badge.tsx
```diff
@@ -57,11 +57,73 @@ const sizes = {
...
+    },
+  }
+  return !props.bg && !props.color && (badgeStyles[props.badgeStyle] || {})
+}
 export interface IBadgeProps extends SpaceProps, React.HtmlHTMLAttributes<HTMLElement> {
   size?: 'small' | 'medium'
   color?: string
   bg?: string
   borderRadius?: string
+  badgeStyle?: string
```
craigpalermo, line 126: Should the type here be one of the allowed `badgeStyle`s instead of `string`?

packages/core/src/Badge/Badge.tsx
trenaryja: Agree with the order of precedence, where badgeStyle is top-level important. I think a warning might be enough, because it will be so visually noticeable

Bsoong (commented):
I saw the thread of comments and agreed with pretty much everything that Justin & Craig said.  I also saw that you updated the PR with their recs, once you fix the merge conflicts I'll approve 😄

PR #1266, opened 2023-02-28
Fix two prop type warnings

pcln-james:
* Update FormField to use an interface instead of prop types
* Pass id prop to Popover idx prop to fix prop type warning

We have a situation where the label for our FormField needs to be outside of the FormField but then a prop warning gets thrown. a11y is satisfied but we still get the warning. I'm removing the prop check as there are edge cases that make this check incorrect.

Here is the component in question:

https://user-images.githubusercontent.com/62619213/221957723-ebc5990d-8347-4376-8aaa-d88c39509ef8.mov

### DIFF
--- a/packages/core/src/Tooltip/Tooltip.stories.tsx
+++ b/packages/core/src/Tooltip/Tooltip.stories.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
-import { Tooltip, InputField, Box, Flex } from '..'
+import { Box, Flex, FormField, Label, Input, Tooltip } from '..'
+import { Check } from 'pcln-icons'
 import { argTypes } from './Tooltip.stories.args'
 import styled from 'styled-components'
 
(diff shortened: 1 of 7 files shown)


packages/core/src/Tooltip/Tooltip.stories.tsx
```diff
@@ -45,94 +46,100 @@ export const WithoutAnchors = () => (
...
-      <InputField
-        icon='circleInfo'
-        color='blue'
-        label='Email Address'
-        defaultValue='albus.dumbledore@priceline.com'
-        id='form-field-3'
-        placeholder='example@test.com'
-      />
-      <Tooltip bottom left bg='blue' color='white'>
+      <FormField>
```
pcln-james, line 61: InputField did not work like this, it's just a regular FormField so I've updated these stories to provide better/more accurate examples of what you want to do.

pcln-james: We don't have an InputField component, it's just a re-export of FormField https://github.com/priceline/design-system/blob/main/packages/core/src/FormField/index.ts#L3

sdalonzo (approved):

PR #1275, opened 2023-03-16
feature(RC-23253): new car transmission icons

michael-alterio:
Adds new Transmission Icons


packages/icons/svg/carAutomatic.svg
```diff
@@ -0,0 +1,7 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
pcln-james, line 1: should this be `fill="currentcolor"`?

pcln-james: sorry never mind for here, but the lines below might want to be that rather than the fix text colour

craigpalermo (approved):

pcln-james (approved):

packages/icons/svg/carAutomatic.svg
pcln-james: discussed in slack and looks like it does work without my suggestion

PR #1282, opened 2023-03-17
Fix responsive styles when using custom breakpoints

craigpalermo:
This fixes a bug when passing `customBreakpoints` to `<ThemeProvider>` and using responsive props arrays. Some changes internally to Styled System 5 seem to have caused this behavior change, but using Styled System's `compose` function fixes it.

### DIFF
--- a/packages/core/src/utils/createTheme.ts
+++ b/packages/core/src/utils/createTheme.ts
@@ -29,6 +29,7 @@ const createTheme = (theme = {}, customBreakpoints = null, existingTheme = defau
   const mediaQueries = customBreakpoints
     ? createMediaQueries(customBreakpoints)
     : finalMergedTheme.mediaQueries
+
   return {
     ...finalMergedTheme,
     contrastRatio: finalMergedTheme.contrastRatio || 2.6,
--- a/packages/modal/src/Modal.js
+++ b/packages/modal/src/Modal.js
@@ -2,7 +2,7 @@ import React from 'react'
 import styled from 'styled-components'
 import PropTypes from 'prop-types'
 import { Transition } from 'react-transition-group'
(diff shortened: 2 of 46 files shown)


pherrera167 (approved):
LGTM

packages/modal/src/Modal.spec.js
```diff
@@ -77,12 +77,12 @@ describe('Modal', () => {
 
     expect(getByText('Content')).toBeTruthy()
     expect(dialogContent).toHaveStyle(`
-      height: 100vh;
-      width: 100vw;
+      height: 100%;
```
sdalonzo, line 80: ?

packages/core/src/utils/createTheme.ts
```diff
@@ -38,6 +39,8 @@ const createTheme = (theme = {}, customBreakpoints = null, existingTheme = defau
     colorStyles: createColorStyles(finalMergedTheme),
     textStyles: createTextStyles(finalMergedTheme),
   }
+  // console.log(t)
```
sdalonzo, line 42: console

packages/modal/src/Modal.js
```diff
@@ -76,6 +73,12 @@ const Dialog = styled(DialogContent)`
     max-height: none;
   `}
   ${getAnimation}
+
+  ${(props) => {
+    const c = compose(color, width, height)(props)
+    console.log(c)
```
sdalonzo, line 79: console

packages/core/src/utils/createTheme.ts
gbellini911: ooooOooOoo it was for your console log.

sdalonzo (approved):

PR #1285, opened 2023-03-21
Storybook redesign 2

craigpalermo:


rush.json
```diff
@@ -332,10 +332,11 @@
 
   "projects": [
     {
-      "packageName": "@priceline/docs-utils",
+      "packageName": "pcln-docs-utils",
       "projectFolder": "packages/docs-utils",
       "reviewCategory": "packages",
-      "shouldPublish": true
+      "shouldPublish": true,
+      "decoupledLocalDependencies": ["@priceline/storybook-preset", "pcln-design-system"]
```
trenaryja, line 339: Do we need the `@priceline/storybook-preset` here? I know I had it in there myself, but I think we only really need `pcln-design-system`

trenaryja (approved):

PR #1299, opened 2023-04-26
clean up of styled-system peers

sdalonzo:


packages/menu/package.json
```diff
@@ -56,6 +56,6 @@
     "react": ">=16.10.0",
     "react-dom": ">=16.10.0",
     "styled-components": ">=4 <5 || >=5 <5.3.4 || >=5.3.7 <6",
-    "styled-system": "^4.2.4"
+    "styled-system": "^4.2.4 || ^5.0.0"
```
craigpalermo, line 59: I wonder if we should restrict it to `^5`. Core has SS5 as a hard dependency, so all of the DS5 packages are only intended to work with SS5.

packages/core/src/Button/Button.tsx
```diff
@@ -351,8 +335,6 @@ const Button = React.forwardRef((props: IButtonProps, ref) => {
   )
 })
 
-Button.propTypes = buttonPropTypes
```
trenaryja, line 354: 👋 Goodbye prop-types, hello typescript!

trenaryja (approved):

packages/menu/package.json
sdalonzo: updated! thanks

PR #1300, opened 2023-04-27
Update Link and Button component with changes to hover state in subtle, plain and white

ralston7:
I'm currently working on this ticket https://priceline.atlassian.net/browse/SITE-20524. I've to use the `Link` with `subtle` type. Which is currently not available. This change will allow us to use it.

Also , there are changes to the `background-color` of hover state in Button. To match the specs in here https://www.figma.com/file/1lLCo0ZnO1RyMDEbnnS0by/Web-Design-System?type=design&node-id=145-21874&t=mqAZdqJ3WFQmYcB2-0 , for subtle, palin and white type

### DIFF
--- a/packages/core/src/Link/Link.stories.tsx
+++ b/packages/core/src/Link/Link.stories.tsx
@@ -27,6 +27,12 @@ export function ForwardRefs() {
           I am a link!
         </Link>
         <br />
+        <Link variation='subtle'>subtle</Link>
+        <br />
+        <Link variation='plain'>plain</Link>
+        <br />
+        <Link variation='white'>white</Link>
+        <br />
         <Button color='error' onClick={onClick} mt={4}>
           Click to update link text via ref
         </Button>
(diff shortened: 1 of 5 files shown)


packages/core/src/Link/Link.stories.tsx
```diff
@@ -27,6 +27,12 @@ export function ForwardRefs() {
           I am a link!
         </Link>
         <br />
+        <Link variation='subtle'>subtle</Link>
+        <br />
+        <Link variation='plain'>plain</Link>
+        <br />
+        <Link variation='white'>white</Link>
+        <br />
```
craigpalermo, line 35: Would you mind moving these to a new story that shows all of the variations?

ralston7: @craigpalermo I just added all the variations of the link, you can view it [here](https://5a85e56213417800200978ab-lxiuwcvals.chromatic.com/?path=/story/link--all-link-examples). Please have a look at this PR when you get a chance.

craigpalermo (approved):

PR #1303, opened 2023-05-10
Add Grid component to pcln-design-system

trenaryja:
This PR introduces a new `Grid` component to the Design System. This is a primitive component, that exposes the [CSS Grid Specification](https://developer.mozilla.org/en-US/docs/Web/CSS/grid) in a react component, allowing for:
  - More complex layouts, especially in 2-dimensional designs (`Flex` is really only suited for 1-dimensional designs)
  - Nesting/Composing of layouts (Grids within Grids, or Gridception)
  - Gaps (independently control `rowGap` or `columnGap`, or both at the same time with the `gap` prop)


packages/core/src/Grid/__snapshots__/Grid.spec.tsx.snap
```diff
@@ -0,0 +1,11 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Grid renders correctly with default props 1`] = `
```
sdalonzo, line 3: this serves no purpose

sdalonzo (changes requested):
just lacking a changefile. this is a great MVP, and we can iterate as use cases present

trenaryja:
> just lacking a changefile. this is a great MVP, and we can iterate as use cases present

@sdalonzo Thanks for the reminder about the changefile! I've added that in now

sdalonzo (approved):

craigpalermo (approved):

PR #1309, opened 2023-06-14
Upgrade Storybook to v7, use @storybook/react-vite, move DocsUtils into core to break cyclic dependency

sdalonzo:
# Major Storybook ecosystem upgrades
- Storybook v7
  - configs in TypeScript
  - adds `storybook-addon-design` for tight Figma integration protected by SSO
- uses Vite as the builder (7 second startup for entire DS monorepo)
  - Vite requires resolving `ambiguous` imports (types must be imported with `import type`) 
  - Vite requires files containing `jsx` to have the appropriate filename (.jsx/.tsx)
- resolves all peer dependency issues in every package except legacy docsite
- migrates `pcln-docs-utils` into `pcln-design-system` core 

image

image

image

image

image


packages/core/src/Container/Container.tsx
```diff
@@ -41,4 +41,4 @@ Container.propTypes = propTypes
 
 Container.displayName = 'Container'
 
-export default Container
+export { Container }
```
pcln-james, line 44: does the index file have to be updated for this?

sdalonzo: it was. Vite's bundler is very specific about ambiguity between imports. all the changes that look like this are for that reason.

pcln-james (approved):

packages/slider/src/global.d.ts
```diff
@@ -4,4 +4,4 @@ declare var JSX
 
 declare module 'testing-library'
 
-declare module 'pcln-icons'
+// declare module 'pcln-icons'
```
craigpalermo, line 7: n-b are you keeping this around for a reason?

rigs/component/profiles/react/config/jest.config.json
```diff
@@ -25,6 +25,8 @@
     }
   },
 
+  "moduleFileExtensions": ["ts", "tsx", "cjs", "js", "jsx", "json", "node"],
```
craigpalermo, line 28: n-b what's a `.node` file extension for?

craigpalermo (approved):

rigs/component/profiles/react/config/jest.config.json
sdalonzo: i'm not sure tbh. in the default, it failed because jsx was not present. when it failed, it listed the current config - so i copied it and added jsx

PR #1317, opened 2023-07-06
Switch to Vite/vitest

sdalonzo:


packages/core/src/Absolute/Absolute.spec.tsx
```diff
@@ -1,29 +1,31 @@
...
-    const consoleError = console.error
-    console.error = jest.fn()
+    expect(container.firstChild?.firstChild).toHaveStyle('position: absolute')
+    expect(container.firstChild?.firstChild).toHaveStyle('top: 10px')
+    expect(container.firstChild?.firstChild).toHaveStyle('left: 0px')
+    expect(container.firstChild?.firstChild).toHaveStyle('z-index: 2')
+  })
 
-    const json = rendererCreateWithTheme(
+  it('renders with Flag and children', () => {
```
sdalonzo, line 18: fails because `cleanup` is not working

PR #1322, opened 2023-07-07
Run prettier on all files tracked by Git

craigpalermo:
* Updated `.prettierignore` to keep it in sync with `.gitignore`
* Ran `prettier . --write`
* Added `eslint-config-prettier` to Heft eslint config


common/config/rush/command-line.json
```diff
@@ -16,6 +16,7 @@
       "name": "prettier",
       "commandKind": "global",
       "summary": "Used by the pre-commit Git hook. This command invokes Prettier to reformat staged changes.",
+      "safeForSimultaneousRushProcesses": true,
```
sdalonzo, line 19: Not sure this is safe

packages/core/package.json
```diff
@@ -115,7 +115,6 @@
     "framer-motion": "^6.5.1",
     "jest-styled-components": "^7.1.1",
     "npm-run-all": "^4.1.5",
-    "prettier": "^2.8.8",
```
sdalonzo, line 118: This is needed afaik for vscode to function correctly

sdalonzo (changes requested):

common/config/rush/command-line.json
craigpalermo: CleanShot 2023-07-10 at 08 06 00

I got it from whatever's current in the Rush docs

packages/core/package.json
craigpalermo: Are you setting this to anything in your VSCode settings? I just realized I was getting results inconsistent with our Prettier config when pointing it to `./common/autoinstallers/rush-prettier/node_modules/prettier`. I'll leave this dependency as is just to be safe.

common/config/rush/command-line.json
sdalonzo: 👍

sdalonzo (approved):

PR #1325, opened 2023-07-10
feat(UXPT-4958): Create ChatTrigger component

pcln-james:
https://priceline.atlassian.net/browse/UXPT-4958

Created Floating Action Button.

Created ChatTrigger that wraps the FAB component but sets it to the bottom right corner of the screen. I did it this way because the Figma designs make it seem like we might add alternative variations for the ChatTrigger component with FAB being just one of them.

### DIFF
--- a/packages/core/src/Absolute/Absolute.tsx
+++ b/packages/core/src/Absolute/Absolute.tsx
@@ -16,13 +16,13 @@ import styledSystemPropTypes from '@styled-system/prop-types'
 
 import { Box, IBoxProps } from '../Box'
 
-export interface IAbsoluteProps extends IBoxProps {
-  top?: TopProps
-  right?: RightProps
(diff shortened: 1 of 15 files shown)


packages/core/src/Absolute/Absolute.tsx
```diff
@@ -1,8 +1,28 @@
...
+
+export interface IAbsoluteProps
+  extends IBoxProps,
+    TopProps,
+    RightProps,
+    BottomProps,
+    LeftProps,
+    ZIndexProps {}
 
 const Absolute = styled(Box)`
```
craigpalermo, line 27: React.FC<IAbsoluteProps>

craigpalermo (approved):

pherrera167 (approved):

PR #1326, opened 2023-07-10
add support for overflow to Box

amyamccaskill:
- Adding support for overflow as a prop to `Box`
- Alphabetized some imports/propTypes

Storybook: 
Screenshot 2023-07-10 at 4 31 44 PM

### DIFF
--- a/packages/core/src/Box/Box.spec.tsx
+++ b/packages/core/src/Box/Box.spec.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { render } from '../__test__/testing-library'
 
 import { theme, Box } from '..'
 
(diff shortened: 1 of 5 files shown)


packages/core/src/Box/Box.stories.tsx
```diff
@@ -152,6 +152,26 @@ export const DirectionalMargin = () => (
   </Box>
 )
 
+export const Overflow = () => (
+  <Box p={3}>
+    <Box overflow='hidden' maxHeight='100px' maxWidth='100px' mb={3} color='black' bg='blue'>
```
pcln-james, line 157: for new code could you use `color='background.darkest'` and `bg='primary'` if that works please

pcln-james (approved):
Awesome work!

craigpalermo (approved):

packages/core/src/Box/Box.spec.tsx
```diff
@@ -49,4 +49,10 @@ describe('Box', () => {
     expect(json).toMatchSnapshot()
     expect(json).toHaveStyleRule('box-shadow', theme.shadows.sm)
   })
+
+  test('overflow prop sets overflow', () => {
```
sdalonzo, line 53: n-b but if possible, create snapshots with `asFragment` from testing-library. we want to eventually remove `rendererCreateWithTheme`

PR #1328, opened 2023-07-11
feat(uxpt-4952): updated the toggle states

sangeethaathiyursridharan:
**Confirmed disabled states for both disabled on and off**
Disabled ON
image

Disabled OFF
image

**Added hover states when toggle is selected and enabled**
image

**Updated the focus to focus-visible for keyed events**
image

**Added boxShadowSize to handle (sm for off/on, xl for off/on hover)** 

xl for off/on hover
image

sm for off/on
image

### DIFF
--- a/packages/core/src/Toggle/Toggle.spec.tsx
+++ b/packages/core/src/Toggle/Toggle.spec.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 import { Toggle } from '.'
 import { createTheme } from '..'
 import { render, fireEvent } from '../__test__/testing-library'
+import { theme as customThemes } from '../theme/theme'
 
 const theme = createTheme()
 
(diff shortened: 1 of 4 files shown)


packages/core/src/Toggle/Toggle.spec.tsx
```diff
@@ -32,4 +32,15 @@ describe('Toggle', () => {
...
   })
+
+  test('add boxshadow when hover over the enabled states', () => {
+    const { getByLabelText, getByRole } = render(<Toggle label='Total price toggle' />)
+    const input = getByLabelText('Total price toggle')
+    fireEvent.mouseOver(input)
+    const container = getByRole('switch').parentNode
+    expect(container).toHaveStyleRule(
+      'box-shadow',
+      'rgba(0, 0, 0, 0.03) 0px -1px 0px 0px, rgba(0, 0, 0, 0.16) 0px 2px 8px 0px, rgba(0, 0, 0, 0.16) 0px 10px 8px -5px, rgba(0, 0, 0, 0.16) 0px 12px 32px -2px'
```
craigpalermo, line 43: Can you get this value from the theme rather than hardcoding it?

https://github.com/priceline/design-system/blob/main/packages/core/src/theme/theme.ts#L298

craigpalermo:
LGTM once you try changing that shadow value in the unit test 👍 Great work!

craigpalermo (approved):

PR #1329, opened 2023-07-11
add Dialog component

sdalonzo:
This PR adds the `Dialog` component to the Priceline Design System. This dialog is based off of the [Radix-UI dialog primitive](https://www.radix-ui.com/docs/primitives/components/dialog), and prescribes some logical defaults to make usability more convenient for the developer. Here is a video demo of the component in action:

https://github.com/priceline/design-system/assets/23641048/255b4ab2-3853-4e27-b946-31915eeaeaf3

Also, this adds the `neutralDarkOnLightest` color scheme to the overall design system

### DIFF
--- a/packages/core/src/Dialog/Dialog.spec.tsx
+++ b/packages/core/src/Dialog/Dialog.spec.tsx
@@ -1,0 +1,28 @@
+import React from 'react'
+import { render } from '../__test__/testing-library'
+
+import { Dialog } from '.'
+
+describe('Dialog', () => {
+  it('should render correctly', () => {
+    const { asFragment } = render(<Dialog>Dialog</Dialog>)
+
+    expect(asFragment()).toMatchInlineSnapshot(`
+      <DocumentFragment>
+        .c0 {
+        font-family: 'Montserrat','Helvetica Neue',Helvetica,Arial,sans-serif;
+        line-height: 1.4;
+        font-weight: 500;
+      }
+
+      .c0 * {
+        box-sizing: border-box;
+      }
+
+      <div
+          class="c0"
+        />
+      </DocumentFragment>
+    `)
+  })
+})
--- a/packages/core/src/Dialog/Dialog.spec.tsx
+++ b/packages/core/src/Dialog/Dialog.spec.tsx
@@ -1,28 +1,71 @@
+import { render, screen } from '@testing-library/react'
 import React from 'react'
-import { render } from '../__test__/testing-library'
+import { Dialog } from '..'
 
-import { Dialog } from '.'
+const triggerText = 'Open Dialog'
+const contentText = 'Dialog Content'
 
 describe('Dialog', () => {
-  it('should render correctly', () => {
-    const { asFragment } = render(<Dialog>Dialog</Dialog>)
+  it('renders the trigger node', () => {
+    render(
+      <Dialog triggerNode={<button>{triggerText}</button>} ariaDescription='Description' ariaTitle='Title'>
+        {contentText}
+      </Dialog>
+    )
 
-    expect(asFragment()).toMatchInlineSnapshot(`
-      <DocumentFragment>
-        .c0 {
-        font-family: 'Montserrat','Helvetica Neue',Helvetica,Arial,sans-serif;
-        line-height: 1.4;
-        font-weight: 500;
-      }
+    const triggerNode = screen.getByText(triggerText)
+    expect(triggerNode).toBeInTheDocument()
+  })
 
-      .c0 * {
-        box-sizing: border-box;
-      }
+  it('renders the dialog content when trigger is clicked', () => {
+    render(
+      <Dialog triggerNode={<button>{triggerText}</button>} ariaDescription='Description' ariaTitle='Title'>
+        {contentText}
+      </Dialog>
+    )
 
-      <div
-          class="c0"
-        />
-      </DocumentFragment>
-    `)
+    const triggerNode = screen.getByText(triggerText)
+    triggerNode.click()
+
+    const contentNode = screen.getByText(contentText)
+    expect(contentNode).toBeInTheDocument()
+  })
+
+  it('calls onOpenChange when trigger is clicked', () => {
+    const onOpenChange = jest.fn()
+    render(
+      <Dialog
+        triggerNode={<button>{triggerText}</button>}
+        ariaDescription='Description'
+        ariaTitle='Title'
+        onOpenChange={onOpenChange}
+      >
+        {contentText}
+      </Dialog>
+    )
+
+    const triggerNode = screen.getByText(triggerText)
+    triggerNode.click()
+
+    expect(onOpenChange).toHaveBeenCalledWith(true)
+  })
+
+  it('handles controlled state when open', () => {
+    render(
+      <Dialog
+        triggerNode={<button>{triggerText}</button>}
+        ariaDescription='Description'
+        ariaTitle='Title'
+        open={false}
+      >
+        {contentText}
+      </Dialog>
+    )
+
+    const triggerNode = screen.getByText(triggerText)
+    triggerNode.click()
+
+    const contentNode = screen.queryByText(contentText)
+    expect(contentNode).not.toBeVisible()
   })
 })
--- a/packages/core/src/Dialog/Dialog.spec.tsx
+++ b/packages/core/src/Dialog/Dialog.spec.tsx
@@ -17,7 +17,7 @@ describe('Dialog', () => {
     expect(triggerNode).toBeInTheDocument()
   })
 
-  it('renders the dialog content when trigger is clicked', () => {
+  it('renders the dialog content when trigger is clicked', async () => {
     render(
       <Dialog triggerNode={<button>{triggerText}</button>} ariaDescription='Description' ariaTitle='Title'>
         {contentText}
@@ -26,6 +26,8 @@ describe('Dialog', () => {
 
     const triggerNode = screen.getByText(triggerText)
     triggerNode.click()
+
+    await new Promise((resolve) => setTimeout(resolve, 1000))
 
     const contentNode = screen.getByText(contentText)
     expect(contentNode).toBeInTheDocument()
--- a/packages/core/src/Dialog/Dialog.stories.args.tsx
+++ b/packages/core/src/Dialog/Dialog.stories.args.tsx
@@ -1,0 +1,50 @@
+import { ArgTypes } from '@storybook/react'
+import React from 'react'
+import type { IDialogProps } from '.'
+import { borderRadii, Button, colorSchemeNames, Grid, paletteColors, Text } from '..'
+import { DialogSizes } from './Dialog.styled'
+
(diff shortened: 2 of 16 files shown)


packages/core/src/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,145 @@
+/* eslint-disable no-script-url */
+import type { Meta, StoryObj } from '@storybook/react'
+import type { IDialogProps, IGridProps } from '..'
+
+import { Discount } from 'pcln-icons'
+import React, { useState } from 'react'
+
+import { Box, Breadcrumbs, Button, Dialog, Grid, Text } from '..'
```
sdalonzo, line 8: not using the barrel file was intentional

packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -0,0 +1,169 @@
...
+  },
+} as const
+
+const FloatingCloseButton: typeof CloseButton = styled(CloseButton)`
+  position: absolute;
+  top: ${(props) => -((props.dialogSize === 'full' ? 0 : 16) + 10)}px;
+  right: ${(props) => -((props.dialogSize === 'full' ? 0 : 16) + 10)}px;
+  margin: 16px;
+  padding: 8px;
+  background-color: white;
```
sdalonzo, line 50: background.lightest or similar, from the palette, please

packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -0,0 +1,169 @@
...
+    width: '100%',
+    height: '100%',
+  },
+} as const
+
+const FloatingCloseButton: typeof CloseButton = styled(CloseButton)`
+  position: absolute;
+  top: ${(props) => -((props.dialogSize === 'full' ? 0 : 16) + 10)}px;
+  right: ${(props) => -((props.dialogSize === 'full' ? 0 : 16) + 10)}px;
+  margin: 16px;
```
sdalonzo, line 48: maybe use `.attrs` for things that are provided by DS props. we should not need to set margin and padding in px in a styled-component

packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -0,0 +1,169 @@
...
+  overflow-y: auto;
+  cursor: ${(props: IDialogProps) => (props.scrimDismiss ? 'pointer' : 'default')};
+  place-items: ${(props: IDialogProps) => (props.sheet ? 'end center' : 'center')};
+  background-color: ${(props: IDialogProps) => `rgba(0, 0, 0, ${props.scrimOpacity})`};
+`
+
+const DialogContentWrapper = styled(motion.div)`
+  position: relative;
+  cursor: initial;
+  background-color: white;
```
sdalonzo, line 70: there's some colors and other things that should be coming from the theme that might need a little cleanup

packages/core/src/Dialog/Dialog.styled.tsx
trenaryja: I tried using the `.attrs()` approach for both padding and margin, but it seems the close button doesn't currently accept those props. I instead changed these to calls to `themeGet` in order to stay within the theme guardrails

rigs/component/profiles/react/config/jest.config.json
```diff
@@ -8,7 +8,6 @@
 
   "collectCoverage": true,
   "collectCoverageFrom": [
-    "src/**/*.{js,ts}",
```
sdalonzo, line 11: ??

packages/core/src/Dialog/Dialog.stories.args.tsx
```diff
@@ -0,0 +1,74 @@
...
+  scrimOpacity: 0.25,
+  sheet: false,
+  showCloseButton: true,
+  size: 'md',
+  triggerNode: <Button>Open Dialog</Button>,
+  headerColorScheme: 'neutralDarkOnLightest',
+  headerShowCloseButton: false,
+  headerIcon: false,
+  fullWidth: false,
+  // headerContent: 'Hello World',
```
sdalonzo, line 73: intentional?

packages/core/src/Dialog/Dialog.stories.tsx
```diff
@@ -0,0 +1,182 @@
...
+export const StickyHeader: DialogStory = {
+  ...Playground,
+  args: {
+    children: <ExampleOverflowingContent overflow='auto' maxHeight={500} />,
+    ...exampleHeaderProps,
+    headerShowCloseButton: false,
+  },
+}
+
+// export const WithShadowOnScroll: DialogStory = {
```
sdalonzo, line 120: intentional? gap in the component or story?

packages/core/src/Dialog/Dialog.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+        {contentText}
+      </Dialog>
+    )
+
+    const triggerNode = screen.getByText(triggerText)
+    triggerNode.click()
+
+    const contentNode = screen.getByText(contentText)
+    expect(contentNode).toBeInTheDocument()
+  })
```
craigpalermo, line 32: These are fine to leave around for now, but I'll probably end up deleting them in favor of some Chromatic tests in my next PR 😅

packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -0,0 +1,183 @@
...
+    background-color: ${themeGet('palette.background.lightest')};
+  }
+`
+
+const DialogOverlayWrapper = styled(motion.div)`
+  position: fixed;
+  inset: 0;
+  display: grid;
+  overflow-y: auto;
+  cursor: ${(props: IDialogProps) => (props.scrimDismiss ? 'pointer' : 'default')};
```
craigpalermo, line 62: I feel like the cursor should always be the default pointer. Using `pointer` might be confusing since it's unclear what the pointer is actually pointing at.

packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -0,0 +1,183 @@
...
+const DialogContentWrapper = styled(motion.div)`
+  position: relative;
+  cursor: initial;
+  background-color: ${themeGet('palette.background.lightest')};
+  width: ${(props: IDialogProps) => (props.fullWidth ? '100%' : sizeStyles[props.size]?.width)};
+  height: ${(props: IDialogProps) =>
+    props.sheet ? `calc(${sizeStyles[props.size]?.height} - 24px)` : sizeStyles[props.size]?.height};
+  margin: ${(props: IDialogProps) => (props.sheet ? '16px 0 0 0' : '32px 16px 64px 16px')};
+  border-radius: ${(props: IDialogProps) =>
+    props.sheet ? '24px 24px 0 0' : sizeStyles[props.size]?.borderRadius};
```
craigpalermo, line 76: It might look verbose, but I think we should be getting the individual border-radius values from `theme`.

packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -0,0 +1,183 @@
...
+`
+
+const DialogContentWrapper = styled(motion.div)`
+  position: relative;
+  cursor: initial;
+  background-color: ${themeGet('palette.background.lightest')};
+  width: ${(props: IDialogProps) => (props.fullWidth ? '100%' : sizeStyles[props.size]?.width)};
+  height: ${(props: IDialogProps) =>
+    props.sheet ? `calc(${sizeStyles[props.size]?.height} - 24px)` : sizeStyles[props.size]?.height};
+  margin: ${(props: IDialogProps) => (props.sheet ? '16px 0 0 0' : '32px 16px 64px 16px')};
```
craigpalermo, line 74: Can you get the margins from the `space` scale?

packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -0,0 +1,183 @@
...
+      >
+        <VisuallyHidden asChild>
+          <Dialog.Title>{ariaTitle}</Dialog.Title>
+        </VisuallyHidden>
+
+        <DialogInnerContentWrapper sheet={sheet} hugColor={hugColor} size={size}>
+          {headerContent && (
+            <Grid
+              colorScheme={headerColorScheme}
+              background='#FFF'
```
craigpalermo, line 159: Could this be a palette color instead?

packages/core/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,87 @@
+/* eslint-disable jsx-a11y/aria-props */
```
craigpalermo, line 1: Is this disabled because Radix handles the `aria` props?

packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -0,0 +1,183 @@
...
+  children,
+  fullWidth,
+  headerColorScheme,
+  headerContent,
+  headerIcon,
+  headerShowCloseButton,
+  hugColor,
+  scrimDismiss,
+  sheet,
+  showCloseButton,
```
craigpalermo, line 121: Would using the same prop for `showCloseButton` and `headerShowCloseButton` cause a functional difference?

packages/core/src/Dialog/Dialog.stories.args.tsx
trenaryja: Unintentional, this is now handled inside the `exampleHeaderProps` within the story file

packages/core/src/Dialog/Dialog.styled.tsx
trenaryja: The intention was to show that the scrim/overlay is clickable in order to be dismissed, but now looking at:
- [chakra](https://chakra-ui.com/docs/components/modal/usage)
- [bootstrap](https://getbootstrap.com/docs/4.0/components/modal/)
- [material-ui](https://mui.com/material-ui/react-modal/)
and even the radix dialog itself, you're right and this isn't a feature probably because the confusion you brought up! I will remove this 😄

packages/core/src/Dialog/Dialog.tsx
trenaryja: This is a remnant from when I had the props named `aria-title` and `aria-description` but I've since changed those prop names. I will remove this line

packages/core/src/Dialog/Dialog.styled.tsx
trenaryja: Yes, if both are enabled then it is possible to get an experience like this:

image

The end user would only use whichever option fulfills the design spec. Both, when clicked, perform the same actions though!

packages/core/src/Dialog/Dialog.styled.tsx
trenaryja: I agree, and I made this change. But then I went behind, and added `borderRadius` as a prop for the Dialog as a whole 😄

craigpalermo (approved):

PR #1330, opened 2023-07-11
Add colorScheme prop to Hug

craigpalermo:
CleanShot 2023-07-11 at 13 26 32

### DIFF
--- a/packages/core/src/Box/Box.tsx
+++ b/packages/core/src/Box/Box.tsx
@@ -44,7 +44,7 @@ import {
   paletteFamilies,
   roundedValues,
 } from '../utils'
-import { ColorSchemes } from '../theme'
+import type { ColorSchemeName } from '../theme'
 
 export const boxPropTypes = {
   ...propTypes.boxShadow,
@@ -121,7 +121,7 @@ export interface IBoxProps
     | 'border'
     | 'background'
   onClick?: (unknown) => unknown
-  colorScheme?: keyof ColorSchemes
+  colorScheme?: ColorSchemeName
 }
 
 const Box: React.FC<InferProps<typeof boxPropTypes>> = styled.div.attrs((props) => ({
--- a/packages/core/src/Hug/Hug.stories.tsx
+++ b/packages/core/src/Hug/Hug.stories.tsx
@@ -1,24 +1,26 @@
 import React from 'react'
+import { Star } from 'pcln-icons'
 
 import { Hug, Hide, Card, Text } from '..'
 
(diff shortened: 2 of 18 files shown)


packages/core/src/Hug/Hug.stories.tsx
```diff
@@ -1,24 +1,26 @@
...
 
 import { Hug, Hide, Card, Text } from '..'
 
 import { argTypes, defaultArgs, iconMap, HugContentText } from './Hug.stories.args'
+import { colorSchemeNames } from '../utils/createColorScheme'
 
 const ChildrenCardInside = () => (
   <Card p={3} color='warning.light'>
-    I‘m a card within a hug!
+    I&apos;m a card within a hug!
```
sdalonzo, line 11: 😍

packages/core/src/Box/Box.tsx
```diff
@@ -44,7 +44,7 @@ import {
   paletteFamilies,
   roundedValues,
 } from '../utils'
-import { ColorSchemes } from '../theme'
+import { ColorSchemeName } from '../storybook/args'
```
pcln-james, line 47: In my mind, storybook and tests are like dev deps and shouldn't make their way into source code. Is this an accident or by design?

packages/icons/src/index.js
```diff
@@ -276,4 +276,4 @@ export { default as Whirlpool } from './Whirlpool.jsx'
 export { default as Widgets } from './Widgets.jsx'
 export { default as Wifi } from './Wifi.jsx'
 export { default as Youtube } from './Youtube.jsx'
-export { default as ZoomOut } from './ZoomOut.jsx'
+export { default as ZoomOut } from './ZoomOut.jsx'
```
pcln-james, line 279: was this intentional?

packages/core/src/Box/Box.tsx
craigpalermo: I don't have a better place to put the type we use in the source code since it's derived from an array of values that are only used in stories that live in the same file.

craigpalermo: @sdalonzo I think it'd make more sense to colocate these values with the `createColorScheme` function now that they're not used solely for story args. Thoughts?

pcln-james (approved):
Looks good to me! My one comment is nb.

packages/core/src/Box/Box.tsx
sdalonzo: they are in both. storybook/args just re-exports for convenience in stories, and i'd only use it for stories.

sdalonzo (approved):

PR #1331, opened 2023-07-11
Fix prettier formatting inconsistencies

craigpalermo:
This PR reverts the formatting changes in https://github.com/priceline/design-system/pull/1322 that didn't respect our Prettier config. Using the same `prettier` from our `rush-prettier` auto-installer should allow us to periodically format all files without worrying about which version of Prettier is globally installed.


pcln-james (approved):
needs a changelog even if it's a none update

sdalonzo (approved):

PR #1333, opened 2023-07-11
add text-transform support to Text

amyamccaskill:
- Add support for text-transfom to Text Component
  -  kept existing support provided by `caps` prop until we ensure use of that is replaced by `textTransform` prop

Screenshot 2023-07-11 at 3 32 20 PM

### DIFF
--- a/packages/core/src/Text/Text.spec.tsx
+++ b/packages/core/src/Text/Text.spec.tsx
@@ -85,8 +85,13 @@ describe('Text', () => {
     expect(screen.getByText('Italic Text')).toHaveStyleRule('font-style', 'italic')
   })
 
-  test('textTransform', () => {
+  test('textDecoration', () => {
     render(<Text textDecoration='underline'>Underlined Text</Text>)
     expect(screen.getByText('Underlined Text')).toHaveStyleRule('text-decoration', 'underline')
+  })
+
+  test('textTransform', () => {
+    render(<Text textTransform='lowercase'>Lowercase Text</Text>)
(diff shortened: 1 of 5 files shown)


packages/core/src/Text/Text.spec.tsx
```diff
@@ -85,8 +85,15 @@ describe('Text', () => {
...
 
-  test('textTransform', () => {
+  test('textDecoration', () => {
     render(<Text textDecoration='underline'>Underlined Text</Text>)
     expect(screen.getByText('Underlined Text')).toHaveStyleRule('text-decoration', 'underline')
   })
+
+  test('textTransform', () => {
+    const { asFragment } = render(<Text textTransform='lowercase'>Lowercase Text</Text>)
+    expect(asFragment()).toMatchSnapshot()
```
pcln-james, line 95: we don't need snapshots if you're testing the actual behaviour

packages/core/src/Text/Text.spec.tsx
```diff
@@ -85,8 +85,15 @@ describe('Text', () => {
     expect(screen.getByText('Italic Text')).toHaveStyleRule('font-style', 'italic')
   })
 
-  test('textTransform', () => {
+  test('textDecoration', () => {
     render(<Text textDecoration='underline'>Underlined Text</Text>)
     expect(screen.getByText('Underlined Text')).toHaveStyleRule('text-decoration', 'underline')
   })
+
+  test('textTransform', () => {
```
pcln-james, line 93: nb: might be valuable to test the other text transforms but not a huge deal either way.

pcln-james (approved):
Nice work!

PR #1334, opened 2023-07-11
Input and Select readonly state and styles

pcln-james:
https://priceline.atlassian.net/browse/UXPT-4990


packages/core/src/Select/Select.tsx
```diff
@@ -47,7 +47,8 @@ const SelectBase: React.FC<InferProps<typeof propTypes>> = styled.select.attrs(b
     display: none;
   }
 
-  &:disabled {
+  &:disabled,
+  &[readOnly] {
```
pcln-james, line 51: Select html tags don't actually support `read-only` but they do accept the attribute so this selector works. This is also why below I have to add the disabled state to all of the options if `readOnly` because otherwise you can still click on them.

DerrickTheodore (approved):
LGTM

PR #1335, opened 2023-07-11
feat(UXPT-4964): Create ChatMessage component

pcln-james:
https://priceline.atlassian.net/browse/UXPT-4964

Create `ChatMessage` component

### DIFF
--- a/packages/core/src/ChatMessage/ChatMessage.stories.tsx
+++ b/packages/core/src/ChatMessage/ChatMessage.stories.tsx
@@ -1,7 +1,5 @@
 import React from 'react'
 import type { StoryObj } from '@storybook/react'
-import { expect } from '@storybook/jest'
-import { within } from '@storybook/testing-library'
 import { Flex, Text } from '..'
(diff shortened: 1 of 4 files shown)


sdalonzo (approved):

packages/core/src/ChatMessage/ChatMessage.stories.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react'
+import type { StoryObj } from '@storybook/react'
+import { Flex, Text } from '..'
+import { Overnight } from 'pcln-icons'
+import ChatMessage, { IChatMessage } from './ChatMessage'
+
+const variations = ['initial', 'incoming', 'outgoing']
```
craigpalermo, line 7: What if you exported this from the file the component is in? Then you could use the same pattern as I did for color scheme values:

```

export const colorSchemeNames = [
  'primary',
...
] as const

export type ColorSchemeName = typeof colorSchemeNames[number]
```

craigpalermo (approved):

PR #1339, opened 2023-07-12
feature(UXPT-4319): add background-position prop to BackgroundImage component

marco-ch3ung:


wesleymartin85:
@marco-ch3ung can you add a story and control for this prop

common/changes/pcln-design-system/feature-4319-background-position-prop_2023-07-12-19-30.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "pcln-design-system",
+      "comment": "add background-position prop to BackgroundImage component",
+      "type": "patch"
```
craigpalermo, line 6: n-b API additions should be minors

packages/core/src/BackgroundImage/BackgroundImage.stories.tsx
```diff
@@ -80,3 +80,11 @@ Rounded.args = {
   rounded: 'top',
   height: '100px',
 }
+
+export const BackgroundPosition = Template.bind({})
+BackgroundPosition.args = {
+  height: 400,
+  width: '360px',
+  children: <StoryChildren />,
+  backgroundPosition: 'center',
```
craigpalermo, line 89: You can omit `backgroundPosition` from the args since it defaults to `center`

packages/core/src/BackgroundImage/BackgroundImage.tsx
```diff
@@ -50,10 +51,11 @@ export interface IBackgroundImageProps
   image?: string
   borderRadius?: string
   rounded?: string
+  backgroundPosition?: string
```
craigpalermo, line 54: If you follow this example, you can define an array value for use in the propTypes at line 42, then use that array to create a type for use in the props interface at 54. The type for `backgroundPosition` must reflect the allowed values, not any string.

https://www.typescriptlang.org/play?#code/MYewdgzgLgBANgS2jAvDA2gcgIaYDQyYBG+hwmAujNhDKJFANwwD0LMAKgMoDMAdABYYEAJ5go2AB4AoKCIAOAUxgA5RYoAmmgKpgE4DguVo5SkADN4SKOjABXALZFFAJwrM2hXAB9imX+SM0kA

packages/core/src/BackgroundImage/BackgroundImage.stories.args.ts
```diff
@@ -73,4 +73,17 @@ export const argTypes = {
       defaultValue: { summary: undefined },
     },
   },
+
+  backgroundPosition: {
+    name: 'backgroundPosition',
+    type: { name: 'string', required: false },
+    description: 'background position',
+    options: ['top', 'bottom', 'left', 'right', 'center'],
```
craigpalermo, line 81: You should be able to get the array of options as an export from the `BackgroundImage` component file (see my comment there first)

craigpalermo (changes requested):
Looks good overall, but there are a couple of places we can reduce duplication and make the type for the new prop more accurate.

common/changes/pcln-design-system/feature-4319-background-position-prop_2023-07-12-19-30.json
marco-ch3ung: thanks

packages/core/src/BackgroundImage/BackgroundImage.tsx
```diff
@@ -39,6 +40,7 @@ const backgroundImagePropTypes = {
   variation: PropTypes.oneOf(Object.keys(variations)),
   borderRadiusSize: PropTypes.oneOf(borderRadiusValues),
   rounded: PropTypes.oneOf(roundedValues),
+  backgroundPosition: PropTypes.oneOf(['top', 'bottom', 'left', 'right', 'center']),
```
craigpalermo, line 43:
```suggestion
  backgroundPosition: PropTypes.oneOf(backgroundPositionList),
```

marco-ch3ung: good catch! I missed here.

packages/core/src/BackgroundImage/BackgroundImage.stories.tsx
```diff
@@ -80,3 +80,10 @@ Rounded.args = {
   rounded: 'top',
   height: '100px',
 }
+
+export const BackgroundPosition = Template.bind({})
+BackgroundPosition.args = {
+  height: 400,
+  width: '360px',
+  children: <StoryChildren />,
```
pcln-james, line 88: should this have a background-position set in args?

marco-ch3ung: we can omit backgroundPosition from the args since it defaults to center

pcln-james: The chromatic snapshot should probably picture something different from the default behaviour and I think that is based on the args passed in initially. Could we add `backgroundPosition: top` or something please

marco-ch3ung: yea sure

pcln-james (approved):

PR #1343, opened 2023-07-13
fix(UXPT-4813): storybook to show how to make card focusbale

pherrera167:
Small PR that changes one of the storybook stories to demonstrate how to make a card focusable and tab-able by adding `as = 'button'`

### DIFF
--- a/apps/storybook/.storybook/main.ts
+++ b/apps/storybook/.storybook/main.ts
@@ -17,7 +17,7 @@ const storiesGlob = 'src/**/*.stories.@(js|jsx|ts|tsx)'
 const config: StorybookConfig = {
   framework,
   stories: packagesWithStories.map((packageFolder) => `../../../packages/${packageFolder}/${storiesGlob}`),
-  addons: storykitAddons,
+  addons: [...storykitAddons, 'storybook-addon-pseudo-states'],
   docs: {
     autodocs: true,
   },
--- a/packages/core/src/Button/Button.stories.tsx
+++ b/packages/core/src/Button/Button.stories.tsx
@@ -62,6 +62,9 @@ export const Playground: ButtonStory = {
     },
   },
 }
+Playground.parameters = {
+  pseudo: { rootSelector: "body" },
+}
 
 export const Size: ButtonStory = {
   render: () => (
--- a/packages/core/src/Card/Card.stories.tsx
+++ b/packages/core/src/Card/Card.stories.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { Box, Card } from '..'
+import styled from 'styled-components'
 
 export default {
   title: 'Card',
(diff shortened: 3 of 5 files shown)


wesleymartin85:
@pherrera167 can you add to the story a card where the focus state is forced on using [Storybook Pseudo States](https://storybook.js.org/addons/storybook-addon-pseudo-states)

apps/storybook/.storybook/main.ts
```diff
@@ -17,7 +17,7 @@ const storiesGlob = 'src/**/*.stories.@(js|jsx|ts|tsx)'
 const config: StorybookConfig = {
   framework,
   stories: packagesWithStories.map((packageFolder) => `../../../packages/${packageFolder}/${storiesGlob}`),
-  addons: storykitAddons,
+  addons: [...storykitAddons, 'storybook-addon-pseudo-states'],
```
craigpalermo, line 20: Any reason not to add this to `storykitAddons`?

packages/core/src/Card/Card.stories.tsx
```diff
@@ -74,6 +88,27 @@ export const BoxShadowsWithVaryingBorderRadii = () => (
...
+  pseudo: {rootSelector: "PABLO" },
+}
+
+export const Hover = () => (
+  <Box>
+  <StyledCard id='PABLO' boxShadowSize='sm'>
+    Card Text
+  </StyledCard>
+  </Box>
+)
```
craigpalermo, line 106: Can you try committing this file again without skipping the pre-commit hook? It looks like this is missing some formatting.

packages/core/src/Card/Card.stories.tsx
```diff
@@ -74,6 +88,27 @@ export const BoxShadowsWithVaryingBorderRadii = () => (
...
 )
 
+export const Default = () => (
+  <a id='PABLO' href='www.priceline.com'>
+    Pablo
+  </a>
+)
+Default.parameters = {
+  pseudo: {rootSelector: "PABLO" },
+}
```
craigpalermo, line 98: It looks like this is creating a story called `Default` that doesn't use Card. Is that intentional?

packages/core/src/Card/Card.stories.tsx
sdalonzo: strongly recommend using the `prettier` vscode extension + format on save. it will find our prettier config

packages/core/src/Card/Card.stories.tsx
sdalonzo: we shouldn't leave ids with our names in the sourcecode, anyway

packages/core/src/Button/Button.stories.tsx
```diff
@@ -62,6 +62,9 @@ export const Playground: ButtonStory = {
     },
   },
 }
+Playground.parameters = {
+  pseudo: { rootSelector: "body" },
```
sdalonzo, line 66: prettier here too - this should be single quoted

apps/storybook/.storybook/main.ts
sdalonzo: i agree with this sentiment. if there are challenges, we could defer, but it should be our preferred approach as we'll need it in pcln-web

PR #1344, opened 2023-07-13
Fix hug color scheme

craigpalermo:
Updated requirements: Using `colorScheme` on `Hug` should not set the header color to the color scheme's foreground by default.

### DIFF
--- a/packages/core/src/Hug/Hug.styled.ts
+++ b/packages/core/src/Hug/Hug.styled.ts
@@ -1,8 +1,9 @@
 import styled from 'styled-components'
 import { themeGet } from '@styled-system/theme-get'
 import { Card } from '../Card'
-import { applyVariations, colorScheme } from '../utils'
+import { applyVariations, colorSchemeCustomForeground } from '../utils'
 import { Flex, Box } from '..'
+import type { IFlexProps } from '../Flex'
 
 export const HugCard = styled(Card)`
   overflow: hidden;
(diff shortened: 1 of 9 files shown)


packages/core/src/Hug/Hug.styled.ts
```diff
@@ -29,5 +29,23 @@ export const BorderConcealer = styled(Box)`
 `
 
 export const Header = styled(Flex)`
-  ${colorScheme}
+  ${colorSchemeCustomForeground}
+
+  ${({ colorScheme, theme, iconUsesColorScheme }) => {
```
sdalonzo, line 34: n-b i think there's the makings of a utility here

sdalonzo (approved):

packages/core/src/theme/theme.ts
```diff
@@ -521,6 +521,7 @@ export type ColorStyles = Record<ColorStyleName, ColorStyle>
 export type ColorScheme = {
   foreground: string
   background: string
+  backgroundName: string
```
trenaryja, line 524: This could be typescript-y. Here's a proposal:

```typescript
type _PaletteFamily = Array<keyof PaletteFamilies>[number]
type _PaletteFamilyOption = Array<keyof PaletteFamily>[number]
export type PaletteColor = `${_PaletteFamily}.${_PaletteFamilyOption}`
```

With this, the type of `backgroundName` can be `PaletteColor`. These types might be useful in general for other props on other components too 🤔

trenaryja (approved):

PR #1346, opened 2023-07-13
Add support for text-transform to badge

amyamccaskill:
- Add support for text-transform to badge
- Set `uppercase` as the default textTransform prop
- Move `textTransform` from Text.tsx to the _very_ bottom of utils.ts (looks like the rest of the file was formatted here, let me know if I should undo that part 😬)

Screenshot 2023-07-13 at 3 18 57 PM

### DIFF
--- a/packages/core/src/Badge/Badge.spec.tsx
+++ b/packages/core/src/Badge/Badge.spec.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { render, screen } from '../__test__/testing-library'
 import { Badge, theme } from '..'
 
 describe('Badge', () => {
@@ -80,5 +81,11 @@ describe('Badge', () => {
     expect(json).toMatchSnapshot()
     expect(json).toHaveStyleRule('background-color', theme.colors.lightBlue)
     expect(json).toHaveStyleRule('color', theme.colors.text)
+  })
+
+  test('textTransform', () => {
+    render(<Badge textTransform='lowercase'>Lowercase Text</Badge>)
+    expect(screen.getByText('Lowercase Text')).toHaveStyleRule('text-transform', 'lowercase')
+    expect(screen.getByText('Lowercase Text')).toHaveStyleRule('letter-spacing', 'normal')
   })
 })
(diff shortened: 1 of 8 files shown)


pcln-james (approved):
LGTM

packages/core/src/Badge/Badge.spec.tsx
```diff
@@ -81,4 +82,10 @@ describe('Badge', () => {
     expect(json).toHaveStyleRule('background-color', theme.colors.lightBlue)
     expect(json).toHaveStyleRule('color', theme.colors.text)
   })
+
+  test('textTransform', () => {
+    render(<Badge textTransform='lowercase'>Lowercase Text</Badge>)
+    expect(screen.getByText('Lowercase Text')).toHaveStyleRule('text-transform', 'lowercase')
+    expect(screen.getByText('Lowercase Text')).toHaveStyleRule('letter-spacing', 'normal')
+  })
```
craigpalermo, line 90: In the future, you don't necessarily need to write a unit test as long as you add a new story that demonstrates the functionality since we'll see any diffs in Chromatic.

packages/core/src/utils/utils.ts
craigpalermo, line 1: Can you check this file out from `main` and add your change again? There shouldn't be any formatting diffs after my recent changes to make formatting consistent. The pre-commit should also handle this.

packages/core/src/Badge/Badge.stories.tsx
```diff
@@ -41,6 +41,20 @@ LightBlueAndTextCustom.args = {
   color: 'text',
 }
 
+const TextTransformTemplate = () => {
+  return (
+    <React.Fragment>
+      {['uppercase', 'Lowercase', 'capitalize', 'none'].map((textTransform) => (
```
craigpalermo, line 47: Can you replace this array with the `textTransformValues` import from utils?

craigpalermo (commented):
Change LGTM once we resolve the formatting diff in utils

craigpalermo (approved):

PR #1347, opened 2023-07-14
[UXPT-4944] Create RadioCheckToggleCard

pcln-jason-v:
**Ticket**
https://priceline.atlassian.net/browse/UXPT-4944

**Screenshots**
image
image
image

image


packages/core/src/RadioCheckToggleCard/RadioCheckToggleCard.stories.tsx
```diff
@@ -0,0 +1,1013 @@
...
+    docs: {
+      page: () => <DefaultDocPage />,
+    },
+  },
+}
+export default meta
+
+export const Radio: Story = {
+  args: { cardType: 'radio', name: 'radio-card', title: 'Title' },
+  argTypes: {
```
craigpalermo, line 939: `argTypes` should go in a separate `RadioCheckToggleCard.stories.args.tsx` to keep the stories file focused

packages/core/src/RadioCheckToggleCard/RadioCheckToggleCard.stories.tsx
```diff
@@ -0,0 +1,1013 @@
...
+            >
+              <Box width='100%' height='75px' />
+            </RadioCheckToggleCard>
+          </HoverStateCard>
+
+          <FocusStateCard style={{ flex: '1 1 30%' }}>
+            <RadioCheckToggleCard
+              cardType='radio'
+              hPosition='right'
+              vPosition='top'
```
craigpalermo, line 252: You can omit these three props everywhere since they're the same as the defaults for `RadioCheckToggleCard`

packages/core/src/RadioCheckToggleCard/RadioCheckToggleCard.stories.tsx
```diff
@@ -0,0 +1,1013 @@
...
+              title='Selected Hover'
+              value='example-2'
+              isSelected={true}
+              onChange={() => {}}
+            >
+              <Box width='100%' height='75px' />
+            </RadioCheckToggleCard>
+          </SelectedHoverStateCard>
+
+          <SelectedFocusStateCard style={{ flex: '1 1 30%' }}>
```
craigpalermo, line 483: Could you include the `flex` style in the styled component instead of using the `style` prop?

packages/core/src/RadioCheckToggleCard/RadioCheckToggleCard.tsx
```diff
@@ -0,0 +1,216 @@
...
+            size='24'
+            onChange={handleChange}
+            style={{
+              ['--icon-border-radius']: cardType === 'checkbox' ? '4px' : '9999px',
+              borderWidth: '1px',
+              margin:
+                hPosition === 'left'
+                  ? `0 var(--default-card-margin) 0 0`
+                  : `0 0 0 var(--default-card-margin)`,
+            }}
```
craigpalermo, line 197: Using `style` and CSS variables is a bit out of pattern for DS; could you achieve the same effect using props and Styled Components?

packages/core/src/RadioCheckToggleCard/RadioCheckToggleCard.tsx
```diff
@@ -0,0 +1,216 @@
...
+        <CardHeader
+          alignItems={vPosition}
+          flexDirection={hPosition === 'left' ? 'row-reverse' : 'row'}
+          data-testId='rdt-card-header'
+        >
+          <CardContent>
+            <Text
+              bold={isTitleBold}
+              fontSize={isTitleBold ? '16px' : '14px'}
+              style={{ lineHeight: '24px' }}
```
craigpalermo, line 179: Please refrain from using the `style` prop and prefer Styled Components if possible.

packages/core/src/RadioCheckToggleCard/RadioCheckToggleCard.tsx
```diff
@@ -0,0 +1,216 @@
...
+    name,
+    value,
+    onChange,
+    isSelected,
+  } = props
+
+  const id = `${name}_${value}`
+  const DisplayButton = buttonIcon(cardType)
+
+  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => onChange(e)
```
craigpalermo, line 157: `onChange` isn't a required prop, right? I think this should have a conditional only to execute `onChange` if it's present.

packages/core/src/RadioCheckToggleCard/RadioCheckToggleCard.tsx
```diff
@@ -0,0 +1,216 @@
...
+    --icon-border-color: ${getPaletteColor('primary.dark')};
+  }
+
+  &:disabled + div {
+    color: ${getPaletteColor('border.base')};
+  }
+`
+
+const RCTCardContainer = styled(Flex)`
+  --background-color: ${getPaletteColor('background.lightest')};
```
craigpalermo, line 72: Is there an advantage to using CSS variables here? It's a novel pattern for DS, so I want to be sure there's justification for doing it this way.

packages/core/src/RadioCheckToggleCard/RadioCheckToggleCard.tsx
```diff
@@ -0,0 +1,216 @@
...
+export type TCardTypes = (typeof cardTypes)[number]
+export type THPositions = (typeof hPositions)[number]
+export type TVPositions = (typeof vPositions)[number]
+
+export interface IRadioCheckToggleCard {
+  children?: React.ReactNode | string
+  cardType?: TCardTypes
+  hPosition?: THPositions
+  vPosition?: TVPositions
+  isHeightOccupied?: boolean
```
craigpalermo, line 21: It's not clear (at least to me) what `height` refers to here

packages/core/src/RadioCheckToggleCard/RadioCheckToggleCard.tsx
```diff
@@ -0,0 +1,216 @@
...
+  title: string
+  isTitleBold?: boolean
+  name: string
+  value: string
+  onChange: (e: React.ChangeEvent<HTMLInputElement>) => void
```
craigpalermo, line 26: `onChange` shouldn't be a required prop IMO. This component should function as expected in a plain old HTML `<form>` if desired.

craigpalermo (changes requested):

packages/core/src/RadioCheckToggleCard/RadioCheckToggleCard.tsx
pcln-jason-v: Hey @craigpalermo, I used the CSS variables here because it is better to manage with different items (like background color, icon color, border color) and easy to see what is going to change between different states (like hover, focus and active) rather than keep setting the same CSS properties.  

If it is a bit out of pattern for DS. I am ok to change another pattern for them.

craigpalermo (approved):

sdalonzo (approved):

PR #1348, opened 2023-07-14
UXPT-4880: Add chromatic interaction test to popover

coolestKev:
Adding simple hover interaction to the popover component to test if open/close action is triggered. 
There's no easy way to check the popover content yet, since its outside the `storybook-root` div, and chromatic interaction tests only check the content within the div.

A nice to have for the future is a method to render popover inside the story book root.
Screenshot 2023-07-17 at 12 58 30 PM
Screenshot 2023-07-17 at 12 58 25 PM


packages/popover/src/Popover/Popover.stories.jsx
```diff
@@ -212,8 +214,17 @@ const InnerContent = ({ handleClose }) => (
...
 )
 
-export const PassesThemeToContent = () => {
-  return (
+export const PassesThemeToContent = {
+  play: async ({ args, canvasElement }) => {
+    const canvas = within(canvasElement)
+    const text = canvas.getByText('hello world')
+    expect(text).toBeInTheDocument()
+    await userEvent.hover(text)
```
craigpalermo, line 222: Shouldn't this be a click event?

coolestKev: Hmm, seems like the `openOnHover` is set to true as part of the args: 

https://github.com/priceline/design-system/pull/1348/files#diff-d81c37b8b7a054bb7e0a0ed0441ad294a5fe202bd3b9667df8750d76647c5001R21

craigpalermo (approved):

sdalonzo (approved):
just needs `rush change`

PR #1364, opened 2023-07-20
Migration to Vite

sdalonzo:
### pcln-slider

- build converted to vite, with dts emission
- test converted to vitest, with v8 coverage reports and vitest/ui interface
- eslint
- typechecking with checker and `tsc --no-emit`

NOTE: `pcln-design-system` is still WIP and not yet committed.


packages/slider/tsconfig.node.json
craigpalermo, line 1: What's the reason for having a separate TS config for Node?

packages/core/package.json
```diff
@@ -69,69 +46,125 @@
...
+    "vite:dev": "vite",
+    "build": "vite build",
+    "build:visualize": "vite-bundle-visualizer",
+    "build:visualize:sunburst": "vite-bundle-visualizer --template sunburst",
+    "build:visualize:network": "vite-bundle-visualizer --template network",
+    "build:visualize:data": "vite-bundle-visualizer --template raw-data --output './vite-bundle-visualizer/stats.json'",
+    "build:visualize:list": "vite-bundle-visualizer --template list --output './vite-bundle-visualizer/stats.yaml'",
+    "build:watch": "vite build --watch",
+    "lint": "eslint src --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "vite:preview": "vite preview",
```
craigpalermo, line 80: @sdalonzo I tried running `vite:dev` and `vite:preview` but they don't seem to do anything. Should we remove those to avoid confusion?

craigpalermo (approved):

packages/core/package.json
sdalonzo: done

PR #1366, opened 2023-07-21
feat(UXPT-4815) Add shadow variation to ButtonChip, FilterChip and ChoiceChip

DerrickTheodore:


packages/core/src/Chip/ChipContentWrapper.tsx
```diff
@@ -31,6 +31,18 @@ const getBorderColor = (props) =>
 const getBackgroundColor = (props) =>
   props.disabled ? 'background.base' : props.selected ? 'light' : 'background.lightest'
 
+const variations = {
+    outline: css` 
+    `,
+    shadow: css`
+      border-color: transparent;
```
DerrickTheodore, line 38: @wesleymartin85 

> the border here should be set to the same color as the white background. (this is just for the default enabled state the primary.dark border on hover is correct. 

Are you asking to make the border-color exactly what the background color is in the stories, or make it transparent so it's what ever color the background is?

wesleymartin85: its not for the storyies but part of the variation settings, look that the styles in the [prototype](https://www.figma.com/proto/gpbXA3aonRBAJbQiZvrayO/Design-Ops-Dev-Week?page-id=0%3A1&type=design&node-id=564-22482&viewport=-25300%2C2359%2C1&scaling=min-zoom&starting-point-node-id=564%3A22482&t=eLmmLiEs6OFOYLsR-1&show-proto-sidebar=1) and [designs](https://www.figma.com/file/gpbXA3aonRBAJbQiZvrayO/Design-Ops-Dev-Week?type=design&node-id=215-7397&mode=design&t=0gi2R1mn5lWSAqlS-11)

DerrickTheodore: Yea, I'm just uncertain if you want the border-color to be a static color or should it be the color of whatever the background is that the chip is on top of.  I think we want it to be transparent so it's whatever the background color is, but not 100% sure.   Currently it's set to transparent, so let me know how that looks in storybook.

DerrickTheodore:
@wesleymartin85 you should be able to review the `Usage` stories we discussed now in Chromatic.  Can you give them a try for me, to make sure I didn't miss anything that's being done on the prototype.

packages/core/config/jest.config.json
```diff
@@ -2,7 +2,7 @@
   "extends": "@priceline/heft-component-rig/profiles/react/config/jest.config.json",
   "coverageThreshold": {
     "global": {
-      "branches": 91,
+      "branches": 90.89,
```
DerrickTheodore, line 5: I put a lot of effort into trying to test the conditional states from the hover event, but everything I tried didn't work.  Ultimately I came upon [this](https://github.com/testing-library/jest-dom/issues/59) when trying it on interaction test.  I didn't try bumping that underline dependency mentioned in the that thread though; I had already invested more time into troubleshooting than I could afford at that point.  So would it be ok to drop this because I have a stories for the unit testing I was trying to do?

packages/core/src/Chip/ChipContentWrapper.tsx
```diff
@@ -78,6 +98,8 @@ const ChipContentWrapper: React.FC<IChipContentWrapper> = styled(Box)`
   position: relative;
   border-radius: 2px;
   white-space: nowrap;
+  ${({ theme, hasChildren }) => applySizes(getSizes({ hasChildren }), undefined, theme.mediaQueries)};
+  ${applyVariations('ChipContentWrapper', variations)};
```
DerrickTheodore, line 102: Line 102 is my change, and 101 is the result of reformatting after merging with main, and not part of my changes.

craigpalermo (approved):

sdalonzo (approved):

PR #1376, opened 2023-08-02
ChoiceChip: Add a topLabel prop and ability to set borderRadius and w…

ryan-hunt-priceline:
Add new props to `ChoiceChp`
  **topLabel**:  label that will display above the **label** prop
  **borderRadius**: allow user to override borderRadius
  **width**: allow user to set the width
  **justifyContent**: allow user to override the justifyContent (Default "center")

### DIFF
--- a/packages/core/src/Chip/ChipContent/ChipContent.tsx
+++ b/packages/core/src/Chip/ChipContent/ChipContent.tsx
@@ -54,6 +54,7 @@ const ChipContent: React.FC<InferProps<typeof propTypes>> = ({
   BridgeIcon = undefined,
   topLabel = undefined,
   borderRadius = undefined,
+  justifyContent = 'center',
   ...props
 }) => (
   <ChipContentWrapper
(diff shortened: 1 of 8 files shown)


packages/core/src/Chip/ChipContent/ChipContent.tsx
```diff
@@ -49,36 +52,53 @@ const ChipContent: React.FC<InferProps<typeof propTypes>> = ({
...
+    <Flex alignItems='center' height='100%'>
+      {Boolean(image) && <ImageWrapper disabled={disabled}>{image}</ImageWrapper>}
+      {Boolean(Icon) && <Icon title={IconTitle} size='20px' />}
+      <Box>
+        {Boolean(topLabel) && (
+          <Text regular ml={2} mb={1} fontWeight='bold'>
+            {topLabel}
+          </Text>
+        )}
+        <Flex>
```
nmalayev, line 81: Using <Flex> here changes the display property to 'flex' vs the original `inline-flex` assigned in `<ChipContentWrapper>`. Is that ok? Same goes for line 72.

ryan-hunt-priceline: Good callout. Everything was looking good on regression though, Chromatic tests showed all the existing components looked the same

nmalayev (approved):

PR #1378, opened 2023-08-03
Add borderStyles to Card component

pcln-james:
Adding the `borderStyle` prop to the Card component because I need to use it for a component I'm working on and assumed others might be able to benefit from this too.

https://styled-system.com/api/#border

I was torn between adding this to Card or to Box and therefore everything would get the benefit of it too. I can go back and make this change if people prefer that approach.

### DIFF
--- a/packages/core/src/storybook/args/index.ts
+++ b/packages/core/src/storybook/args/index.ts
@@ -15,6 +15,21 @@ export const colors = ['', ...paletteFamilies, 'NOTVALID']
 
 export const borderRadii = ['', ...borderRadiusValues, 'NOTVALID']
 
+export const borderStyles = [
+  'none',
+  'hidden',
+  'dotted',
+  'dashed',
+  'solid',
+  'double',
+  'groove',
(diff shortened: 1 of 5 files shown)


craigpalermo (approved):

trenaryja (approved):

packages/core/src/storybook/args/index.ts
```diff
@@ -15,6 +15,21 @@ export const colors = ['', ...paletteFamilies, 'NOTVALID']
...
+  'dashed',
+  'solid',
+  'double',
+  'groove',
+  'ridge',
+  'inset',
+  'outset',
+  'initial',
+  'inherit',
+]
```
huzefakagdi, line 31: Sorry for seeing the PR late. But we should enforce `as const` for every true constant. Provides excellent TypeScript support in IDE while looking at the plausible values for `borderStyles`, even in JS-only projects. IMHO reduces the need to look at the documentation
CC @craigpalermo @trenaryja

PR #1379, opened 2023-08-08
Replace more Jest tests with Storybook stories and interaction tests

craigpalermo:
I'm not changing anything in the `modal` package because we're going to deprecate it in favor of the new `Dialog` component.


packages/icons/package.json
```diff
@@ -42,6 +38,7 @@
     "@priceline/babel-preset": "workspace:*",
     "@priceline/eslint-config": "workspace:*",
     "@rushstack/eslint-patch": "^1.1.0",
+    "@storybook/jest": "^0.1.0",
```
craigpalermo, line 41: This is the only change, the rest are from sorting package.json.

pcln-james (approved):

PR #1385, opened 2023-08-23
Accordion types use typescript

pcln-james:
Found an issue with Accordion types while trying to use the latest version of the design system. Not sure if this is due to my typescript config in my app or if this would never work but I decided to fix it this way.

### DIFF
--- a/packages/core/src/Accordion/Accordion.tsx
+++ b/packages/core/src/Accordion/Accordion.tsx
@@ -8,7 +8,7 @@ import { Box, Flex } from '..'
 export interface IAccordion {
   items: IAccordionItem[]
   itemsState?: string | string[]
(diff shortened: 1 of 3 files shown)


packages/core/src/Accordion/Accordion.tsx
```diff
@@ -30,6 +29,7 @@ export const Accordion = ({
 }: IAccordion) => {
   return items ? (
     <RadixAccordion.Root
+      // @ts-ignore
```
pcln-james, line 32: How Radix sets up their TS type makes it impossible (at least to my knowledge) to satisfy the type check... not sure why this is possible with TS or at least so difficult and unintuitive to resolve

sdalonzo (approved):

PR #1396, opened 2023-09-11
Fix broken interaction tests

pcln-james:
I found two broken interaction tests when working on storybook interaction test coverage. While that work has been put on hold I should still resolve these issues.


pcln-james:
I can't get this to pass Chromatic's build, closing for now

PR #1403, opened 2023-09-26
bugfix: Allows elements inside the Modal to overflow on the y axis

huzefakagdi:
### Description/Summary of changes

- Revert the `overflow-y: auto` change that was introduced in this [PR](https://github.com/priceline/design-system/pull/1319/files#diff-5b5f252abb56263bc4b86a7916c35586d4e9c96da30f35e18e66d1e9ed013eb2R57)
- The change was breaking the modals where form elements are hosted that naturally overflow the modal (when the elements are in an active state)

### Screenshots

Broken state
Modal Overflow missing

Current Production state 
image

### DIFF
--- a/packages/modal/src/Modal.stories.jsx
+++ b/packages/modal/src/Modal.stories.jsx
@@ -1,8 +1,7 @@
 import { action } from '@storybook/addon-actions'
-import { Box, Button, Text } from 'pcln-design-system'
+import { Box, Button, Accordion, Text } from 'pcln-design-system'
 import { Menu, MenuItem } from 'pcln-menu'
-import PropTypes from 'prop-types'
-import React from 'react'
+import React, { useState, useRef } from 'react'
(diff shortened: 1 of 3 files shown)


pherrera167 (approved):

packages/modal/src/Modal.stories.jsx
```diff
@@ -15,50 +14,32 @@ const CUSTOM_ANIMATION = (transitionState) => `
   transition: transform 0.3s linear;
   ${transitionState === 'entered' ? `transform: translateY(0%);` : ''}
 `
-
-class ModalStory extends React.Component {
-  constructor(props) {
-    super(props)
-    this.state = {
-      isOpen: props.isOpen,
+function ModalStory(props) {
```
huzefakagdi, line 17: Converted to a functional component

sdalonzo (approved):

PR #1414, opened 2023-11-15
add zIndex for Dialog

jinchi2013:


packages/core/src/Dialog/Dialog.tsx
```diff
@@ -25,6 +25,7 @@ export interface IDialogProps {
   showCloseButton?: boolean
   size?: DialogSize
   triggerNode?: React.ReactNode
+  zIndex?: number | string
```
craigpalermo, line 28: It looks like we have a type defined for this https://github.com/priceline/design-system/blob/main/packages/core/src/theme/theme.ts#L271C13-L271C21

jinchi2013: Will update to that, thanks

jinchi2013: Updated

pcln-james (approved):

sdalonzo (approved):

PR #1423, opened 2023-11-29
Fix/label z index

luizrochaecore:


DerrickTheodore (approved):
LGTM, the CI failure related to the snapshot change looks to be the result of your change, so you can resolve it by running `rush test:snapshots` as shown [here](https://github.com/priceline/design-system/blob/main/CONTRIBUTING.md#running-tests)

luizrochaecore:
thanks @DerrickTheodore , pushed a commit fixing the snapshot

pcln-james (approved):

PR #1424, opened 2023-12-05
feat(SITE-21794): fix the outgoing ChatMessage footer position issue

jinchi2013:
image

image

### DIFF
--- a/packages/core/src/ChatMessage/ChatMessage.tsx
+++ b/packages/core/src/ChatMessage/ChatMessage.tsx
@@ -55,6 +55,12 @@ function ChatMessage({
 }: IChatMessage) {
   const marginTop = header ? 3 : 0
   const marginBottom = footer ? 3 : 0
+  const footerPosition = variation === 'outgoing'
+    ? {
+      right: 0
+    }: {
+      left: 0
+    }
 
   return (
     <Message borderRadius='18px' variation={variation} mt={marginTop} mb={marginBottom} p='12px' {...props}>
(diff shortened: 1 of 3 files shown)


packages/core/src/ChatMessage/ChatMessage.tsx
```diff
@@ -79,7 +79,14 @@ function ChatMessage({
       )}
       <Text textStyle='paragraph2'>{message}</Text>
       {footer && (
-        <Absolute bottom='-20px' left='0px' minWidth='300px' width='100%' px={3}>
+        <Absolute
+          bottom='-20px'
+          left={variation === 'outgoing' ? '' : '0px'}
+          right={variation === 'outgoing' ? '0px' : ''}
```
DerrickTheodore, line 85: An empty string doesn't look to be a valid value for `left`, `right`, ect. ([source](https://developer.mozilla.org/en-US/docs/Web/CSS/left#values)).  

```
variation === 'outgoing' ?
<Absolute
     ...
    right={0}
    px={0}
>
   {footer}
</Absolute> : 
<Absolute
     ...
    left={0}
    px={3}
>
   {footer}
</Absolute> 
```

DerrickTheodore (approved):
LGTM - It'll also be helpful to add an additional story `OutGoingWithFooter`.  With that change, we can update the [WithFooter](https://github.com/priceline/design-system/blob/d3ce5c622ba571f37fbbc06841d2f233671c983e/packages/core/src/ChatMessage/ChatMessage.stories.tsx#L68) story to `IncomingWithFooter`, so these changes are represented in Chromatic.

sdalonzo (approved):

PR #1428, opened 2023-12-14
(AIR-59161):fix for loosing focus of menu after selecting menuitem

takhila499:
This Pr is to fix the accessibility issue, user lose the focus on Menu after selecting a menu item. The focus goes to top of the body after selecting any menu item. This issues is caused because of the popover position. In order to fix it we are passing the `querySelectorPortal` prop from menu.

### DIFF
--- a/common/changes/pcln-menu/chore-AIR-59161-FixAccessibilityMenu_2023-12-14-16-45.json
+++ b/common/changes/pcln-menu/chore-AIR-59161-FixAccessibilityMenu_2023-12-14-16-45.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "pcln-menu",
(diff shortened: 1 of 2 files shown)


common/changes/pcln-menu/chore-AIR-59161-FixAccessibilityMenu_2023-12-14-16-45.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "pcln-menu",
+      "comment": "Adding querySelectorPortal prop to menu item to fix the focus issue.",
+      "type": "patch"
```
pcln-james, line 6: this should be a minor since a new prop was added

pcln-james (approved):

PR #1430, opened 2023-12-26
Remove prop-types from more core components

craigpalermo:
This PR updates the remaining `core` components to use TS types for prop types instead of React prop types. It also includes cleanup of many imports/exports.

### DIFF
--- a/packages/core/src/BlockLink/BlockLink.spec.tsx
+++ b/packages/core/src/BlockLink/BlockLink.spec.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 
-import { BlockLink } from '..'
+import { BlockLink } from './BlockLink'
 
 describe('BlockLink', () => {
   test('renders', () => {
--- a/packages/core/src/BlockLink/index.ts
+++ b/packages/core/src/BlockLink/index.ts
@@ -1,1 +1,1 @@
-export { default as BlockLink } from './BlockLink'
+export { BlockLink } from './BlockLink'
--- a/packages/core/src/ChatMessageContainer/ChatMessageContainer.stories.tsx
+++ b/packages/core/src/ChatMessageContainer/ChatMessageContainer.stories.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Box } from '../Box'
 import type { StoryObj } from '@storybook/react'
-import ChatMessageContainer, { IChatMessageContainer } from './ChatMessageContainer'
+import { ChatMessageContainer, type ChatMessageContainerProps } from './ChatMessageContainer'
 import { shortConversation, longConversation } from './ChatMessageContainer.stories.args'
 
 export default {
@@ -12,7 +12,7 @@ export default {
   },
 }
 
-type ChatMessageContainerStory = StoryObj<IChatMessageContainer>
+type ChatMessageContainerStory = StoryObj<ChatMessageContainerProps>
 
 export const _ChatMessageContainer: ChatMessageContainerStory = {
   render: (args) => <ChatMessageContainer {...args} messages={shortConversation} />,
--- a/packages/core/src/Divider/Divider.stories.tsx
+++ b/packages/core/src/Divider/Divider.stories.tsx
@@ -1,6 +1,7 @@
 import React from 'react'
-import { Divider, Flex } from '..'
 import styled from 'styled-components'
+import { Divider } from '../Divider'
+import { Flex } from '../Flex'
 
 const description = 'Horizontal rule with settings for padding, margin, width, and borderColor'
 const ColumnFlex = styled(Flex)`
--- a/packages/core/src/FormField/index.ts
+++ b/packages/core/src/FormField/index.ts
@@ -1,5 +1,1 @@
-import type { IFormFieldProps } from './FormField'
-import FormField from './FormField'
-
-export type { IFormFieldProps }
-export { FormField, FormField as InputField }
+export { FormField, type FormFieldProps, FormField as InputField } from './FormField'
--- a/packages/core/src/Label/Label.tsx
+++ b/packages/core/src/Label/Label.tsx
@@ -1,30 +1,22 @@
 import React from 'react'
-import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import {
-  space,
+  FontSizeProps,
+  FontWeightProps,
+  LineHeightProps,
+  SpaceProps,
+  TextStyleProps,
+  WidthProps,
+  compose,
   fontSize,
   fontWeight,
   letterSpacing,
   lineHeight,
+  space,
   textStyle,
   width,
-  SpaceProps,
-  FontSizeProps,
-  FontWeightProps,
-  LineHeightProps,
-  TextStyleProps,
-  WidthProps,
-  compose,
 } from 'styled-system'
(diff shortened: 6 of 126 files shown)


packages/core/src/Layout/Layout.tsx
```diff
@@ -113,58 +113,46 @@ const getGapValues = (gapProp, rowGapProp) => {
...
-  flexWrap,
-  ...props
-}) => {
+type LayoutGap = (typeof ALLOWED_GAP_VALUES)[number] | Array<(typeof ALLOWED_GAP_VALUES)[number]>
+type LayoutVariation = (typeof ALLOWED_LAYOUT_VALUES)[number] | Array<(typeof ALLOWED_LAYOUT_VALUES)[number]>
+
+export type LayoutProps = FlexWrapProps & {
+  children: React.ReactElement | React.ReactElement[]
+  gap?: LayoutGap
+  rowGap?: LayoutGap
```
dansternfeld1, line 138: (n-b) So natural to look at them this way 👍

packages/core/src/Layout/Layout.tsx
```diff
@@ -173,15 +161,18 @@ const Layout: React.FC<InferProps<typeof propTypes>> = ({
...
+          React.isValidElement(child) && (
             <ZIndexBox
               width={widths[idx]}
               px={boxPaddingX}
               py={boxPaddingY}
               data-testid={`box-${idx}`}
-              zIndex={child.props.zIndex}
+              // TS ignored because the type of child.props is unknown
+              // @ts-ignore
+              zIndex={child?.props?.zIndex}
```
dansternfeld1, line 173: (n-b) Does this need to be `undefined` to avoid collisions with `0`/falsy values?
```suggestion
              zIndex={child?.props?.zIndex || undefined}
```

packages/core/src/Radio/__snapshots__/Radio.spec.tsx.snap
```diff
@@ -198,7 +198,7 @@ exports[`Radio Selected, rendering 1`] = `
 }
 
 .c0:hover > svg {
-  color: #049 !important;
+  color: #0068ef !important;
```
dansternfeld1, line 201: Ooh, is this an actual change caught by a snapshot?

dansternfeld1 (commented):
Really nice work here, @craigpalermo 👍 
Please take a look at the coverage level shifts, I think most of them can be resolved by changing where the test files point; other than those, just a couple of small questions before a ✅

packages/core/src/BlockLink/index.ts
```diff
@@ -1 +1 @@
-export { default as BlockLink } from './BlockLink'
+export { BlockLink } from './BlockLink'
```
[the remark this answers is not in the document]
craigpalermo: No new functionality was added or tests deleted, so import changes are the only other explanation.

packages/core/src/FormField/index.ts
```diff
@@ -1,5 +1 @@
-import type { IFormFieldProps } from './FormField'
-import FormField from './FormField'
-
-export type { IFormFieldProps }
-export { FormField, FormField as InputField }
```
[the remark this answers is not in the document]
craigpalermo: Ah, I didn't realize that was there as an alias. I don't know if we discussed deprecating it or not, but I can put it back if we want to keep it.

packages/core/src/Radio/__snapshots__/Radio.spec.tsx.snap
craigpalermo: I think it might be! I wasn't paying close attention because we're going to address Chromatic diffs on the `DSv6` branch, but I can address this before merging.

packages/core/src/Layout/Layout.tsx
craigpalermo: I had to check, but this should evaluate to `undefined` as it is. I'm not sure I actually _needed_ to add the optional chaining now that I'm looking at it again.
Screenshot 2023-12-26 at 5 04 40 PM

packages/core/src/BlockLink/BlockLink.spec.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 
-import { BlockLink } from '..'
+import { BlockLink } from './BlockLink'
```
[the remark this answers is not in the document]
craigpalermo: I think we're trying to avoid imports from barrel files, so we may need to figure out a strategy for handling coverage reductions that's more targeted than excluding all `index.*` from the report.

dansternfeld1 (approved):
Thanks for the updates, @craigpalermo 
Let's figure out what's the best way to cover things with the barrels

packages/core/src/BlockLink/BlockLink.spec.tsx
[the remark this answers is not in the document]
sdalonzo: i think `src/**/index.ts` or similar should be fine. after this PR, all they do is export the component and the type

packages/core/src/Divider/Divider.stories.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
-import { Divider, Flex } from '..'
 import styled from 'styled-components'
+import { Divider, Flex } from '..'
```
sdalonzo, line 3: n-b missed one

packages/core/src/Label/Label.tsx
```diff
@@ -43,18 +35,7 @@ const accessiblyHide = (props) =>
...
-  ...propTypes.fontWeight,
-  ...propTypes.lineHeight,
-  ...propTypes.textStyle,
-  ...propTypes.width,
-  color: deprecatedColorValue(),
-  textStyle: PropTypes.oneOf(textStylesValues),
-}
-
-export interface ILabelProps
+export interface LabelProps
```
sdalonzo, line 38: n-b type for consistency

packages/core/src/ChatMessageContainer/ChatMessageContainer.stories.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Box } from '../Box'
 import type { StoryObj } from '@storybook/react'
-import ChatMessageContainer, { IChatMessageContainer } from './ChatMessageContainer'
+import { ChatMessageContainer, type ChatMessageContainerProps } from './ChatMessageContainer'
```
pcln-james, line 4: TIL

packages/core/src/PasswordInput/PasswordInput.tsx
```diff
@@ -49,6 +55,9 @@ const PasswordInput: React.FC<InferProps<typeof propTypes>> = ({
 
     if (hasProgressBar) {
       regexChecks.reduce(
+        // TODO TypeScript doesn't like the return type in the reducer, so there may
+        // be a bug to fix here.
+        // @ts-ignore
```
sdalonzo, line 60: n-b: easy fix here 

make the function body, and return acc

craigpalermo: We weren't using it like a `reduce` anyway, changed to `forEach`

sdalonzo (approved):

PR #1436, opened 2024-01-16
Add OverflowX & OverflowY Props to the dialog component

Bsoong:
Ticket: https://priceline.atlassian.net/browse/SITE-22333

As per the discussion from DS guild on Friday this PR adds an overflowX & overflowY Control to the dialog while also not losing functionality.  

I changed the overflow on one of the containers in devtools within NL to confirm the change that was needed. Screenshot here: 
Screenshot 2024-01-16 at 3 35 46 PM

Attached is the dialog without devtools changes using controls in storybook: DEFAULT
Screenshot 2024-01-16 at 3 43 33 PM

OverflowX & Y set to visible: 
Screenshot 2024-01-16 at 3 46 46 PM

One unintended addition of adding overflowX is that the scrollbar when overflowX="visible" is not rounded. 

Unfortunately with the DSV6 upgrade we will not be able to implement this change at this exact moment - but once Next-landing is migrated we will pick the effort back up.

### DIFF
--- a/packages/core/src/Dialog/Dialog.stories.args.tsx
+++ b/packages/core/src/Dialog/Dialog.stories.args.tsx
@@ -5,7 +5,7 @@ import { Grid } from '../Grid'
 import { Text } from '../Text'
 import { borderRadii, colorSchemeNames, paletteColors } from '../theme'
 import type { DialogProps } from './Dialog'
-import { dialogSizes } from './Dialog.styled'
+import { dialogSizes, overflow } from './Dialog.styled'
 
 export const argTypes: Partial<ArgTypes<DialogProps>> = {
   borderRadius: { control: { type: 'select' }, options: Object.keys(borderRadii) },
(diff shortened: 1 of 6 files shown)


packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -103,7 +104,8 @@ const DialogContentWrapper = styled(motion.div)`
 
 const DialogInnerContentWrapper = styled.div`
   position: relative;
-  overflow: auto;
+  overflow-x: ${(props) => props.overflowX && props.overflowX};
```
sdalonzo, line 107: this is awkward and unnecessary, since you there's a default prop

sdalonzo (approved):

packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -103,7 +104,8 @@ const DialogContentWrapper = styled(motion.div)`
 
 const DialogInnerContentWrapper = styled.div`
   position: relative;
-  overflow: auto;
+  overflow-x: ${(props) => props.overflowX};
+  overflow-y: ${(props) => props.overflowY};
```
craigpalermo, line 108: Aren't these supported by styled-system?

packages/core/src/Dialog/Dialog.stories.args.tsx
```diff
@@ -7,6 +7,7 @@ import { borderRadii, colorSchemeNames, paletteColors } from '../theme'
 import type { DialogProps } from './Dialog'
 import { dialogSizes } from './Dialog.styled'
 
+const overflowArgs = ['scroll', 'visible', 'hidden', 'auto']
```
craigpalermo, line 10: n-b I think `clip` is missing

PR #1440, opened 2024-01-24
remove lodash

sdalonzo:


packages/codemods/transforms/rename-deprecated-props-flex.js
```diff
@@ -1,10 +1,8 @@
-import get from 'lodash.get'
-
 const update = (path) => {
   const Node = path.value
 
   // Dumb way to skip text elements since they also can have an align prop
-  if (get(Node, 'openingElement.name.name', '').toLowerCase().includes('text')) {
+  if (Node?.openingElement?.name?.name?.toLowerCase()?.includes('text')) {
```
dansternfeld1, line 5: (super-n-b) If that last `name` returns, `toLowerCase()` should work, no?
```suggestion
  if (Node?.openingElement?.name?.name?.toLowerCase().includes('text')) {
```

packages/core/src/DocsUtils/Section/Section.tsx
```diff
@@ -1,7 +1,7 @@
-import kebabCase from 'lodash/kebabCase'
 import React from 'react'
-
-import { Box, Text } from '../..'
```
dansternfeld1, line 4: (n-b) More barrels to clear out?

packages/core/src/stories/Colors.stories.tsx
```diff
@@ -77,7 +77,7 @@ export const Palette = () => (
                 const shades = theme.palette[color]
                 if (typeof shades === 'object') {
                   return (
-                    <div style={{ width: '100%' }}>
+                    <div key={color} style={{ width: '100%' }}>
```
dansternfeld1, line 80: (n-b) 👍

packages/core/src/utils/strings.ts
```diff
@@ -0,0 +1,10 @@
+export function kebabCase(string: string) {
+  return string
+    .replace(/([a-z])([A-Z])/g, '$1-$2')
+    .replace(/[\s_]+/g, '-')
+    .toLowerCase()
```
dansternfeld1, line 5: This function doesn't return exactly the same results as `lodash`'s version, but I'm not sure how true to the original we need to be. The most obvious difference is that hyphens can end up leading or trailing characters -- is it worth an enhancement to trim that?
Screenshot 2024-01-24 at 7 15 24 AM
Screenshot 2024-01-24 at 7 16 10 AM

packages/carousel/src/helpers.js
```diff
@@ -1,14 +1,14 @@
...
 import {
-  VISIBLE_SLIDES_BREAKPOINT_1,
-  VISIBLE_SLIDES_BREAKPOINT_2,
   CAROUSEL_BREAKPOINT_1,
   CAROUSEL_BREAKPOINT_2,
   MEDIA_QUERY_MATCH,
+  VISIBLE_SLIDES_BREAKPOINT_1,
+  VISIBLE_SLIDES_BREAKPOINT_2,
 } from './constants'
-import debounce from 'lodash.debounce'
```
dansternfeld1, line 11: (n-b) Kind of amazing that this was the only usage in DS

packages/core/src/utils/utils.ts
```diff
@@ -432,3 +432,18 @@ export const colorSchemeCustomForeground = ({ colorScheme, color, iconUsesColorS
 export const textTransform = (props) => (props.textTransform ? { textTransform: props.textTransform } : null)
 
 export const textWrap = (props) => (props.textWrap ? { textWrap: props.textWrap } : null)
+
+export function debounce(func, wait, immediate) {
+  let timeout
+  return function () {
+    const context = this
```
dansternfeld1, line 439: (n-b) At least you didn't call it `that` 🤣

packages/core/src/utils/utils.ts
```diff
@@ -432,3 +432,18 @@ export const colorSchemeCustomForeground = ({ colorScheme, color, iconUsesColorS
 export const textTransform = (props) => (props.textTransform ? { textTransform: props.textTransform } : null)
 
 export const textWrap = (props) => (props.textWrap ? { textWrap: props.textWrap } : null)
+
+export function debounce(func, wait, immediate) {
+  let timeout
+  return function () {
```
dansternfeld1, line 438: Don't you need to spread the arguments into this function? I had assumed that this creates a new scope for the `args` setting below and comes up empty.

packages/icons/src/Icon.jsx
```diff
@@ -1,9 +1,12 @@
+import PropTypes from 'prop-types'
 import React from 'react'
 import styled from 'styled-components'
-import upperFirst from 'lodash.upperfirst'
-import PropTypes from 'prop-types'
 import * as icons from './index'
 
+function upperFirst(string) {
```
dansternfeld1, line 6: (n-b) You've written this function twice -- the other instance is in `utils`
It's so small that repetition is probably fine, but do you want to consolidate?

rush.json
```diff
@@ -21,7 +21,7 @@
    * Specify a SemVer range to ensure developers use a Node.js version that is appropriate
    * for your repo.
    */
-  "nodeSupportedVersionRange": ">=16",
+  "nodeSupportedVersionRange": ">=18",
```
dansternfeld1, line 24: (n-b) This get missed in the v6 pull?

packages/popover/package.json
```diff
@@ -33,7 +33,6 @@
     "@babel/core": "^7.22.15",
     "@priceline/babel-preset": "workspace:*",
     "@priceline/eslint-config": "workspace:*",
-    "@reach/component-component": "^0.16.0",
```
dansternfeld1, line 36: (n-b) Oh, missed this one -- unused?

packages/core/src/DocsUtils/Section/Section.tsx
sdalonzo: yea i left DocsUtils alone the last time, still think it might be a separate package soon

packages/core/src/utils/strings.ts
sdalonzo: good catch! it's a naive copy from the internet, first goal was to get it building and measure the bundle change

PR #1441, opened 2024-01-24
fix(AIR-60631): update accordion shadow state for open

Andrewleel:
When card or flatCard variant  closed no changes are needed, when open both would have no shadow and use the border of 1px and color border.base.

### DIFF
--- a/packages/core/src/Accordion/Accordion.stories.tsx
+++ b/packages/core/src/Accordion/Accordion.stories.tsx
@@ -101,7 +101,9 @@ export const Ladder = {
   ),
 }
 
-export const Card = {
+export const Card = { render: (args) => <Accordion {...args} items={items} variation='card' /> }
+
+export const DefaultStateCard = {
   render: (args) => <Accordion {...args} items={items} variation='card' itemsState={['item-1', 'item-2']} />,
 }
 
--- a/packages/core/src/Accordion/Accordion.stories.tsx
+++ b/packages/core/src/Accordion/Accordion.stories.tsx
@@ -103,7 +103,7 @@ export const Ladder = {
 
 export const Card = { render: (args) => <Accordion {...args} items={items} variation='card' /> }
 
-export const DefaultStateCard = {
+export const ItemStatePropCard = {
   render: (args) => <Accordion {...args} items={items} variation='card' itemsState={['item-1', 'item-2']} />,
 }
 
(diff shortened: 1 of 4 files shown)


packages/core/src/Accordion/Accordion.styled.tsx
```diff
@@ -148,14 +147,18 @@ export const StyledItem = styled(Box)<IStyledItem>`
...
+        : ''}
   }
   &[data-state='closed'] {
     ${(props) =>
       props.variation === 'underline'
         ? `border-bottom-left-radius: 0px;
     border-bottom-right-radius: 0px;`
         : ''}
+    box-shadow: ${(props) =>
+      props.variation === 'card' || props.variation === 'flatCard' ? themeGet('shadows.sm') : ''};
```
ryan-hunt-priceline, line 162: NB: might be cleaner to use `['card', 'flatCard'].includes(props.variation)`

packages/core/src/Accordion/Accordion.styled.tsx
```diff
@@ -148,14 +147,18 @@ export const StyledItem = styled(Box)<IStyledItem>`
       : ''}
   &[data-state='open'],
   &:hover {
-    box-shadow: ${(props) =>
-      props.variation === 'card' || props.variation === 'flatCard' ? themeGet('shadows.xl') : ''};
+    ${(props) =>
+      props.variation === 'card' || props.variation === 'flatCard'
+        ? `border: solid 1px ${getPaletteColor('border.base')(props)}`
```
sdalonzo, line 152: don't forget the trailing `;` since we're writing a css literal

packages/core/src/Accordion/Accordion.stories.tsx
```diff
@@ -100,7 +100,10 @@ export const Ladder = {
     />
   ),
 }
-export const Card = { render: (args) => <Accordion {...args} items={items} variation='card' /> }
+
+export const Card = {
+  render: (args) => <Accordion {...args} items={items} variation='card' itemsState={['item-1', 'item-2']} />,
+}
```
craigpalermo, line 106: Could you make a new story for when the `itemState` prop is present?

Andrewleel: Can you clarify on that?

craigpalermo: It would be helpful to keep the existing `Card` story, but then create an additional story that uses the `itemsState` prop

sdalonzo (approved):

PR #1451, opened 2024-02-08
Reimplement Layout component using CSS Grid

craigpalermo:


huzefakagdi:
SiteX uses `Layout` extensively. Are we sure this update will not break any existing usage?

sdalonzo (approved):

craigpalermo:
> SiteX uses `Layout` extensively. Are we sure this update will not break any existing usage?

@huzefakagdi About as sure as we can be based on manual testing in Storybook and comparing against the old stories in Chromatic, but this should be verified in apps that use it like any other change.

PR #1453, opened 2024-02-09
change color of base orange to dark

pherrera167:
Change the color orange to color of orange dark


packages/core/src/theme/colors.ts
```diff
@@ -29,7 +29,7 @@ const shadeRed = '#6d0000'
 // orange
 const lightOrange = '#fef2e7'
 const tintOrange = '#faaf69'
-const orange = '#f68013'
+const orange = '#f06f20'
```
pherrera167, line 32: This is the only change. Changing the color of orange. Everything else is snapshots 📸

Bsoong (approved):

sdalonzo (approved):

PR #1455, opened 2024-02-12
Feat fab provider

ryan-hunt-priceline:
Create new `FloatingActionButtonProvider` component. This can be used to position, delay rendering by time, or hide until user has scrolled by a certain percentage. Prop options are

-   **floatingActionButtons**: Required Array taking in objects of FloatingActionButton props.
-  **enterAnimation**: String to specify Animate variation to use on render
- **exitAnimation**: String to specify Animate variation to use on exit
- **position**: String to specify where to place the component (Currently **topRight**, **centerRight**, **bottomRight**)
-  **delayDisplayMs**: Optional number that will delay render by milliseconds passed
-   **hideUntilScrollPercent**:  Optional number that will hide the FAB button until user scrolls down by percentage passed here

Also updated the `FloatingActionButton` component to add in new prop `isAbsolutePosition` to allow wrapping in a **Box** instead of an **Absolute**. This will default to `true` to maintain existing position

Also added in 4 new Animate options:
- slideInRight
- slideOutRight
- slideInBottom
- slideOutBottom


packages/core/config/jest.config.json
```diff
@@ -2,10 +2,10 @@
   "extends": "@priceline/heft-component-rig/profiles/react/config/jest.config.json",
   "coverageThreshold": {
     "global": {
-      "statements": 96,
-      "branches": 95,
-      "functions": 90,
-      "lines": 96
+      "statements": 95,
```
ryan-hunt-priceline, line 5: Used Chromatic for interactive testing. Had to lower the coverage for this

packages/core/src/FloatingActionButtonProvider/useFloatingActionButton.ts
```diff
@@ -0,0 +1,60 @@
...
+  const handleScroll = async () =>
+    getScrollPercentage() > hideUntilScrollPercent
+      ? enterAnimation()
+      : exitAnimation()
+          .then(() => {
+            //Do Nothing
+          })
+          .catch(() => {
+            //Do nothing
+          })
```
pcln-james, line 30: Can this be an `await` with `try/catch` instead?

packages/core/src/FloatingActionButtonProvider/useFloatingActionButton.ts
```diff
@@ -0,0 +1,60 @@
...
+  useEffect(() => typeof window !== 'undefined' && hideUntilScrollPercent && addScrollListener(handleScroll))
+
+  useEffect(() => {
+    delayDisplay()
+      .then(() => {
+        //Do nothing
+      })
+      .catch(() => {
+        //Do nothing
+      })
```
pcln-james, line 50: if these don't do anything do we need them?

pcln-james (commented):
one small concern with the empty `.then.catch`

packages/core/src/FloatingActionButtonProvider/useFloatingActionButton.ts
ryan-hunt-priceline: @James300, I was getting linting errors when I had await, until I added these

ryan-hunt-priceline: Code changed after using AnimatePresence

craigpalermo (approved):

ryan-hunt-priceline:
> I'm wondering if there should be any conditional style updates to the tooltip so that it doesn't get cut off depending on where the button is: Screenshot 2024-02-21 at 1 26 29 PM

Just updated to set tooltips to right align so that it will fit. Good callout :)

PR #1456, opened 2024-02-13
Update Dialog close button and header

craigpalermo:
* Use `CloseButton` in Dialog
* Change default Dialog header color to `heading`

### DIFF
--- a/packages/core/src/Toast/Toast.spec.tsx
+++ b/packages/core/src/Toast/Toast.spec.tsx
@@ -36,7 +36,7 @@ describe('Toast', () => {
     jest.runAllTimers()
     expect(mockRemoveClick).toHaveBeenCalledTimes(0)
 
-    fireEvent.click(screen.getAllByTitle('close-toast')[0])
+    fireEvent.click(screen.getByRole('button'))
     expect(mockRemoveClick).toHaveBeenCalledTimes(1)
     expect(mockRemoveClick).toHaveBeenCalledWith(10)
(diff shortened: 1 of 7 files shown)


packages/core/src/Toast/Toast.spec.tsx
```diff
@@ -36,7 +36,7 @@ describe('Toast', () => {
     jest.runAllTimers()
     expect(mockRemoveClick).toHaveBeenCalledTimes(0)
 
-    fireEvent.click(screen.getAllByTitle('close-toast')[0])
+    fireEvent.click(screen.getAllByTestId('close-toast')[0])
```
huzefakagdi, line 39: `getByRole('button')` should be preferred here. Also, should an `aria-label` be added to the close button? to improve accessibility?

huzefakagdi: NB
Also, one more recommendation from react-testing-library is to use the `user-event` package over `fireEvent`, where the corresponding actions are available
https://testing-library.com/docs/user-event/intro/#differences-from-fireevent

craigpalermo: I updated the tests to look for the button by role. I agree with the suggestion about adding `aria-label`, but it's a bit out of scope for this change. We can ticket that separately.

sdalonzo (approved):

Bsoong (approved):

PR #1458, opened 2024-02-15
Carousel and SlideBox: add in new props to allow overflow on x, y, an…

ryan-hunt-priceline:
…d top

### DIFF
--- a/packages/core/src/SlideBox/SlideBox.tsx
+++ b/packages/core/src/SlideBox/SlideBox.tsx
@@ -24,6 +24,9 @@ export type SlideBoxProps = {
   arrowPosition?: ArrowPosition
   slideScrollNum?: number
   mobileSlideScrollNum?: number
+  overflowAllowanceX?: number
+  overflowAllowanceY?: number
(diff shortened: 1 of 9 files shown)


packages/core/src/SlideBox/SlideBox.tsx
```diff
@@ -24,6 +24,9 @@ export type SlideBoxProps = {
   arrowPosition?: ArrowPosition
   slideScrollNum?: number
   mobileSlideScrollNum?: number
+  overflowAllowanceX?: number
+  overflowAllowanceY?: number
+  overflowAllowanceTop?: number
```
craigpalermo, line 29: n-b would including `px` in these prop names clarify the type of value expected?

craigpalermo (approved):

packages/core/src/SlideBox/SlideBox.tsx
ryan-hunt-priceline: Updated prop names to include "px"

PR #1463, opened 2024-02-27
feat: update DS components to match unified search

JordonHong:
* Button updates
  * Focus styles trigger on `:focus-visible` and disabled colour to use `background.tone`
* Select updates
  * Truncates long options
  * Add subtle variation based on Button
* Dialog updates
  * Add `headerContentProps` to allow changes to spacing and grid styles
* Changelog

## Screenshots

https://github.com/priceline/design-system/assets/156821810/efbd3d6f-91ab-4fa7-bad5-505d844fe094

### DIFF
--- a/packages/core/src/Dialog/Dialog.tsx
+++ b/packages/core/src/Dialog/Dialog.tsx
@@ -1,7 +1,7 @@
 import * as Dialog from '@radix-ui/react-dialog'
 import { AnimatePresence } from 'framer-motion'
 import React, { useEffect } from 'react'
-import { OverflowProps, SpaceProps } from 'styled-system'
+import { OverflowProps } from 'styled-system'
 import type { BorderRadius, ColorSchemes, PaletteColor, ZIndex } from '../theme/theme'
 import type { DialogSize } from './Dialog.styled'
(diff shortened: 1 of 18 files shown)


packages/core/src/Dialog/Dialog.tsx
```diff
@@ -19,6 +20,7 @@ export type DialogProps = Omit<OverflowProps, 'overflow'> & {
   fullWidth?: boolean
   headerColorScheme?: keyof ColorSchemes
   headerContent?: string | React.ReactNode
+  headerContentProps?: GridProps & SpaceProps
```
pcln-james, line 23: If you remove `SpaceProps` does it still work? I believe GridProps already supports SpaceProps from one of the types it unions with

pcln-james (approved):
LGTM if Wes is happy with all of the changes, great work!

packages/core/src/Dialog/Dialog.tsx
JordonHong: It does! I'll remove the import.

craigpalermo (approved):

PR #1471, opened 2024-03-06
DialogInnerContentWrapper full height if footerContent exists

pcln-james:
* only full height Dialog content if footer content exists
* only overflow scroll content if footer content exists

### DIFF
--- a/packages/core/src/Dialog/Dialog.styled.tsx
+++ b/packages/core/src/Dialog/Dialog.styled.tsx
@@ -106,7 +106,8 @@ const DialogInnerContentWrapper = styled.div`
   display: grid;
   grid-template-columns: 1fr;
   grid-template-rows: auto 1fr auto;
-  height: 100%;
+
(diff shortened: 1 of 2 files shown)


packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -213,6 +214,7 @@ export const DialogContent = ({
           overflowX={overflowX}
           overflowY={overflowY}
           sheet={sheet}
+          hasFooterContent={footerContent}
```
ryan-hunt-priceline, line 217: Just need to add a prop type for `hasFooterContent`

ryan-hunt-priceline (approved):

sdalonzo (approved):

packages/core/src/Dialog/Dialog.styled.tsx
pcln-james: I just checked and this is an internal only prop and TS is not throwing warnings/errors on build/lint

PR #1472, opened 2024-03-08
fix(AIR-61584): avoid render the hidden children

CalebBoLiuPriceline:
The carouse would render the children twice, which caused the same children component with same test-id appear twice. 

Before fix
Screenshot 2024-03-11 at 10 44 06 AM

After fix
Screenshot 2024-03-11 at 10 45 23 AM

### DIFF
--- a/packages/carousel/src/RenderInView/RenderInView.jsx
+++ b/packages/carousel/src/RenderInView/RenderInView.jsx
@@ -8,7 +8,7 @@ const FullHeightInView = styled(InView)`
   height: 100%;
 `
 
-const RenderInView = ({ children, onSlideChange, index }) => {
+const RenderInView = ({ children, onSlideChange, index, displayArrowsMobile }) => {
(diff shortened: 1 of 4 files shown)


ryan-hunt-priceline (approved):

sdalonzo (approved):

packages/carousel/src/RenderInView/RenderInView.jsx
```diff
@@ -8,17 +8,19 @@ const FullHeightInView = styled(InView)`
...
     if (inView) {
       onSlideChange(index)
     }
   }
   return (
     <>
-      <Hide md lg xl xxl height='100%'>
-        <FullHeightInView onChange={slideVisible}>{children}</FullHeightInView>
-      </Hide>
+      {displayArrowsMobile === true ? (
```
sdalonzo, line 19: i don't think overloading this prop is the correct approach, for the record.

PR #1474, opened 2024-03-15
feat(RC-25645): Disable floating-ui for popover using a prop.

LKeithLopez:
# Disable floating-ui for Popover component using a prop.
By default the Popover component implements floating. I have a case where this is not desired. Rather than recreate the component without it I found that it is possible to disable the floating using `floating-ui` props:  `crossAxis` and `mainAxis`. 

When both of those are set to false no floating occurs. The component remains "sticky" to its original `placement`

I implemented this with a defaultProp that doesn't change the existing implementation. To disable it you must pass the new prop.

disableFloating-Popover-prop

### DIFF
--- a/packages/popover/src/Popover/Popover.stories.jsx
+++ b/packages/popover/src/Popover/Popover.stories.jsx
@@ -202,6 +202,24 @@ const SimpleTextContent = () => (
   </Box>
 )
 
+const LongTextContent = () => (
+  <Box p={2}>
+    <Text textStyle='heading2'> Disable Floating: </Text>
+    <Text textStyle='heading5'>
+      {' '}
+      When the prop disableFloating is set to true, the popover will not float on the screen. Resize your
+      window/view-port to see the effect.{' '}
+    </Text>
+    <Text textAlign='center'>
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et
+      dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip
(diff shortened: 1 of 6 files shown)


jligi (approved):

packages/popover/src/Popover/Popover.stories.jsx
```diff
@@ -334,3 +352,22 @@ export const RendersOnSpecificArea = {
...
+        width={330}
+        disableFloating={args.disableFloating}
+      >
+        <Button color='primary' mx={2}>
+          Open popover
+        </Button>
+      </Popover>
+    </Flex>
+  )
+}
```
craigpalermo, line 373: @LKeithLopez Mind adding an arg so the story will default to `disableFloating`? Here's an example https://github.com/priceline/design-system/blob/main/packages/carousel/src/Carousel.stories.jsx#L119

packages/popover/src/Popover/Popover.stories.jsx
```diff
@@ -334,3 +352,22 @@ export const RendersOnSpecificArea = {
...
+export const DisableFloating = (args) => {
+  return (
+    <Flex m={200}>
+      <Popover
+        color='primary'
+        renderContent={LongTextContent}
+        placement='bottom'
+        ariaLabel='Default Popover'
+        width={330}
+        disableFloating={args.disableFloating}
```
craigpalermo, line 365: Can you spread the `args` as props on `<Popover>`? That's how Storybook forwards the args to the component in the story.

LKeithLopez: Will do!

craigpalermo (approved):

jligi (approved):

sdalonzo (approved):

PR #1477, opened 2024-03-28
Site 22740 - Fix to strange behavior for next-landing

Bsoong:
Ticket: https://priceline.atlassian.net/browse/SITE-22740

While attempting to bump design system up in NL, the dialog behavior for our express deal was showing the overflow issue again. this is the code to fix said issue via a default true set prop that we will set false as to not disrupt others that rely on the `overflowY: scroll`.

### DIFF
--- a/packages/core/src/Dialog/Dialog.styled.tsx
+++ b/packages/core/src/Dialog/Dialog.styled.tsx
@@ -173,6 +173,7 @@ export const DialogContent = ({
   overflowX,
   overflowY,
   showScrollShadow,
+  innerContentScroll,
 }: DialogProps) => {
   const headerSizeArray = [
     headerIcon ? 'heading5' : 'heading4', // xs
@@ -254,7 +255,7 @@ export const DialogContent = ({
               )}
             </SmoothTransitionBox>
           ) : (
-            <Box height='100%' style={{ overflowY: 'scroll' }}>
+            <Box height='100%' style={innerContentScroll && { overflowY: 'scroll' }}>
               {children}
             </Box>
           )}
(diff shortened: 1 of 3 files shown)


packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -254,7 +255,7 @@ export const DialogContent = ({
               )}
             </SmoothTransitionBox>
           ) : (
-            <Box height='100%' style={{ overflowY: 'scroll' }}>
+            <Box height='100%' style={innerContentScroll && { overflowY: 'scroll' }}>
```
huzefakagdi, line 258: Using `&&` while assigning prop values is not the best solution. In this context if `innerContentScroll` is falsy, then `style` would be set to `false`. You can easily see that error in VS Code as `false` is not a valid value for `style` 
image

huzefakagdi (changes requested):

packages/core/src/Dialog/Dialog.styled.tsx
Bsoong: My editor did not throw that error for me 🤔, pushing a fix for that now.

sdalonzo (approved):

packages/core/src/Dialog/Dialog.styled.tsx
```diff
@@ -254,7 +255,7 @@ export const DialogContent = ({
               )}
             </SmoothTransitionBox>
           ) : (
-            <Box height='100%' style={{ overflowY: 'scroll' }}>
+            <Box height='100%' style={innerContentScroll ? { overflowY: 'scroll' } : {}}>
```
huzefakagdi, line 258:
```suggestion
            <Box height='100%' style={innerContentScroll ? { overflowY: 'scroll' } : undefined}>
```

huzefakagdi: There is a prop `showScrollShadow`. When the value for it is `true`, this `Box` is not shown. Instead, the `SmoothTransitionBox` is shown above. It seems that will have the same problem. So we need to tackle this on all fronts

PR #1490, opened 2024-04-29
Rename innerContentScroll prop to innerContentOverflow

pcln-james:
I need similar behaviour to what is being done with the `innerContentScroll` prop but I need to set the overflow to `hidden`. I don't want to create a whole new prop just to control this behaviour in a slightly different way so I thought renaming the prop and setting the changing the default value from `true` to `scroll` should help with backwards compatibility. This obviously isn't perfect, if anyone is setting this prop to false their code will be hurt by this but as this is the default functionality and a new prop anyway I think this is safe but I am happy to change the version bump if people disagree.

### DIFF
--- a/common/changes/pcln-design-system/rename-innerContentOverflow-prop_2024-04-29-21-42.json
+++ b/common/changes/pcln-design-system/rename-innerContentOverflow-prop_2024-04-29-21-42.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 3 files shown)


sdalonzo (changes requested):
removing a prop that we know is in use has to be a major or coordinated with all the teams using it

BeniCheni (approved):

sdalonzo (approved):

PR #1491, opened 2024-05-12
feat(UXPT-7188): accordion add custom color for header content and hover

apatidar1986:


packages/core/src/Accordion/Accordion.styled.tsx
```diff
@@ -41,7 +41,9 @@ export const StyledContent = styled(Accordion.Content)`
   overflow: hidden;
   ${space}
   background-color: ${(props) =>
-    props.variation === 'default' ? getPaletteColor('background.light')(props) : '#fff'};
+    props.variation === 'default'
+      ? getPaletteColor(props.contentBg || 'background.light')(props)
+      : getPaletteColor(props.contentBg || 'background.lightest')(props)};
```
PhilWelsh, line 46: why is the non default fallback different?

PhilWelsh: oh, that's #fff, got it

PhilWelsh (approved):

PR #1504, opened 2024-07-17
Refactor Popover using Radix UI

craigpalermo:
This adds a new Popover to the core package with a subset of the props on `pcln-popover`.

CleanShot 2024-07-17 at 08 58 03

### DIFF
--- a/packages/core/src/Popover/Overlay.tsx
+++ b/packages/core/src/Popover/Overlay.tsx
@@ -1,0 +1,33 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Box } from '../Box/Box'
+
+const StyledOverlay = styled(Box)`
+  position: fixed;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  opacity: ${({ overlayOpacity }) => overlayOpacity};
+  cursor: pointer;
+  display: ${({ popoverOpen }) => (popoverOpen ? 'block' : 'none')};
+`
+
+export type OverlayProps = {
+  overlayOpacity?: number
+  zIndex?: number
+}
+
+const Overlay = ({ overlayOpacity, isOpen }) => {
+  return (
+    <StyledOverlay
+      aria-hidden='true'
+      tabIndex='-1'
+      popoverOpen={isOpen}
+      overlayOpacity={overlayOpacity}
+      color='background.darkest'
+    />
+  )
+}
+
+export default Overlay
--- a/packages/core/src/Popover/Popover.stories.tsx
+++ b/packages/core/src/Popover/Popover.stories.tsx
@@ -1,0 +1,69 @@
+import { Button, Flex, Text, ThemeProvider } from '..'
+import React from 'react'
+import { Popover } from './Popover'
+import { argTypes, defaultArgs } from './Popover.stories.args'
+import { action } from '@storybook/addon-actions'
+
+export default {
+  title: 'Popover',
+  component: Popover,
+  args: defaultArgs,
+  argTypes: argTypes,
+}
+
+const TemplateContent = ({ ...args }) => (
+  <Flex justifyContent='center' alignItems='center' width={1} my={6}>
+    <Popover
+      idx='basic-popover'
+      onOpen={action('onOpen')}
+      onClose={action('onClose')}
+      renderContent={() => (
+        <Flex flexDirection='column' p={2} onClick={(e) => e.stopPropagation()}>
+          <Text fontSize={1} mb={1} mt={0} fontWeight='bold'>
+            Tip:
+          </Text>
+          <Text mt={0} mb={1} fontSize={0}>
+            Use arrow keys to navigate the calendar
+          </Text>
+        </Flex>
+      )}
+      overlayOpacity={0.3}
+      width={350}
+      p={2}
+      {...args}
+    >
+      <Button className='IconButton' aria-label='Sample popover'>
+        Open
+      </Button>
+    </Popover>
+  </Flex>
+)
+
+const BaseTemplate = ({ children, theme, ...args }) => {
+  return <TemplateContent {...args} />
+}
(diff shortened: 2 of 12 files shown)


packages/core/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+// Currently broken in plcn-popover:
+// borderColor: PropTypes.string,
+
+// Is this still needed?
+// display: PropTypes.string,
+// zIndex: PropTypes.number, <-- handled automatically by radix
+// trapFocus: PropTypes.bool, <-- fully managed by radix
+// stopPropagation: PropTypes.bool,
+// disableFloating: PropTypes.bool, <-- this controlled 'flip' in the old useFloating hook
```
pcln-james, line 157: Is this all needed still?

craigpalermo: I think `disableFloating` may still have some utility, but I'm unsure about the rest.

packages/core/src/Popover/Popover.stories.tsx
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable react/prop-types */
```
sdalonzo, line 1: we should just remove it from the whole lint ruleset imo

packages/core/src/Popover/Popover.tsx
sdalonzo: i'd start with none of it, most of it is probably intrinsic.  we can fill in gaps as they come up

packages/core/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,172 @@
...
+        {children}
+      </Popover.Trigger>
+      <Popover.Portal>
+        <Popover.Content sideOffset={5} side={side} align={align} id={idx}>
+          <ThemeProvider>
+            <Box color={color} boxShadowSize='overlay-md' borderRadius={borderRadius} p={p} width={width}>
+              <RenderContent />
+
+              <FixedCloseButton>
+                {/* @ts-ignore */}
```
sdalonzo, line 145: why?

packages/core/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,172 @@
...
+        asChild
+        onClick={handleTriggerClick}
+        onMouseEnter={handleMouseEnter}
+        onMouseLeave={handleMouseLeave}
+      >
+        {children}
+      </Popover.Trigger>
+      <Popover.Portal>
+        <Popover.Content sideOffset={5} side={side} align={align} id={idx}>
+          <ThemeProvider>
```
sdalonzo, line 140: need to accept an optional theme prop, or read it off context. otherwise, this will brand-break

packages/core/src/Popover/Popover.tsx
craigpalermo: I was doing it to get around the lack of an `as` type on `CloseButton,` but it looks like in the few other places we have an `as` prop we just use `undefined.`

tools/eslint-config/index.js
```diff
@@ -30,12 +30,7 @@ module.exports = {
     'progress/activate': 0,
 
     // React
-    'react/prop-types': [
-      'error',
-      {
-        ignore: ['children'],
-      },
-    ],
+    'react/prop-types': 'off',
```
sdalonzo, line 33: 😍

packages/core/src/Popover/Popover.tsx
```diff
@@ -0,0 +1,165 @@
...
+    | 'right-start'
+    | 'right-end'
+}
+
+function parsePlacement(placement: PopoverProps['placement']): {
+  side: 'top' | 'bottom' | 'left' | 'right'
+  align: 'start' | 'end' | 'center'
+} {
+  const match = /([a-z]+)-?([a-z]+)?/.exec(placement)
+  if (match) {
```
sdalonzo, line 49: if you add Array.isArray and a length check, can we drop ts-ignore?

craigpalermo: The issue is with the parts of the string we're trying to extract. There's no guarantee that the `string` that's extracted matches the expected return values.

CleanShot 2024-07-19 at 08 52 49@2x

wesleymartin85:
@craigpalermo two pieces of feedback:

1. can you update the default close button styles to use the fill variation, size sm
2. for mobile screens and smaller breakpoints we need to set max-widths that will respond based on the width of the screen. I dont know if a single rule works like max-width 100% of the vw -32px, because as seen in the attached example when setting for the pop-over are to display to the left it reduces the available vw space. I think it may be a combination of providing a default max width rule for smaller breakpoints plus a story or two that provides guidance on how to set widths accordingly for such a use case. 
image (2)

sdalonzo (approved):

packages/core/src/Popover/Overlay.tsx
```diff
@@ -0,0 +1,33 @@
...
+const StyledOverlay = styled(Box)`
+  position: fixed;
+  top: 0;
+  bottom: 0;
+  left: 0;
+  right: 0;
+  opacity: ${({ overlayOpacity }) => overlayOpacity};
+  cursor: pointer;
+  display: ${({ popoverOpen }) => (popoverOpen ? 'block' : 'none')};
+`
```
pcln-james, line 14: I would prefer if we replaced this with the `Shadow(Overlay|Effect)` component so that it was easier to maintain consistency

pcln-james (approved):
LGTM! just that one comment but it's nb. great work on this!

packages/core/src/Popover/Overlay.tsx
craigpalermo: That'd be a nice future enhancement. I took this from the old popover to be sure the new one starts with the same styles.

PR #1508, opened 2024-07-25
Slidebox key access

cjoshy:
* Slide component was trying to access key from props.
* `key` is a special prop https://react.dev/warnings/special-props  and shouldn't be read from a component.
* This pr removes the read and just keeps the map index as slide key


cjoshy:
Closing this pr. Will revisit this later because index shouldn't be used as keys as well.

PR #1521, opened 2024-10-21
Update: phone call related icons

bobbylkchao:
**Add icons:**
Merge, Microphone, CallReceive, PlayArrow

**Update icons:**
CallEnd, MicOff, PhoneDialPad


packages/icons/src/index.js
```diff
@@ -161,7 +161,9 @@ export { default as Map } from './Map.jsx'
 export { default as MapFolded } from './MapFolded.jsx'
 export { default as Mask } from './Mask.jsx'
 export { default as Menu } from './Menu.jsx'
+export { default as Merge } from './Merge.jsx'
 export { default as MicOff } from './MicOff.jsx'
+export { default as Microphone } from './Microphone.jsx'
```
joingram-priceline, line 166: Can you rename Microphone to MicOn? Then it will match MicOff

bobbylkchao: > Can you rename Microphone to MicOn? Then it will match MicOff

Good question, I originally wanted to name it micOn, but since this is a reusable icons library of design system, and the icon itself actually embodies the meaning of "microphone", which is an object, while "micOn" is our definition in Sales-CCP.

victoriaIsk (approved):

joingram-priceline (approved):

PR #1528, opened 2024-11-05
feat(HTL-111103): Enable custom header for drawer component

coolestKev:
1. Adding a unit test to verify how custom headings behave.
2. Removed redundant `string | React.ReactNode` type declaration.
Screenshot 2024-11-05 at 9 19 20 AM


packages/core/src/Drawer/Drawer.tsx
```diff
@@ -14,6 +14,7 @@ export type PlacementOptions = 'top' | 'bottom' | 'right' | 'left'
 export type DrawerProps = SpaceProps &
   LayoutProps & {
     children?: string | React.ReactNode
+    customHeader?: string | React.ReactNode
```
huzefakagdi, line 17: `React.ReactNode` allows string, no need to add it explicitly in the union. Also, feel free to clean up the types for `children` and `heading` also

packages/core/src/Drawer/Drawer.tsx
```diff
@@ -14,6 +14,7 @@ export type PlacementOptions = 'top' | 'bottom' | 'right' | 'left'
...
     heading?: string | React.ReactNode
```
huzefakagdi, line 18: IMHO, `customHeader` and `heading` both allowing a `ReactNode` is going to create confusion

craigpalermo (approved):

packages/core/src/Drawer/Drawer.styled.tsx
```diff
@@ -67,3 +68,40 @@ export const HeaderButton = styled(IconButton)`
...
+  height: 8px;
+
+  /* Background/Tint */
+  background: #d5dce4;
+  box-shadow: inset 0px 2px 2px rgba(0, 0, 0, 0.08);
+  border-radius: 999px;
+
+  /* Inside auto layout */
+  flex: none;
+  order: 0;
```
huzefakagdi, line 91: Could you not inline the `DrawerHandle` component in the stories file if it is only used there?

packages/core/src/Drawer/Drawer.tsx
coolestKev: Good point!

packages/core/src/Drawer/Drawer.tsx
```diff
@@ -13,7 +13,7 @@ import { theme } from '../theme'
 export type PlacementOptions = 'top' | 'bottom' | 'right' | 'left'
 export type DrawerProps = SpaceProps &
   LayoutProps & {
-    children?: string | React.ReactNode
+    children?: React.ReactNode
     heading?: string | React.ReactNode
```
huzefakagdi, line 17:
```suggestion
    heading?: React.ReactNode
```

huzefakagdi (approved):

Bsoong (approved):

PR #1536, opened 2024-11-21
Task/htl 111885 Drawer enhancements

coolestKev:
Changelog:

### **1. Added Snap functionality**

- The snap functionality works by changing the absolute positioning of a motion.div within a relative parent container. 
- On each drag action we update the position of the motion.div based on the percentage values we pass in as `snapHeights`. This prop accepts an array of 3 exact values `[top: string, middle: string, bottom: string]` as snap points.
- The extent of scroll and rigidity can be adjusted via `snapDimensions`, which controls the dimensions of the motion.div container.

Video 1: POC with div
https://github.com/user-attachments/assets/e4aa562a-3721-4d25-a833-f74e2d752fd5

Video 2: With Drawer
https://github.com/user-attachments/assets/e15739c8-6e37-4f0c-bb5b-de7c23a5095b

### **2. Added prop to toggle enter animation**

When resizing the viewport, it can trigger an enter animation based on a position change. Sometimes it'll push the component outside the screen. Adding `disableEnterAnimation` disables enter animation and prevents this side-effect, effectively fixating the component to a certain position.

### **3. Refactoring + Added directories for helper function and custom hooks**

Here we will implement a custom hook for the snap use-case to prevent the Drawer.tsx from becoming too verbose.
This way it's modular enough to move between codebases. 
The animation configs and helper functions were also moved into a new directory for better organization.

### DIFF
--- a/packages/core/src/Drawer/Drawer.tsx
+++ b/packages/core/src/Drawer/Drawer.tsx
@@ -28,9 +28,6 @@ export type DrawerProps = SpaceProps &
     placement?: PlacementOptions
     position?: string
 
-    /**
-     * Neighborhoods-related props
-     */
     // top, middle, bottom - 3 snap points required
     snapHeights?: [string, string, string]
 
--- a/packages/core/src/Drawer/Drawer.tsx
+++ b/packages/core/src/Drawer/Drawer.tsx
@@ -7,10 +7,12 @@ import { ChevronDown, Close } from 'pcln-icons'
 import { AnimatePresence, motion } from 'framer-motion'
 import { SpaceProps, LayoutProps } from 'styled-system'
 import { useScrollWithShadow } from '../useScrollWithShadows/useScrollWithShadow'
-import { MotionVariants } from '../Animate/Animate'
 import { theme } from '../theme'
+import { useSnap } from './hooks/useSnap'
+import { getDragToDismissAnimation, getDividerStyle, getEnterAnimation } from './helpers'
 
 export type PlacementOptions = 'top' | 'bottom' | 'right' | 'left'
+
 export type DrawerProps = SpaceProps &
   LayoutProps & {
     children?: React.ReactNode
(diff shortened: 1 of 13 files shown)


packages/core/src/Drawer/helpers/getDragToDismissAnimation.ts
```diff
@@ -0,0 +1,20 @@
...
+  return {
+    drag: 'y',
+    dragConstraints: { top: 0, bottom: 0 },
+    dragElastic: { top: 0, bottom: 0.3 },
+    transition: { duration: 0.8, ease: [0.04, 0.62, 0.23, 0.98] },
+    onDragEnd: (event, info) => {
+      if (info.offset.y > 200) {
+        onDragEnd()
+      }
+    },
```
coolestKev, line 15: Unit tests coming soon

craigpalermo: Storybook interaction tests would be ideal for this

packages/core/src/Drawer/Drawer.tsx
```diff
@@ -67,108 +56,135 @@ export const Drawer: React.FC<DrawerProps> = ({
...
-          width={isMobile ? '100%' : props.width}
-          {...props}
-          height={!isCollapsed && props.height ? props.height : 'fit-content'}
-        >
-          <DrawerRoot
-            data-testid='drawer'
-            isCollapsed={isCollapsed}
-            isFloating={isFloating}
-            key='drawer'
+    <SnapContainer
```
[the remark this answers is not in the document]
coolestKev: It won't affect the existing behavior no, and ideally it shouldn't render (which is why on line 68 I made a toggle) - open to suggestions on other approaches (e.g. is it better to have a null rather than react.fragement?)

sdalonzo (approved):

PR #1539, opened 2024-12-17
Task/htl 112172 additional drawer work

coolestKev:
**Changelog**

1. Fixed a few type declaration bugs + added null checks to remove TS warnings.

2. On first render, assign the starting drag position to be the initial height. This way it enables user to drag bidirectionally (up/down), rather than just up.
https://github.com/user-attachments/assets/f125fa70-14e0-4fd0-9617-862996104e11

3. Propogating maxHeight from container

### DIFF
--- a/packages/core/src/Drawer/Drawer.tsx
+++ b/packages/core/src/Drawer/Drawer.tsx
@@ -29,7 +29,7 @@ export type DrawerProps = SpaceProps &
     position?: string
     zIndex?: number
 
-    // 3 snap points relative to the bottom of the viewport
+    // top, middle (initial position), bottom snap points relative to viewport
     snapHeights?: [string, string, string]
 
     // When snap enabled, lock-in drawer dimensions
--- a/packages/core/src/Drawer/Drawer.tsx
+++ b/packages/core/src/Drawer/Drawer.tsx
@@ -40,9 +40,6 @@ export type DrawerProps = SpaceProps &
 
     // Disable enter animation to eliminate side effects on viewport change
     disableEnterAnimation?: boolean
-
-    // Disable vertical length limit constraint to show variable height on smaller devices
-    disableMaxHeight?: boolean
   }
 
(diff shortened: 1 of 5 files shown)


packages/core/src/Drawer/Drawer.tsx
```diff
@@ -72,21 +78,27 @@ export const Drawer: React.FC<DrawerProps> = ({
...
       data-testid='snap-container'
     >
       <AnimatePresence>
         {isOpen && (
           <DrawerWrapper
             placement={isMobile ? 'bottom' : placement}
             padding={isFloating ? 3 : 0}
-            maxHeight={isMobile ? ['290px', '400px', '480px', 'calc(100vh - 64px)'] : props.height ?? '100%'}
+            maxHeight={
+              isMobile && !disableMaxHeight
```
coolestKev, line 98: On 2nd thoughts, wonder if instead of adding `disableMaxHeight` as a prop, we should add `maxHeight` as a prop instead

packages/core/src/Drawer/Drawer.tsx
```diff
@@ -53,13 +57,15 @@ export const Drawer: React.FC<DrawerProps> = ({
   onClose,
   onCollapse,
   placement = 'right',
+  zIndex = 1,
```
craigpalermo, line 60: Should the default `zIndex` be `auto` since we weren't explicitly setting it before?

packages/core/src/Drawer/Drawer.tsx
craigpalermo: That would make sense since it gives consumers more flexibility.

packages/core/src/Drawer/Drawer.tsx
coolestKev: Actually by not setting zIndex it defaults to auto, hence why the default value

PhilWelsh (approved):
LGTM

PR #1540, opened 2024-12-17
task(HTL-112172): Remove old prop

coolestKev:
Removing an old prop, follow-up to https://github.com/priceline/design-system/pull/1539


packages/core/src/Drawer/Drawer.tsx
```diff
@@ -81,7 +81,6 @@ export const Drawer: React.FC<DrawerProps> = ({
       transition={{ type: 'spring', bounce: 0 }}
       drag='y'
       dragConstraints={{ top: 0, bottom: 0 }}
-      onDragEnd={handleSnap}
```
coolestKev, line 84: Forgot to remove this after making the line 85 change, removed now.

craigpalermo (approved):

wesleymartin85 (approved):

PR #1542, opened 2024-12-20
task(HTL-113265): Pointer fix first commit

coolestKev:
**Problem:**
On mobile devices, touch scrolling can be misinterpreted as a drag action, potentially causing the Drawer container to respond incorrectly to the event.

See video for symptoms attached in previous PR, had to close due to a branch issue: https://github.com/priceline/design-system/pull/1541

**Solution:**
```
    /**
     * Differentiating between content scrolling and container dragging on mobile:
     *
     * 1. Content scrolling:
     *    - Triggered by touch events
     *    - Has a 'pointercancel' event type
     *
     * 2. Container dragging:
     *    - Should not occur accidentally during content scrolling
     *    - Has a 'pointerup' event type
     *
     * To prevent unintended container dragging, we only allow
     * dragging when the event type is specifically 'pointerup'.
     * This ensures smooth content scrolling without accidental
     * container movement.
     */
```

### DIFF
--- a/packages/core/src/Drawer/hooks/useSnap.ts
+++ b/packages/core/src/Drawer/hooks/useSnap.ts
@@ -19,22 +19,43 @@ export function useSnap(snapHeights, snapDimensions) {
   const [snapPosition, setSnapPosition] = useState(MIDDLE)
 
   const handleSnap = (...args) => {
+    const pointerType = args?.[0]?.pointerType // mouse, touch, etc.
+    const type = args?.[0]?.type // click, pointerup (drag), pointercancel (scroll)
     const info = args?.[1]
     const scrollOffset = info.offset.y
 
     const SCROLL_DOWN = scrollOffset > SCROLL_THRESHOLD
     const SCROLL_UP = scrollOffset < -SCROLL_THRESHOLD
 
-    // Scroll down logic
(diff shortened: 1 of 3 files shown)


craigpalermo (approved):

packages/core/src/Drawer/hooks/useSnap.ts
```diff
@@ -19,22 +19,43 @@ export function useSnap(snapHeights, snapDimensions) {
...
+     *    - Should not occur accidentally during content scrolling
+     *    - Has a 'pointerup' event type
+     *
+     * To prevent unintended container dragging, we only allow
+     * dragging when the event type is specifically 'pointerup'.
+     * This ensures smooth content scrolling without accidental
+     * container movement.
+     */
+
+    if (pointerType === 'touch' && type === 'pointerup') {
```
PhilWelsh, line 47: i'm confused, how can it be both, aren't they each specifying a different action touch/mouse ?

PhilWelsh (approved):

packages/core/src/Drawer/hooks/useSnap.ts
coolestKev: Note: Hopped on a call and addressed this

PR #1550, opened 2025-02-20
feat(HTL-114826): add toggleText prop for accordion

tommyliang-dev:
Bringing a previously approved PR back to life: https://github.com/priceline/design-system/pull/1537

- **Accordion**: Added useToggleText prop to allow customization of the toggle element. When set to true, the default arrow will be replaced with "More" and "Less" text

image
image

### DIFF
--- a/packages/core/src/Accordion/Accordion.tsx
+++ b/packages/core/src/Accordion/Accordion.tsx
@@ -3,6 +3,7 @@ import React from 'react'
 import { SpaceProps } from 'styled-system'
 import { Box } from '../Box/Box'
 import { Flex } from '../Flex/Flex'
+import { Text } from '../Text/Text'
 import {
   IconContainer,
   StyledAccordionRoot,
@@ -23,6 +24,7 @@ export type AccordionProps = SpaceProps & {
   variation?: string
   isExternallyControlled?: boolean
   headerDividerColor?: string
+  useTextToggle?: boolean
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/modal/src/__snapshots__/Headers.spec.js.snap
```diff
@@ -24,6 +24,12 @@ exports[`ModalHeader render 1`] = `
   align-items: center;
 }
 
+.c3 {
+  font-weight: 700;
+  font-size: 14px;
+  line-height: 16px;
+}
+
```
tommyliang-dev, line 32: Added from running `rush test:snapshots`

sdalonzo (approved):
code lgtm

packages/core/src/Accordion/Accordion.tsx
```diff
@@ -66,35 +69,48 @@ export function Accordion({
...
+                {useTextToggle ? (
+                  <>
+                    <Text.span textStyle='captionBold' color='primary.base' className='lessText'>
+                      Less
+                    </Text.span>
+                    <Text.span textStyle='captionBold' color='primary.base' className='moreText'>
+                      More
+                    </Text.span>
+                  </>
+                ) : (
```
tommyliang-dev, line 101: Main code change here

tommyliang-dev:
https://priceline.atlassian.net/browse/HTL-115698

Bsoong (approved):

PR #1556, opened 2025-05-28
fix(UXPT-8374): replace defaultProps in svgr template file

Remy-Gratwohl:
Unfortunately missed this the first round. A lot of our shared components rely on the individual svg components' default props.

# Background

Working in our first React 19 application, I noticed some of the SVGs are missing defaults

# Description of Changes

- Replaced some of the old deprecated React defaultProps for the generated SVG components

### DIFF
--- a/packages/icons/svgr.config.js
+++ b/packages/icons/svgr.config.js
@@ -20,6 +20,7 @@ const BaseComponent = ({
   tabIndex = -1,
   focusable = false,
   role = 'img',
+  'aria-hidden': ariaHidden = true,
   ...props
 }) => {
   let ariaLabelledBy = titleId ? titleId : ''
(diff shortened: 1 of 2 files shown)


packages/icons/svgr.config.js
```diff
@@ -35,14 +37,6 @@ const ${componentName} = styled(BaseComponent)\`
 
 ${componentName}.isIcon = true
 
-${componentName}.defaultProps = {
-  size: 24,
-  tabIndex: -1,
-  focusable: false,
-  'aria-hidden': true,
```
Remy-Gratwohl, line 42: As discussed with @sdalonzo it seems the original intention might have been to have 'aria-hidden' set to `!ariaLabelledBy`? However this was always being overridden by either the passed in prop or the `defaultProp` of `true`

PR #1566, opened 2026-02-27
Chore/pts 708 fixing react19 UI issues

takhila499:
- React 19 compatibility — `@floating-ui/react-dom-interactions` is deprecated and it is not compatible with react19.  Replaced `@floating-ui/react-dom-interactions`(deprecated) with `@floating-ui/react` (^0.27.18)

- React 19 support: Updated peer dependencies to include React 19. Also fixed two React 19 breaking change issues where ref was being overridden by prop spreads — in React 19, ref is treated as a regular prop, so it must be placed after any {...props} spread to avoid being overwritten

- Flash-of-misplaced-content fix: Added isPositioned support — popover renders with visibility: hidden until floating-ui calculates its position

- handleToggle correctness: Moved callbacks outside the setOpen updater to keep it a pure function (React best practice)


takhila499:
Closing this PR found a alternative fix instead of touching DS

PR #1568, opened 2026-07-03
Consolidate styled-system onto @soroush.tech/styled-system rewrite via typescript

soroushm:
# Migrate `styled-system` to the maintained, TypeScript-native `@soroush.tech/styled-system` 

## Motivation

The design-system is built on [`styled-system`](https://www.npmjs.com/package/styled-system), which is **no longer maintained** — the last release is `5.1.5` and the project has been inactive for years. Consequences we carry today:

- **No type ownership.** `styled-system` ships no types of its own; we rely on the community‑maintained `@types/styled-system` (equally stale), plus separate `@styled-system/theme-get` and `@styled-system/prop-types` packages. Types and runtime drift independently.
- **No fixes, no performance work.** Bugs and perf are frozen at the 2020 state.
- **Fragmented dependency surface.** Three separate `@styled-system/*` packages to keep in lockstep.

**`@soroush.tech/styled-system`** is a drop-in fork that addresses this:

- **TypeScript‑native** — the library *is* the source of its types (first‑class `.d.cts`/`.d.mts` shipped via `exports`), so types and runtime never drift. No `@types/*` companion needed.
- **Actively maintained** and API‑compatible with `styled-system@5`.
- **Single package** — `theme-get`, `prop-types`, and every category parser are exposed as subpaths of one package instead of three.
- **Faster** on our hot path (see [Benchmark](#benchmark)).

This PR swaps the design-system onto the fork with **zero behavioural change** to components — only the styling engine underneath changes.

## What changed

**Dependencies (9 `package.json` files)**
- Every package now declares **one** styled‑system dependency: `"styled-system": "npm:@soroush.tech/styled-system@^5.6.0"`.
- Removed the separate `@styled-system/theme-get` and `@styled-system/prop-types` dependencies — they're now subpaths of the single fork package.
- Fixed a latent misclassification: `@styled-system/theme-get` (autocomplete) and `@styled-system/prop-types` (slider) were declared as **devDependencies** despite being imported in *shipped* source. They only resolved in production by luck of monorepo hoisting + the peer `pcln-design-system` carrying them. They're now provided by the single runtime `styled-system` dependency.

**Source imports (33 files)**
- `@styled-system/theme-get` → `styled-system/theme-get`
- `@styled-system/prop-types` → `styled-system/prop-types`
- Bare `import … from 'styled-system'` (Box, Flex, Text, …) is unchanged — it already resolves to the fork via the alias.

**No config changes required**
- **No `declare module` shims** and **no `moduleResolution` change.** The fork ships a `typesVersions` map, so its subpath types resolve under the heft rig's classic `node` module resolution as‑is.
- **No consumer type changes.** The fork's public `themeGet`/`Props` signatures match upstream's permissive shapes, so existing call sites (e.g. `themeGet('colorSchemes.x')(props).foreground`, components passing `MotionButtonProps`) type‑check unchanged.

**Lockfile**
- `common/config/rush/pnpm-lock.yaml` regenerated.

_Diff: 41 files, +124 / −231 (net simplification — three deps collapse to one)._

### Verification

`tsc --noEmit` on the two TypeScript‑compiled packages, with the fork installed from npm and no `declare` / no `moduleResolution` change:

| Package | Type errors |
| --- | --- |
| `pcln-design-system` (core) | **0** |
| `pcln-slider` | **0** |

Pre‑commit `rush lint:fix` passes across all projects; `rush prettier` clean.

## Benchmark

### Why and how

The only thing that differs between the old and new design-system builds is the styled‑system engine, so the meaningful comparison is the **styling work the library performs on every render**. `<Box>` is the base primitive nearly every component builds on; on each render it composes **13 styled‑system parsers** and applies them to props:

```
compose(width, display, height, maxHeight, maxWidth, minHeight, minWidth,
        size, space, textAlign, borderRadius, boxShadow, overflow)(props)
```

The benchmark reproduces that exact composition and feeds it a heavy, **fully‑responsive** `<Box>` props object using the design-system's **real theme scales** (5 breakpoints `[32,40,48,64,80]em`, `radii [0,2,6]`, space/sizes/shadows). Array values force each parser to emit per‑breakpoint media queries — the heaviest realistic call the library makes.

- **Tool:** [`@soroush.tech/bench`](https://www.npmjs.com/package/@soroush.tech/bench) (mitata‑based), run from npm.
- **Isolation:** CPU/RAM‑pinned Docker sandbox — 1 CPU (`--cpus 1 --cpuset-cpus 0`), 512 MB with swap off — so numbers are low‑noise and reproducible on the same host.
- **Method:** GC before every iteration, 1000‑iteration warmup, **median of 3 rounds**.
- **Both sides from npm** (installed side‑by‑side under aliases): original `styled-system@5.1.5` vs ours `@soroush.tech/styled-system@5.6.0`.

### Result

| case | avg | p75 | gc/iter | vs fastest |
| :--- | ---: | ---: | ---: | ---: |
| ours — @soroush.tech/styled-system@5.6.0 | 722 µs | 1.54 ms | 5.25 ms | fastest |
| orig — styled-system@5.1.5 (jxnblk) | 851 µs | 1.55 ms | 5.68 ms | +17.9% |

- **~15% faster** on the design-system's styling hot path — the original engine is **+17.9% slower** per parse. Consistent across settled rounds (round 2: 1007 vs 1210 µs; round 3: 722 vs 851 µs).
- **Tail latency tied** — p75 is 1.54 vs 1.55 ms; the win is in the typical case, not just the average.
- **Slightly more memory** — the fork allocates **+5.3%** heap per iteration (60.6 vs 57.5 KB); GC time is marginally lower, so it nets out.

**Net:** meaningfully faster per‑render styling with equal tail latency, at the cost of ~5% more transient allocation per parse.

_Caveats: this measures the styled‑system parsing that dominates render cost (not a full React SSR render), which is the only thing differing between the two builds — so it isolates the real delta. Absolute µs figures are host‑specific; the **ratio** is the portable result._

## Risk

Low. Runtime API is compatible with `styled-system@5`, components are unchanged, and both TypeScript packages type‑check to zero errors with no config workarounds. Rollback is a one‑line revert of the dependency alias.

# Bench source
```js

  // pcln-design-system <Box> styling hot path — ORIGINAL styled-system vs OUR fork.                                                                                                                                                             
  //                                                                                                                                                                                                                                             
  // Box is the base primitive nearly every design-system component builds on. On                                                                                                                                                                
  // every render it composes these 13 styled-system parsers and applies them to                                                                                                                                                                 
  // props. This measures that exact work under the original engine                                                                                                                                                                              
  // (styled-system@5.1.5, jxnblk) vs ours (@soroush.tech/styled-system@5.6.0),                                                                                                                                                                  
  // using pcln-design-system's real theme scales and responsive Box props (array                                                                                                                                                                
  // values force per-breakpoint media-query emission across the DS's 5 breakpoints).                                                                                                                                                            
  //                                                                                                                                                                                                                                             
  //   npx '@soroush.tech/bench' ./design-system-box.bench.ts --md                                                                                                                                                                               
                                                                                                                                                                                                                                                 
  // pcln-design-system theme scales (breakpoints [32,40,48,64,80]em, radii [0,2,6]).                                                                                                                                                            
  const theme = {                                                                                                                                                                                                                                
    breakpoints: ['32em', '40em', '48em', '64em', '80em'],                                                                                                                                                                                       
    space: [0, 4, 8, 16, 32, 64, 128, 256].map((n) => n + 'px'),                                                                                                                                                                                 
    sizes: [16, 32, 64, 128, 256, 512],                                                                                                                                                                                                          
    radii: [0, 2, 6],                                                                                                                                                                                                                            
    shadows: { sm: '0 1px 2px rgba(0,0,0,.2)', lg: '0 8px 24px rgba(0,0,0,.3)' },                                                                                                                                                                
  }                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                 
  // A heavy but realistic <Box>: most props responsive so each parser emits                                                                                                                                                                     
  // media queries at every breakpoint.                                                                                                                                                                                                          
  const props = {                                                                                                                                                                                                                                
    theme,                                                                                                                                                                                                                                       
    width: [1, 1 / 2, 1 / 3, 1 / 4],                                                                                                                                                                                                             
    m: [1, 2, 3, 4],                                                                                                                                                                                                                             
    mt: [0, 1, 2],                                                                                                                                                                                                                               
    p: [4, 3, 2],                                                                                                                                                                                                                                
    px: [2, 3],                                                                                                                                                                                                                                  
    display: ['block', 'flex', 'inline-block'],                                                                                                                                                                                                  
    height: [64, 128, 256],                                                                                                                                                                                                                      
    minHeight: [32, 64],                                                                                                                                                                                                                         
    maxHeight: [128, 256],                                                                                                                                                                                                                       
    minWidth: [16, 32],                                                                                                                                                                                                                          
    maxWidth: [256, 512],                                                                                                                                                                                                                        
    size: [32, 64],                                                                                                                                                                                                                              
    textAlign: ['left', 'center', 'right'],                                                                                                                                                                                                      
    borderRadius: [1, 2],                                                                                                                                                                                                                        
    boxShadow: ['sm', 'lg'],                                                                                                                                                                                                                     
    overflow: ['hidden', 'auto'],                                                                                                                                                                                                                
  }                                                                                                                                                                                                                                              
                                       // Box's exact composition (packages/core/src/Box/Box.tsx).                                                                                                                                                                                    
  const BOX_PARSERS = [                                                                                                                                                                                                                          
    'width',                                                                                                                                                                                                                                     
    'display',                                                                                                                                                                                                                                   
    'height',                                                                                                                                                                                                                                    
    'maxHeight',                                                                                                                                                                                                                                 
    'maxWidth',                                                                                                                                                                                                                                  
    'minHeight',                                                                                                                                                                                                                                 
    'minWidth',                                                                                                                                                                                                                                  
    'size',                                                                                                                                                                                                                                      
    'space',                                                                                                                                                                                                                                     
    'textAlign',                                                                                                                                                                                                                                 
    'borderRadius',                                                                                                                                                                                                                              
    'boxShadow',                                                                                                                                                                                                                                 
    'overflow',                                                                                                                                                                                                                                  
  ]                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                 
  // compose() once per module (setup, not the measured job); each measured call is                                                                                                                                                              
  // a single parse of the maximal props — exactly what Box does per render.                                                                                                                                                                     
  const cache = new WeakMap<object, (p: unknown) => unknown>()                                                                                                                                                                                   
  const boxParser = (mod: Record<string, unknown>) => {                                                                                                                                                                                          
    let parser = cache.get(mod)                                                                                                                                                                                                                  
    if (!parser) {                                                                                                                                                                                                                               
      const fns = BOX_PARSERS.map((n) => mod[n]).filter((f) => typeof f === 'function')                                                                                                                                                          
      parser = (mod.compose as (...f: unknown[]) => (p: unknown) => unknown)(...fns)                                                                                                                                                             
      cache.set(mod, parser)                                                                                                                                                                                                                     
    }                                                                                                                                                                                                                                            
    return parser                                                                                                                                                                                                                                
  }                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                 
  export default {                                                                                                                                                                                                                               
    name: 'pcln-design-system <Box> styling',                                                                                                                                                                                                    
    packages: {                                                                                                                                                                                                                                  
      ours: '@soroush.tech/styled-system@5.6.0',                                                                                                                                                                                                 
      orig: 'styled-system@5.1.5',                                                                                                                                                                                                               
    },                                                                                                                                                                                                                                           
    cases: {                                                                                                                                                                                                                                     
      'ours (fork 5.6.0)': ({ modules }: { modules: Record<string, Record<string, unknown>> }) =>                                                                                                                                                
        boxParser(modules.ours)(props),                                                                                                                                                                                                          
      'orig (jxnblk 5.1.5)': ({ modules }: { modules: Record<string, Record<string, unknown>> }) =>                                                                                                                                              
        boxParser(modules.orig)(props),                                                                                                                                                                                                          
    },                                                                                                                                                                                                                                           
    options: { gc: 'inner', rounds: 3, warmup: 1000 },                                                                                                                                                                                           
  }                   
  ```

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -174,8 +174,8 @@ importers:
         specifier: ^7.6.2
         version: 7.6.2(react@18.3.1)
       styled-system:
-        specifier: ^5.1.5
-        version: 5.1.5
+        specifier: npm:@soroush.tech/styled-system@^5.6.0
+        version: /@soroush.tech/styled-system@5.6.0(prop-types@15.8.1)
     devDependencies:
       '@babel/cli':
         specifier: ^7.23.4
@@ -201,9 +201,6 @@ importers:
       '@storybook/testing-library':
         specifier: ^0.2.2
         version: 0.2.2
-      '@styled-system/theme-get':
-        specifier: ^5.1.2
-        version: 5.1.2
       '@testing-library/jest-dom':
         specifier: ^6.1.5
         version: 6.1.5(@types/jest@28.1.3)(jest@29.7.0)
@@ -264,9 +261,6 @@ importers:
 
   ../../packages/carousel:
     dependencies:
-      '@styled-system/theme-get':
-        specifier: ^5.1.2
-        version: 5.1.2
       lodash.debounce:
         specifier: ^4.0.8
         version: 4.0.8
@@ -282,6 +276,9 @@ importers:
       react-intersection-observer:
         specifier: ^9.5.3
         version: 9.5.3(react@18.3.1)
+      styled-system:
+        specifier: npm:@soroush.tech/styled-system@^5.6.0
+        version: /@soroush.tech/styled-system@5.6.0(prop-types@15.8.1)
       uuid:
         specifier: ^9.0.0
         version: 9.0.1
@@ -410,8 +407,8 @@ importers:
         specifier: ^5.3.11
         version: 5.3.11(@babel/core@7.23.5)(react-dom@18.3.1)(react-is@18.3.1)(react@18.3.1)
       styled-system:
-        specifier: ^5.1.5
-        version: 5.1.5
+        specifier: npm:@soroush.tech/styled-system@^5.6.0
+        version: /@soroush.tech/styled-system@5.6.0(prop-types@15.8.1)
     devDependencies:
       '@babel/cli':
         specifier: ^7.22.15
@@ -482,9 +479,6 @@ importers:
       '@radix-ui/react-visually-hidden':
         specifier: ^1.0.3
         version: 1.0.3(@types/react-dom@18.3.0)(@types/react@18.3.5)(react-dom@18.3.1)(react@18.3.1)
-      '@styled-system/theme-get':
-        specifier: ^5.1.2
-        version: 5.1.2
       '@types/styled-system':
         specifier: ^5.1.22
         version: 5.1.22
@@ -507,8 +501,8 @@ importers:
         specifier: ^9.5.3
         version: 9.5.3(react@18.3.1)
       styled-system:
-        specifier: ^5.1.5
-        version: 5.1.5
+        specifier: npm:@soroush.tech/styled-system@^5.6.0
+        version: /@soroush.tech/styled-system@5.6.0(prop-types@15.8.1)
     devDependencies:
       '@babel/cli':
         specifier: ^7.22.15
@@ -642,9 +636,6 @@ importers:
 
   ../../packages/icons:
     dependencies:
-      '@styled-system/theme-get':
-        specifier: ^5.1.2
-        version: 5.1.2
       core-js:
         specifier: ^3.37.1
         version: 3.37.1
@@ -728,14 +719,11 @@ importers:
         specifier: ^5.3.11
         version: 5.3.11(@babel/core@7.23.5)(react-dom@18.3.1)(react-is@18.3.1)(react@18.3.1)
       styled-system:
-        specifier: ^5.1.5
-        version: 5.1.5
+        specifier: npm:@soroush.tech/styled-system@^5.6.0
+        version: /@soroush.tech/styled-system@5.6.0(prop-types@15.8.1)
 
   ../../packages/menu:
     dependencies:
-      '@styled-system/theme-get':
-        specifier: ^5.1.2
-        version: 5.1.2
       prop-types:
         specifier: ^15.8.1
         version: 15.8.1
@@ -819,17 +807,14 @@ importers:
         specifier: ^5.3.11
         version: 5.3.11(@babel/core@7.23.5)(react-dom@18.3.1)(react-is@18.3.1)(react@18.3.1)
       styled-system:
-        specifier: ^5.1.5
-        version: 5.1.5
+        specifier: npm:@soroush.tech/styled-system@^5.6.0
+        version: /@soroush.tech/styled-system@5.6.0(prop-types@15.8.1)
 
   ../../packages/modal:
     dependencies:
       '@reach/dialog':
         specifier: ^0.16.2
         version: 0.16.2(react-dom@18.3.1)(react@18.3.1)
-      '@styled-system/theme-get':
-        specifier: ^5.1.2
-        version: 5.1.2
       react-transition-group:
         specifier: ^4.4.5
         version: 4.4.5(react-dom@18.3.1)(react@18.3.1)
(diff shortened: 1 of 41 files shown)


soroushm:
this is an inactive repo i remove the fork repo

PR #1569, opened 2026-07-14
fix(DX-2098): restore children on FormFieldProps in emitted types

raclee7:


raclee7:
Closing with workaround https://github.com/pcln/pcln-web/pull/20528 (in favor of reducing development efforts around legacy code; we are deprecating legacy design-system and migrating to a new one)

